PR #6, opened 2015-05-13
Issue 1316: The traineddata file must be closed after it was opened

jimregan:
Migrating from https://code.google.com/p/tesseract-ocr/issues/detail?id=1316


jimregan:
For posterity: http://web.archive.org/web/20150509202204/https://code.google.com/p/tesseract-ocr/issues/detail?id=1316

tfmorris:
:+1:  LGTM, but it's on an error path, so it's a pretty unlikely resource leak.

stweil:
@amitdo, I suggest to apply this PR.

PR #9, opened 2015-05-13
Issue 1106: [PATCH] Fixed potential usage of uninitialized bool varia…

jimregan:
…ble in textord/tospace.cpp

https://code.google.com/p/tesseract-ocr/issues/detail?id=1106

Reported by ettl.martin78, Feb 11, 2014
Please review the attached patch. It fixes a potential usage of an uninitialized bool variable ('fuzzy_sp'). The fix simply initialized the variable by default to 'FALSE'. Before the fix, the value of 'fuzzy_sp' was not set, but used in the else-branch in line 1107:

```
      else {
        prev_blanks = blanks;
        prev_fuzzy_sp = fuzzy_sp;
        prev_fuzzy_non = fuzzy_non;
      }
```

Best regards and many thanks

Feb 11, 2014
#1 ettl.martin78

The updated patch fixes two more uninitialized variable usages in the same function.

Many thanks for reviewing.

(http://web.archive.org/web/20150509223835/https://code.google.com/p/tesseract-ocr/issues/detail?id=1106)

### DIFF
--- a/textord/tospace.cpp
+++ b/textord/tospace.cpp
@@ -932,6 +932,9 @@ ROW *Textord::make_prop_words(
   box_it.set_to_list (row->blob_list ());
   word_it.set_to_list (&words);
   bol = TRUE;
+  fuzzy_non = FALSE;
+  fuzzy_sp = FALSE;
+  blanks = 0;
   prev_blanks = 0;
   prev_fuzzy_sp = FALSE;
   prev_fuzzy_non = FALSE;


textord/tospace.cpp
```diff
@@ -932,6 +932,9 @@ ROW *Textord::make_prop_words(
   box_it.set_to_list (row->blob_list ());
   word_it.set_to_list (&words);
   bol = TRUE;
+  fuzzy_non = FALSE;
```
tfmorris: These initializations look redundant to me. They are already initialized at declaration time near line 894.

Having said that, I find the mix of declaration time and non-declaration initializations needlessly complex and think all of these later initializations should get rolled back up into the declarations.

amitdo:
@orbitcowboy,
Can you reply to @tfmorris's comment?
https://github.com/tesseract-ocr/tesseract/pull/9#r53199684

textord/tospace.cpp
stweil: Ray addressed that issue after Martin's  bug report: commit 97080412f fixes all three cases, so this pull request can be closed.

stweil:
This PR can be closed, see commit 97080412f which fixes the same issue [here](https://github.com/tesseract-ocr/tesseract/commit/97080412fdc8cef05b1f917e698277cd01c6f004#diff-262ca20242c273c5ad4f245d251b6aa8R894).

amitdo:
Thanks Stefan. Somehow I missed that commit.

PR #10, opened 2015-05-13
Issue 1199: Remove the vestigal traces of custom memory allocators

jimregan:
(http://web.archive.org/web/20150509171841/https://code.google.com/p/tesseract-ocr/issues/detail?id=1199)

I've put both into the same commit, even though they need to be considered separately: I didn't think they would apply cleanly.


jimregan:
Split into pull request 11 and pull request 12

PR #11, opened 2015-05-13
Issue 1199: Remove the vestigal traces of custom memory allocators, part 1, updated

jimregan:
1. memry removal.

This should be no-brainer. These functions never seemed to do anything useful. The only oddity was with alloc_string which rounded the allocation upwards to nearest multiple of 4. Based on my review of code, I could not identify any value in this feature, so even alloc_string was simply transformed to malloc.

memry.\* are removed. All code is adjusted to simply call malloc and free instead.

listio.\* are removed. The list reading function in these files appears to have been unused, and contained the only invocation of strsave define, which was one of the alloc_string users, so I removed both of them.

https://code.google.com/p/tesseract-ocr/issues/detail?id=1199


tfmorris:
Archive copy of issue (but full text is included above): https://web.archive.org/web/20151128081631/http://code.google.com/p/tesseract-ocr/issues/detail?id=1199

amitdo:
I'm not sure about the `alloc_string()` removal. The other things seem fine.

amitdo:

PR #12, opened 2015-05-13
Gcode issue1199 p2

jimregan:
Split commits; builds on -p1, so includes that patch. Updated to current version


tfmorris:
Internet Archive copy of discussion in original issue: https://web.archive.org/web/20151128081631/http://code.google.com/p/tesseract-ocr/issues/detail?id=1199

jimregan:
I think this one can be safely closed.

PR #13, opened 2015-05-13
Issue 1353: Patch for /training/tessopt.cpp

jimregan:
https://code.google.com/p/tesseract-ocr/issues/detail?id=1353

When training batch boxes and tifs this causes training to fail after 2nd iteration. The issue is that the tessopt iterator index is a global. The fix is to 1.) reset before each tessopt usage 2.) wrap the iterator state in a class/generator. Low priority, but useful if we want to generate a bunch of traineddatas in the same process or even multiple threads.

(http://web.archive.org/web/20150413012203/https://code.google.com/p/tesseract-ocr/issues/detail?id=1353)

### DIFF
--- a/training/tessopt.cpp
+++ b/training/tessopt.cpp
@@ -57,3 +57,9 @@ const char *arglist                    //string of arg chars
   else
     return EOF;
 }
+
+void tessopt_reset()
+{
+  tessoptind = 0;
+  tessoptarg = 0;
+}


Shreeshrii:
It is a 3 year old PR. I will request @stweil to review. Thanks.

stweil:
It looks like the start of a bug fix. Please keep it open as a reminder. I added the issue / pull request to the [planning page](https://github.com/tesseract-ocr/tesseract/wiki/Planning) in the wiki.

amitdo:
I think the OP @oulrich1

oulrich1:
Wow, it's been a while. Seems a lot has changed since then. Doesn't look like 'tessopt' is used in 'lstmtraining', but still definitely used in 'cntraining' and maybe 'mftraining'.  
It does look like the patch is missing the call, maybe i didn't include it originally, but i believe should be callable from anywhere since the tessopt_reset is a free function. Might have been calling that from within a facade class, i dont remember exactly

zdenop:
Closing as this it not (at least full) patch, there is no test case for replicating problem, so we can try to fix it...

PR #15, opened 2015-05-13
Ocricola (H)FST support

jimregan:
This adds support from the ocricola project to use finite state transducers in HFST format instead of DAWGs.

Branch is currently quite stale, but while I'm pulling all these pull requests together, I thought I should put this one in too.


jimregan:
Things to do:
- [ ] Match code formatting to rest of Tesseract
- [ ] Remove exceptions
- [ ] Remove STL

amitdo:
I don't think it is going to happen.
@theraysmith, should we close this PR?

theraysmith:
Not needed. The new engine works well with the dawgs.

jimregan:
This is intended more to be an alternative to dawgs for morphologically complex languages that have a morphology available in HFST format.

jimregan:
@mpsilfve do you have anything to add?

stweil:
How can we track this pull request and also #31? I think both should be considered for a future release of Tesseract.

zdenop:
There are still branches with this code. IMHO if ray with go with dawgs alternatives will not very used... For several years nothing happened in this way and I am afraid that this will be in future the same.

amitdo:
>mpsilfve and others added some commits on **Mar 7, 2014**

PR #16, opened 2015-05-13
if data_ is not null (has been used), destroy it before alloc - *may* fix gcode issue 1390

jimregan:
I can't coax valgrind into installing, though, so I can't confirm

https://code.google.com/p/tesseract-ocr/issues/detail?id=1390

### DIFF
--- a/ccutil/strngs.cpp
+++ b/ccutil/strngs.cpp
@@ -52,6 +52,9 @@ const int kMaxDoubleSize = 15;
 const int kMinCapacity = 16;
 
 char* STRING::AllocData(int used, int capacity) {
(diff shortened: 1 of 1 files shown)


jimregan:
http://web.archive.org/web/20150413012242/https://code.google.com/p/tesseract-ocr/issues/detail?id=1390

jimregan:
==30342== Command: ./api_test_init
==30342== 
==30342== Invalid write of size 4
==30342==    at 0x5193183: STRING::AllocData(int, int) (strngs.cpp:62)

Valgrind doesn't like it; back to the drawing board.

PR #17, opened 2015-05-13
Issue 1351: OpenCL build - kernel_ThresholdRectToPix() not accounting for padding bits in the output pix?!

jimregan:
https://code.google.com/p/tesseract-ocr/issues/detail?id=1351
(http://web.archive.org/web/20150413012200/https://code.google.com/p/tesseract-ocr/issues/detail?id=1351)

### DIFF
--- a/opencl/oclkernels.h
+++ b/opencl/oclkernels.h
@@ -1045,19 +1045,19 @@ KERNEL(
 // imageData is input image (24-bits/pixel)
 // pix is output image (1-bit/pixel)
 KERNEL(
-\n#define CHAR_VEC_WIDTH 8 \n
+\n#define CHAR_VEC_WIDTH 4 \n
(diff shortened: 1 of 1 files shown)


zdenop:
I tested this patch and IMO it is only partial solution (for some cases). Issue was escalated to original OpenCL contributor.

markboettcher:
Just glanced at the code and wondered why on Earth multiplication and division is being done in every iteration, especially on inner, nested loops. Look at this:

PIXELS_PER_BURST*NUM_CHANNELS)/CHAR_VEC_WIDTH

This is inside a loop nested 3 deep. How many wasted clock cycles is this? Are we assuming that the compiler will optimize this out? If you're benchmarking with debug builds, your numbers are going to be way off, as in meaningless.

zdenop:
Merged, so we have at least partial fix...

PR #27, opened 2015-05-18
Monitor

jimregan:
Zdenko writes:

I found one older patch[1] comming from interesting androidapp Text Fairy (OCR Text Scanner) [2], [3]. I put it to separate branch (monitor) and spitted original patch to 3 commits for testing/cherry picking:
adds monitor/ETEXT_DESC to GetHOCRText
extends ETEXT_DESC ETEXT_DESC with PROGRESS_FUNC field and changed the percentage progress values to start with 0% instead of 30%.
extends hocr output by row attribute I skipped part where patch hard-code font (size?) to 15...

[1] https://www.mail-archive.com/tesseract-ocr@googlegroups.com/msg08089.html
[2] https://github.com/renard314/textfairy
[3] https://play.google.com/store/apps/details?id=com.renard.ocr

### DIFF
--- a/api/baseapi.h
+++ b/api/baseapi.h
@@ -587,6 +587,16 @@ class TESS_API TessBaseAPI {
    * Make a HTML-formatted string with hOCR markup from the internal
    * data structures.
    * page_number is 0-based but will appear in the output as 1-based.
+   * monitor can be used to
+   * 	cancel the regocnition
+   * 	receive progress callbacks
+   */
+  char* GetHOCRText(struct ETEXT_DESC* monitor, int page_number);
+
+  /**
+   * Make a HTML-formatted string with hOCR markup from the internal
+   * data structures.
+   * page_number is 0-based but will appear in the output as 1-based.
    */
   char* GetHOCRText(int page_number);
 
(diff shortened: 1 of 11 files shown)


jimregan:
n-way history is in #26

api/baseapi.h
```diff
@@ -587,6 +587,16 @@ class TESS_API TessBaseAPI {
    * Make a HTML-formatted string with hOCR markup from the internal
    * data structures.
    * page_number is 0-based but will appear in the output as 1-based.
+   * monitor can be used to
+   * 	cancel the regocnition
```
jimregan: Typo: 'recognition'

theraysmith:
I like the spirit of the changes, except that I think it would be better to add some progress code to layout analysis than to start the progress at 0 for word recognition.
The time taken by layout analysis is heavily dependent on the PageSegMode, so it would be better if it (Tesseract::SegmentPage/Tesseract::AutoPageSeg/ColumnFinder::FindBlocks) could call the callbacks and return a number that becomes the base percentage to the recognition phase instead of using either an arbitrary 30, 0 or even worse, making it depend on the existence of a callback function.
That is a bigger change of course.

zdenop:
If there is no space to implement progress monitor for layout analysis for 3.04 release I would suggest to merge this change with explanation that this is only progress monitor for word recognition and it does not cover layout analysis...

PR #38, opened 2015-06-16
free clusterer fixed a memory leak. TODO: free protolists referenced in normprotolist

oulrich1:
cntraining clusterer's are deleted, the last one is saved and used during final write, it seems. However there are still leaks with the protolists.. we still store the old protolists in the NormProtoList. This needs to be fixed somehow. 

mftraining memory leaks are fixed.


theraysmith:
I think I fixed this issue with the code I pushed in 4.00.

PR #45, opened 2015-07-13
Rename README as README.md.

jwilk:
The file itself already contained a reference to the new name.


zdenop:
autotool requires README!

jwilk:
Oops.

Do you want me to teach automake not to require existence of README;
or just fix the URL in the file?

zdenop:
File was renamed to README.md (in a0ea634) therefore there is reference to README.md in README(it is fixed in 4f082a9). 
But autogen.sh fails without README, so I put it back (in 9b7f252)

jimregan:
http://stackoverflow.com/questions/15013672/use-autotools-with-readme-md says 
```
AM_INIT_AUTOMAKE([foreign])
```
is enough to get around the need for README.

PR #60, opened 2015-07-23
strcasestr needed on Cygwin too

jimregan:
See: https://groups.google.com/d/msgid/tesseract-ocr/55B12C3C.3010908%40vol.at
```
pango_font_info.cpp:223:46: error: 'strcasestr' was not declared in this scope
   is_fraktur_ = (strcasestr(family, "Fraktur") != NULL);
```

### DIFF
--- a/training/pango_font_info.cpp
+++ b/training/pango_font_info.cpp
@@ -22,11 +22,11 @@
 #include "config_auto.h"
 #endif
 
-#ifdef MINGW
+#if (defined MINGW) || (defined __CYGWIN__)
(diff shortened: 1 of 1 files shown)


matzeri:
the solution is wrong. As strcasestr is not standard,  it is enough to

#define _GNU_SOURCE
before 
#include <string.h>

matzeri:
#include <string.h>

zdenop:
Just for future investigation: https://stackoverflow.com/questions/9935642/how-do-i-use-strcasestr

PR #69, opened 2015-07-29
Update README link to point to the wiki

McPolemic:
The current link points to the README file in master branch, rather than the wiki that the intro text implies it should be pointing to.

### DIFF
--- a/README
+++ b/README
@@ -1,7 +1,7 @@
 Note that this is a text-only and possibly out-of-date version of the 
 wiki ReadMe, which is located at:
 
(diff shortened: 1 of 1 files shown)


zdenop:
You got it wrong: this README is part of distribution packages and that it the reason why it points to its online version.
Another topic is source of information: This README can be modified only by project team members. https://github.com/tesseract-ocr/tesseract/wiki can be modified by any github users.

McPolemic:
That's fair. I saw "wiki" and was confused that it pointed to the README. Thanks for the clarification.

PR #84, opened 2015-08-24
Fixes for OpenCL issues reported on Apple Mac. Still get -54 on Apple…

gargrahul:
… Mac while running on OpenCL CPU, however it is ignored now.

### DIFF
--- a/opencl/openclwrapper.cpp
+++ b/opencl/openclwrapper.cpp
@@ -1,6 +1,7 @@
 #ifdef _WIN32
 #include <Windows.h>
 #include <io.h>
+
 #else
 #include <sys/types.h>
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


opencl/openclwrapper.cpp
```diff
@@ -14,23 +15,24 @@
...
-#else
-#define TIMESPEC timespec
+#include <stdio.h>
+#include <mach/mach_time.h>
 #endif
+#ifdef USE_OPENCL
 
 #include "opencl_device_selection.h"
+#ifdef _MSC_VER
+int LeptMsgSeverity = 3;  // L_SEVERITY_INFO
```
zdenop: This causes build error in VS2010

PR #87, opened 2015-08-26
Get --enable-opencl to compile (but not work)

jbarlow83:
I had to redo how flags were handled for OpenCL, and in the process I quite likely broke something on the *nix side. I think the template I've set up is more in line with autotools best practices (https://stackoverflow.com/questions/20230827/how-to-set-include-paths-with-autotools) and should be flexible enough to fix.

Unfortunately, with these changes OS X/OpenCL compiles, but the output of the binary is garbage, as opposed to crashing. :(

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,8 @@ training/wordlist2dawg
 *.cube.*
 *.tesseract_cube.*
 *.traineddata
+
+# OpenCL
+tesseract_opencl_profile_devices.dat
+kernel*.bin
+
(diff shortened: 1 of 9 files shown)


jbarlow83:
found that this enables OpenCL unconditionally which is bad

jbarlow83:
Should be safe-ish to pull now

zdenop:
Why you replaced standard OPENCL_CFLAGS (defined by autogen) and with user defined -I$(OPENCL_HDR_PATH)?

jbarlow83:
Where is OPENCL_* defined by autogen? I don't think it defines them automatically anywhere; it's a convention.

According to this page, it's more flexible to set up a LDFLAGS and CPPFLAGS for each third party dependency, rather than *assuming* that the third party dependency needs only a list of libraries (OPENCL_LIBS) and a single header path (OPENCL_HDR_PATH). If the third party dependency is optional, then set the flags to an empty string rather than conditionally use it.
https://stackoverflow.com/questions/20230827/how-to-set-include-paths-with-autotools

zdenop:
I am sorry I mixed it with OpenMP.

PR #106, opened 2015-09-29
Update README.md

AashishV:
Link to wiki ReadMe points to the same markdown file.


zdenop:
This file is part of source distribution and that is reason why its point to "the same" online markdown file

PR #109, opened 2015-10-09
CMake improvements

egorpugin:
Fix linux library soname.
Fix finding system liblept on linux.
Minimum lept version is set to 1.71.


egorpugin:
You see, now travis-ci does a check of any new PR and tells you about whether it passes or not.

PR #119, opened 2015-10-28
OSD: Print script name instead of meaningless script id

amitdo:
test:
```
tesseract eurotext.tif eurotext -l osd -psm 0
```


zdenop:
thanks!

amitdo:
that was quick! :-)

PR #141, opened 2015-11-06
Remove register attribute for local variables

stweil:
This fixes clang compiler warnings like this one:

wordrec/gradechop.cpp:52:3: warning:
 'register' storage class specifier is deprecated [-Wdeprecated-register]

### DIFF
--- a/ccstruct/polyaprx.cpp
+++ b/ccstruct/polyaprx.cpp
@@ -214,18 +214,18 @@ EDGEPT edgepts[]                 //output is array
 void fix2(                //polygonal approx
           EDGEPT *start,  /*loop to approimate */
           int area) {
-  register EDGEPT *edgept;       /*current point */
-  register EDGEPT *edgept1;
-  register EDGEPT *loopstart;    /*modified start of loop */
-  register EDGEPT *linestart;    /*start of line segment */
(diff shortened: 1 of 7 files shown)


zdenop:
I am not sure if this is the right way of dealing with such warning:
- register is [deprecated in C++11](http://en.cppreference.com/w/cpp/language/storage_duration), but it will exists until C++17
- AFAIK it should be possible to build tesseract without support of C++11
My suggestion is that if solving of warning does not bring any value, it should be ignored...

stweil:
@zdenop: Any good modern compiler should be able to assign CPU registers to local variables, and at least for gcc and clang I know that they do a very good job here. I assume that 'register' for local variables was deprecated because it no longer makes sense today. That's the reason why I suggest to use these modifications.

For function parameters, this is different. Therefore they are still supported.

stweil:
Extract from gcc documentation:

"Some developers use Local Register Variables in an attempt to improve gcc's allocation of registers, especially in large functions. In this case the register name is essentially a hint to the register allocator. While in some instances this can generate better code, improvements are subject to the whims of the allocator/optimizers. Since there are no guarantees that your improvements won't be lost, this usage of Local Register Variables is discouraged."

PR #143, opened 2015-11-07
Fix Travis build error (error when getting cmake)

stweil:
This command fails currently:

wget http://www.cmake.org/files/v3.3/cmake-3.3.1-Linux-x86_64.sh

Obviously cmake.org now redirects to https connections, so we have to
fix the URL.


stweil:
Maybe we have to add --no-check-certificate. Just waiting for the result of the new Travis build...

Issue #147, opened 2015-11-11
Unable to parse cyrillic


kurkale6ka:
This is the failing test: (all locales, packages seem correctly installed).
For info, parsing english works fine.

```
% tesseract test.png test -lbul
Tesseract Open Source OCR Engine v3.04.00 with Leptonica

% cat test.txt
Ila mnpmm "mm npanHM Mama m m: m: m :3 mm mm. Ms M:

% locale -a | grep bg
bg_BG.utf8

% pacman -Qs tesseract-data-bul
local/tesseract-data-bul 3.04.00-1 (tesseract-data)
```

Also tried:
* `LANG=bg_BG.UTF-8 tesseract test.png test -lbul`
* a different bul traindata: https://code.google.com/p/tesseract-ocr/downloads/detail?name=bul.traineddata.gz

kurkale6ka:
![test](https://cloud.githubusercontent.com/assets/52182/11091324/3871125e-8871-11e5-8f9d-0472f36b68b2.png)

amitdo:
Try this:
> tesseract test.png test -l bul

kurkale6ka:
Ha, that's funny, I actually tried it like this the very first time but I had an error and thought that that caused it!

Works much better like this:
`tesseract test.png test -l bul`:
`Да извършва всякакви правни „щения пт на: им: и за мая см:-дка. пя „=`

Still not 100% accurate but I guess that's up to me/options to fix now?

kurkale6ka:
Suggestion: using `-lbul` should fail with an error

amitdo:
The font in this image is quite small. You should try to rescale the image (X2-X4).
https://github.com/tesseract-ocr/tesseract/wiki/ImproveQuality

In general, the right place to ask questions is here:   
https://groups.google.com/forum/?hl=en#!forum/tesseract-ocr

kurkale6ka:
I found out exactly what were the encountered issues. Here is a sum up:

* First I ran `tesseract test.png -l bul` which fails because it sees an option instead of `outputbase|stdout` but I thought it was because of the space between `-l` and `bul`
* Then I tried `tesseract test.png -lbul` which doesn't give errors because it wrongly thinks that the output file is `-lbul` and it uses the default `eng` data. Indeed `-lbul.txt` gets created
* And eventually: `tesseract test.png test -lbul` which creates `test.txt` and uses `eng` as default and simply discards `-lbul` which is nonsense to it

[...]

PR #158, opened 2015-11-29
tesseractmain: Prettify help message

stweil:
Commit 99110df75781c6907c84a3d23695a6900b933a97 improved the help text
in several aspects, but also introduced new inconsistencies which this
patch tries to fix.

* Align columns (this needed replacing tabs by spaces).
* Start explaining text with uppercase.
* Replace "the stdout" by "stdout.
* Small changes in help text for page segmentation modes.
* Split options in OCR options and single options
  (partially revert commit 99110df75781c6907c84a3d23695a6900b933a97).

In addition, whitespace characters at end of lines were removed.

### DIFF
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -71,21 +71,21 @@ void PrintVersionInfo() {
 }
 
 void PrintUsage(const char* program) {
-  fprintf(stderr, 
+  fprintf(stderr,
       "Usage:\n"
-      "  %s --help | --help-psm | --version\n" 
+      "  %s --help | --help-psm | --version\n"
       "  %s --list-langs [--tessdata-dir PATH]\n"
       "  %s --print-parameters [options...] [configfile...]\n"
-      "  %s imagename|stdin outputbase|stdout [options...] [configfile...]\n", 
+      "  %s imagename|stdin outputbase|stdout [options...] [configfile...]\n",
       program, program, program, program);
 }
 
 void PrintHelpForPSM() {
-  const char* msg = 
-      "Page Segmentation Modes:\n"
+  const char* msg =
+      "Page segmentation modes:\n"
         "  0    Orientation and script detection (OSD) only.\n"
         "  1    Automatic page segmentation with OSD.\n"
-        "  2    Automatic page segmentation, but no OSD, or OCR\n"
+        "  2    Automatic page segmentation, but no OSD, or OCR.\n"
         "  3    Fully automatic page segmentation, but no OSD. (Default)\n"
         "  4    Assume a single column of text of variable sizes.\n"
         "  5    Assume a single uniform block of vertically aligned text.\n"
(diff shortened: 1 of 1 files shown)


amitdo:
>Commit 99110df... introduced new inconsistencies which this
patch tries to fix.
> * Start explaining text with uppercase.
> * Replace "the stdout" by "stdout.

It didn't introduce these inconsistencies. They were there before.

stweil:
@amitdo, I'm sorry, my commit text was not precise. Some of the issues which my patch tries to address existed before your patch. Should I update the patch with a modified commit message?

There remain more issues to be discussed, for example whether stdout instead of stderr would be better for the help text. You also raised the question about more entries for page segmentation mode.

amitdo:
> Should I update the patch with a modified commit message?

Yes please.


>whether stdout instead of stderr would be better for the help text.

I've noticed that too. I don't know why it's stderr right now.

>You also raised the question about more entries for page segmentation mode.

We need to publish them, IMHO.
 
> Split options in OCR options and single options (partially revert commit 99110df).

'list-langs' can be followed by 'tessdata-dir'.
'print-parameters' can be followed by -c option(s) which will change the output.
So why call them 'single options'?

I'm not really sure what is the right usage for 'print-parameters'.

Testing:
```
$tesseract --print-parameters | grep pageseg_mode
tessedit_pageseg_mode	6

$ tesseract --print-parameters -psm 1 | grep pageseg_mode
tessedit_pageseg_mode	6

$ tesseract --print-parameters -c tessedit_pageseg_mode=1 | grep pageseg_mode
tessedit_pageseg_mode	1

$ tesseract --print-parameters pdf | grep pageseg_mode
tessedit_pageseg_mode	6

$ tesseract nosuchfile stdout --print-parameters pdf | grep pageseg_mode
tessedit_pageseg_mode	1
```

zdenop:
--print-parameters is option that list all tesseract parameters with its default values.  This is useful if you would like to change tesseract behavior.

stweil:
@amitdo, I think it is reasonable to split the options in two groups: one group for those options which are used in production to make OCR, one group for options which show information like help texts, version or supported parameters. The old help text called the 2nd ones 'single options', that's why my patch reverted to that title. Maybe there exists a better naming.

zdenop:
in case of 'single options': it was not expected (tested) their usage in combination with other options.

amitdo:
> --print-parameters is option that list all tesseract parameters with its default values. This is useful if you would like to change tesseract behavior.

@ zdenop, I knew that, of course.

What I meant was: Should other options followed by 'print-parameters' change the output of the command?

With my commit, the CL help message include:
>Usage:
  ...
  tesseract --print-parameters [options...] [configfile...]

The current behaviour of the command line only partly match the above usage pattern. Only the '-c var=value' option changes the output.

I think this

>$ tesseract --print-parameters pdf | grep pageseg_mode

should produce

> tessedit_pageseg_mode	1

Just like:
$ tesseract --print-parameters -c tessedit_pageseg_mode=1 | grep pageseg_mode
tessedit_pageseg_mode	1

Strangly, this works right now

> tesseract nosuchfile stdout --print-parameters pdf | grep pageseg_mode
tessedit_pageseg_mode	1


> in case of 'single options': it was not expected (tested) their usage in combination with other options.

So, maybe the usage part of the help message should change accordingly?

>  tesseract --print-parameters

amitdo:
> in case of 'single options': it was not expected (tested) their usage in combination with other options.

So, maybe the usage part of the help message should change accordingly?

>  tesseract --print-parameters

amitdo:
@stwell, feel free to change it as you like, no objection.

amitdo:
@stweil,

> Should I update the patch with a modified commit message?

on second thought - Don't bother!

zdenop:
@amitdo: tesseract executable was(is) simple example how to use tesseract library. AFAIK nobody plan to implement smart handling of options (e.g. you still need to provide some information in specific order)... So the correct result of your option combination is side effect... ;-)  See e.g. https://github.com/tesseract-ocr/tesseract/issues/147#issuecomment-155892373

Issue #101, opened 2015-09-17
Windows Installer


IngoBelka:
Written in the WIKI
"Windows
An installer is available for Windows from our download page."

But as far as I can see there are no Windows installers to download. May be I'm not right, so, please, tell me, where I can get a current Windows installer.

amitdo:
There is a Windows installer for tesseract version 3.02 here:

https://code.google.com/p/tesseract-ocr/downloads/list
tesseract-ocr-setup-3.02.02.exe

I don't think there is a Windows installer for newer versions.

IngoBelka:
Ok, thank you. I took a further look for a newer version and I found:
http://domasofan.spdns.eu/tesseract/
with the newest development version. But the server is delivering the files really slow.

solomon18:
Thank you for the latest Windows installer files :).

stweil:
The NSIS installer configuration was added and removed twice, the last time with commit b8862b33dfa3bf42c87913911f62da7bc4ed2e45. Is there currently another method to build an installer for Windows? If not, I'd like to add the NSIS configuration files again. I just finished cross building Tesseract for Windows (including an installer) based on latest Git master and using the MinGW-w64 packages on Debian GNU Linux.

zdenop:
This issue is not only about creating windows installer. It is also about providing support. And on windows the there is demand for VS based library (it is quiet easy to create mingw based library with msys, but VS based library need to rebuild all dependencies with VS)

amitdo:
See also PR #162.

zdenop:
@amitdo: Future is cmake. There is no sense to make and maintain  build system for each version of cmake. cmake correctly detect installed compiler and use it. We are just waiting for leptonica transfer to github, there cmake will be presented to.

stweil:
@amitdo, that's interesting, but it still does not include an installer as far as I see.
@zdenop, maybe those Windows developers who are interested in libraries can also compile Tesseract. But those who just want to do OCR on Windows need an installer. Will cmake provide one?

Regarding leptonica, you can start with https://github.com/stweil/leptonica until there is an official GitHub version.

amitdo:
>Future is cmake. There is no sense to make and maintain build system for each version of cmake. cmake correctly detect installed compiler and use it. We are just waiting for leptonica transfer to github, there cmake will be presented to.

https://github.com/DanBloomberg/leptonica

Shreeshrii:
[...]

PR #162, opened 2015-12-02
vs2015

tofutim:
If https://github.com/loqu8/tesseract-vs2015 is placed parallel to the tesseract repo, you can use the vs2015 sln to compile tesseract in Visual Studio with giflib, libpng, libjpeg, libtiff, libwebp and zlib. Note that the solution will look for includes and lib in ../../../tesseract-vs2015/release rather than ../../../ like the VS2010 sln.


egorpugin:
-1. It's better to improve cmake build system rather than each solution per VS version.

amitdo:
@zdenop said in #101:
>Future is cmake. There is no sense to make and maintain build system for each version of cmake. cmake correctly detect installed compiler and use it. We are just waiting for leptonica transfer to github, there cmake will be presented to.

So, this PR should be closed.

egorpugin:
Use cmake for any Visual studio builds.

PR #167, opened 2015-12-09
Get tessdata prefix from executable path (only for Windows)

stweil:


zdenop:
VS 2010 build fails with:
..\..\ccutil\mainblk.cpp(28): fatal error C1083: Cannot open include file: 'libgen.h': No such file or directory

stweil:
Obviously VS does not support the POSIX function `dirname` and its header file `libgen.h`. I'll update the patch to use `splitpath` (or is there a better alternative for Windows code?).

stweil:
VS does not know the `F_OK` needed for function `access`, so I need one more iteration which replaces that parameter by 0.

PR #218, opened 2016-02-10
Remove const from STRING::string()

eyolfson:
See comment in change. Perhaps this should be refactored a better way?


tfmorris:
This doesn't even compile (which is why the tests fail).

Issue #182, opened 2016-01-04
Some programs can't find OCR text in Tesseract's PDFs (3.04)


jbarlow83:
While Acrobat XI can find text in a PDF, it appears that [poppler](https://en.wikipedia.org/wiki/Pdftotext)'s `pdftotext` program, OS X's Preview app, and the library [PyPDF2](https://github.com/mstamy2/PyPDF2/tree/master/PyPDF2)'s extractText() function all fail to locate text. It seems that Tesseract is encoding text in a way that makes it inaccessible to many PDF viewers.

`pdftotext` produces empty output.
Preview app allows highlighting of text in the appropriate locations, but it cannot be copied to the clipboard or searched.
PyPDF2 extractText also produces an empty string as text.

amitdo:
See #170.

jbarlow83:
#170 might be related, but the files I checked did not have tilted or skewed text.

Input file:
![linnsequencer](https://cloud.githubusercontent.com/assets/1825843/12100857/6fba9e66-b2e6-11e5-8df1-d928e75a1b44.jpg)

Output:
[linn.pdf](https://github.com/tesseract-ocr/tesseract/files/77727/linn.pdf)

tesseract version
``
tesseract 3.04.00
 leptonica-1.72
  libjpeg 8d : libpng 1.6.19 : libtiff 4.0.6 : zlib 1.2.5
``

amitdo:
Chromium's pdf reader output (cut&paste):

>The LinnSequencer
32 Track MIDI Sequence Recorder
The LinnSequencer is state—of—the-art composition and performance tool for the professional musician. It is
extremely powerful, yet amazingly simple to learn and use. It’s many remarkable features include:
Operation is similar to multi-track tape recorder with PLAY, STOP, RECORD, FAST
FORWARD, REWIND, and LOCATE controls.
Each of the 100 sequences contains 32 simultaneous, polyphonic tracks. Each track may
be assigned to one of 16 MIDI channels. Simultaneously plays up to 16 polyphonic
synthesizers
Ultra-fast 1/2 disk drive stores complex songs in seconds and holds over 110,000 notes
per disk!
One or all tracks may be TRANSPOSED at the touch of key.
Exclusive real—time ERASE function makes editing FAST.
Exclusive REPEAT function automatically repeats any held notes at pre-selected
rhythmic value.
TIMING CORRECTION works during playback and operates without ‘chopping’ notes.
Optional SMPTE time code synchronization.
[...]

PR #220, opened 2016-02-11
Replace pdf.ttf with sharp2.ttf, keep name the same

jbarlow83:
As discussed at length in issue #182, the existing pdf.ttf causes difficulties
for certain PDF viewers, in part because the old file had zero advance width.

With testing, sharp2.ttf seems to be the best available compromise, although
it's not perfect and causes some visual difficulties in Evince.  It does
seem to fix Kindle and OS X Preview.


amitdo:
Did someone test that there is no regression in pdf output with Adobe Acrobat
and Chromium?

jbarlow83:
Test PDF is here:
https://github.com/tesseract-ocr/tesseract/files/119875/simple-1.pdf

jbarlow83:
I have checked: Acrobat XI, Google Chrome PDF Viewer, OS X Preview, Safari PDF viewer; all on El Capitan.

jbarlow83:
See #182 for other tests people did with sharp2.ttf

jbreiden:
Chromium and Adobe Reader on Linux are fine. I have reports that Ghostscipt and friends are okay. I should probably double check Android right now.

jbreiden:
Latest stock Android (Marshmallow) is fine.

amitdo:
Okay, but I think you should also test this with an image that has more than two words...

rossj:
Using `sample-1.pdf` from above in the latest pdf.js results in the 2nd word highlighting properly, while the 1st word's highlight is offset. I don't think this is a regression (and it might actually be an improvement) as there were highlighting-offset issues using the previous pdf.ttf with pdf.js. Related to https://github.com/mozilla/pdf.js/issues/6863. 

screen shot 2016-02-12 at 8 10 39 am

jbreiden:
I've been tracking the Firfox offset problem in https://github.com/mozilla/pdf.js/issues/6509

PR #243, opened 2016-02-28
Add TessBaseApi::GetTessDataDir

manisandro:
Trivial patch to expose the used tessdata_prefix in the public API.


amitdo:
https://github.com/tesseract-ocr/tesseract/blob/master/api/baseapi.cpp#L954

>const char *  TessBaseAPI::GetDatapath() {
  return tesseract_->datadir.c_str();
}

manisandro:
Oh, not sure how I managed to miss that one. Thanks!

PR #273, opened 2016-03-16
Print version to stdout instead to stderr

stweil:
Most command line programs print the version to stdout.
This seams to be reasonable for Tesseract, too.

Now a shell statement like "VERSION=$(tesseract --version)" works
without I/O redirection.


stweil:
If there are no objections, I'd create a pull request to write help messages to `stdout`, too.

Issue #195, opened 2016-01-21
text2image segfault


LinusU:
I'm trying to use the `text2image` utility to train tesseract. Unfortunately it keeps crashing every time I try to use it :(

```sh
text2image --text=training_text.txt --outputbase=test.MenloMedium.exp0 --font='Menlo Medium' --fonts_dir=/Library/Fonts/
```

```text
(lldb) run
Process 49926 launched: '/usr/local/bin/text2image' (x86_64)
Process 49926 stopped
* thread #1: tid = 0x1d2b8cb, 0x0000000100b74358 libpangoft2-1.0.0.dylib`pango_fc_font_get_glyph + 25, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000100b74358 libpangoft2-1.0.0.dylib`pango_fc_font_get_glyph + 25
libpangoft2-1.0.0.dylib`pango_fc_font_get_glyph:
->  0x100b74358 <+25>: movq   (%rcx), %rdi
    0x100b7435b <+28>: testq  %rdi, %rdi
    0x100b7435e <+31>: je     0x100b74369               ; <+42>
    0x100b74360 <+33>: movq   %rax, %rsi
(lldb) bt
* thread #1: tid = 0x1d2b8cb, 0x0000000100b74358 libpangoft2-1.0.0.dylib`pango_fc_font_get_glyph + 25, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000100b74358 libpangoft2-1.0.0.dylib`pango_fc_font_get_glyph + 25
    frame #1: 0x000000010000edc1 text2image`tesseract::PangoFontInfo::CanRenderString(char const*, int, std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >*) const + 321
    frame #2: 0x000000010000ec57 text2image`tesseract::PangoFontInfo::CanRenderString(char const*, int) const + 33
    frame #3: 0x0000000100015227 text2image`tesseract::StringRenderer::StripUnrenderableWords(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >*) const + 193
[...]

PR #275, opened 2016-03-16
Add missing % char from format specifier in tlog()

amitdo:
- In training/ango_font_info.cpp

### DIFF
--- a/training/pango_font_info.cpp
+++ b/training/pango_font_info.cpp
@@ -534,7 +534,7 @@ bool FontUtils::IsAvailableFont(const char* input_query_desc,
        pango_font_description_get_weight(selected_desc));
 
   char* selected_desc_str = pango_font_description_to_string(selected_desc);
-  tlog(2, "query_desc: '%s' Selected: 's'\n", query_desc.c_str(),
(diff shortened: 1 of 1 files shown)


stweil:
Good catch. There seem to be more errors of this kind.

amitdo:
>Good catch. 

Thanks. I've noticed that missing symbol in #195.

>There seem to be more errors of this kind.

I see you already hunted two... :-)

stweil:
The trick is to compile with `-Wmissing-format-attribute`, fix all related warnings and iterate this until no format related warnings remain. Projects like Tesseract which never used gcc's format attribute typically have lots of format errors. I just started the search, and the current count is 6.

Fixing those format errors is one task, another one is adding the format attribute and the compiler options where needed, so future contributions won't introduce new errors.

amitdo:
Would that flag catch that missing % from format specifier?

stweil:
That flag would result in warnings for all functions with printf or scanf like arguments which don't use the format attribute. Then it is necessary to add a format function attribute to each of those function declarations. After this, gcc will know which functions work similar to printf or scanf and report all errors (if that warning is enabled, for example with `-Wall -Wextra`).

PR #298, opened 2016-04-12
When off_t is unavailable, define it as long long

mmcco:
It is currently defined as long if unavailable. This doesn't violate any
C specifications, but it's suboptimal because it can prevent >2 GB files
from being handled as expected. Because of this, some platforms always
make it 64-bit, and this is probably the best option.

### DIFF
--- a/ccutil/scanutils.cpp
+++ b/ccutil/scanutils.cpp
@@ -39,7 +39,7 @@
 
 // workaround for "'off_t' was not declared in this scope" with -std=c++11
 #if !defined(HAVE_OFF_T)
-typedef long off_t;
+typedef long long off_t;
 #endif  // off_t
 
 enum Flags {


stweil:
I don't think that this is a good idea. I'd remove that code part completely. All modern operating systems have header files which define the correct off_t, so local definitions are not needed. As off_t can be part of system defined structures and is also used as parameter for system calls, choosing an arbitrary type like `long long` won't work.

mmcco:
Sounds good to me. That would be my preferred outcome, but I figured there was a reason why this macro condition existed. Hopefully it's obsolete and can be removed.

And good point about the syscall APIs - I hadn't thought of that. It's worth mentioning, though, that in that context defining `off_t` as `long` is as broken as defining it as `long long`.

stweil:
Yes, any local definition is bad here. That's why I suggest to remove that code. Any problems which might occur after the removal need a different solution (correct include files, maybe setting of other macros if the off_t declaration is made conditionally).

PR #299, opened 2016-04-13
Remove conditional definition of off_t

mmcco:
As pointed out by Stefan Weil, conditionally defining off_t using a
macro isn't a valid approach. off_t does not have a fixed size and is
used in ABI definitions (e.g. syscalls), so silently guessing its size
risks breaking the build. Additionally, all sane and modern platforms
will have off_t.


mmcco:
As discussed in #298

stweil:
@zdenop, could you please apply this PR? Thanks, Stefan

PR #330, opened 2016-05-22
Training tools: Print help message when (argv == 1)

amitdo:


training/commandlineflags.cpp
```diff
@@ -144,6 +144,12 @@ void PrintCommandLineFlags() {
 void ParseCommandLineFlags(const char* usage,
                            int* argc, char*** argv,
                            const bool remove_flags) {
+  if (*argc == 1) {
+    tprintf("USAGE: %s\n", usage);
```
stweil: I suggest to use the  more common "Usage:" instead of "USAGE:".

training/commandlineflags.cpp
```diff
@@ -144,6 +144,12 @@ void PrintCommandLineFlags() {
...
+    PrintCommandLineFlags();
+    exit(0);
```
stweil: training/unicharset_extractor.cpp and training/wordlist2dawg.cpp use exit code 1.
I therefore suggest using exit(1) here, too.

amitdo:
@stweil
A few lines below my new code:

```
unsigned int i = 1;
  for (i = 1; i < *argc; ++i) {
    const char* current_arg = (*argv)[i];
    // If argument does not start with a hyphen then break.
    ...
    // If this is asking for usage, print the help message and abort.
    if (!strcmp(current_arg, "help") ||
        !strcmp(current_arg, "helpshort")) {
      tprintf("USAGE: %s\n", usage);
      PrintCommandLineFlags();
      exit(0);
    }
```

amitdo:
Anyway, you'll have to submit a new PR, since Zdenko already merged this one.

amitdo:
Also check what we are doing here:
https://github.com/tesseract-ocr/tesseract/blob/master/api/tesseractmain.cpp

stweil:
Yes, the code is not consistent – sometimes it exits with 0, sometimes it exits with 1.

IMHO, missing arguments are an error condition, so a program should not exit with 0 (which means no error) in that case.

amitdo:
You also might want to read about `EXIT_FAILURE` and `EXIT_SUCCESS`
https://www.google.co.il/search?q=EXIT_FAILURE
`EXIT_SUCCESS` is used in `text2image.cpp`.

amitdo:
Another issue.

What do you think about the code in https://github.com/tesseract-ocr/tesseract/pull/330#issuecomment-220828244

It search for 'help' string in all `argv` strings. I think it should only check `argv[1]`. This is what is being done in `tesseractmain.cpp`.

stweil:
I personally use command lines with several arguments and an added `--help` quite often. This pattern is usually supported well. Examples: `ls -l --help`, cp -R --help`.

amitdo:
Okay...

PR #340, opened 2016-06-05
Bypass Leptonica error message with pixGenHalftoneMask()

### LINKED ISSUE #292: pixGenHalftoneMask error
test

(I'm using head: 3.05.00dev)

When ocr'ing a small one or two line image (with the default settings), I'm getting the following error:
Error in pixGenHalftoneMask: pix too small: w = 610, h = 71

But the ocr is successful.

In textord/imagefind.cpp, FindImages is calling pixGenHalftoneMask (a leptonica function) which is throwing the error and returning null.

FindImages is then executing it's backup plan, which works fine (such a small image doesn't need multiple images regions). It's backup plan is to call pixCreate and then return the result.

The minimum size for pixGenHalftoneMask to work is 100x100 pixels. Perhaps FindImages should
just go with plan 'b' if the size of the image is too small. This would still work fine and avoid throwing the error. I don't know if there is a way for teseract to know what the minimum size is (in case it changes).

amitdo:
Fixes #292

### DIFF
--- a/textord/imagefind.cpp
+++ b/textord/imagefind.cpp
@@ -67,11 +67,21 @@ Pix* ImageFind::FindImages(Pix* pix) {
   if (pixGetWidth(pix) < kMinImageFindSize ||
       pixGetHeight(pix) < kMinImageFindSize)
     return pixCreate(pixGetWidth(pix), pixGetHeight(pix), 1);
+
   // Reduce by factor 2.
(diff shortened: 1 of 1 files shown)


amitdo:
I will wait a day or two before merging it...

amitdo:
@jbreiden
You commented in #292, so if you have something to add here, please do!

PR #341, opened 2016-06-09
Replace deliberate segv with raise(SIGABRT)

manisandro:
For front-ends to tesseract, the deliberate segfault [1] created when tesseract encounters a critical error is rather inconvenient, resp. makes it hard to keep the application alive ([2] works first time but fails to recover a second time). I'd like to propose replacing this with something from which applications can recover more easily, such as `raise(SIGABRT)`.

[1] https://github.com/tesseract-ocr/tesseract/blob/master/ccutil/errcode.cpp#L86
[2] https://github.com/manisandro/gImageReader/blob/0628f8c653169cb6bf53d51c6eb8921fcd5f66cb/gtk/src/Recognizer.cc#L171

### DIFF
--- a/ccutil/errcode.cpp
+++ b/ccutil/errcode.cpp
@@ -21,9 +21,7 @@
 #include          <stdlib.h>
 #include          <stdarg.h>
 #include          <string.h>
-#ifdef __UNIX__
 #include          <signal.h>
-#endif
 #include          "tprintf.h"
 #include          "errcode.h"
 
@@ -74,7 +72,6 @@ const char *format, ...          // special message
   // %s is needed here so msg is printed correctly!
   fprintf(stderr, "%s", msg);
 
-  int* p = NULL;
   switch (action) {
     case DBG:
     case TESSLOG:
@@ -82,9 +79,7 @@ const char *format, ...          // special message
     case TESSEXIT:
       //err_exit();
     case ABORT:
-      // Create a deliberate segv as the stack trace is more useful that way.
-      if (!*p)
-        abort();
+      raise(SIGABRT);
     default:
       BADERRACTION.error ("error", ABORT, NULL);
   }


ccutil/errcode.cpp
```diff
@@ -74,17 +72,14 @@ const char *format, ...          // special message
...
     case DBG:
     case TESSLOG:
       return;                    //report only
     case TESSEXIT:
       //err_exit();
     case ABORT:
-      // Create a deliberate segv as the stack trace is more useful that way.
-      if (!*p)
-        abort();
+      raise(SIGABRT);
```
stweil: Why not simply call `abort()` (which raises the same signal)?

stweil:
IMHO removing the deliberate null pointer access is a good thing. I cannot see why the SEGV would give a stack trace which is more useful than with ABRT (as it is written in the code comment).

manisandro:
Because it appears that `abort()` terminates the program even if the signal is caught, as soon as the signal handler exits. See also [1] and example program below.

    #include <csignal>
    #include <iostream>
    
    void signal_handler(int signal) {
            std::cout << "Caught signal " << signal << std::endl;
    }
    
    int main() {
            std::signal(SIGABRT, signal_handler);
            std::raise(SIGABRT);
    //      abort();
            std::cout << "Program survived" << std::endl;
            return 0;
    }

(Actually IMO the proper solution would be to allow applications to register an error handler to also make it easier to catch the error messages thrown.)

[1] http://stackoverflow.com/questions/20212927/difference-between-raisesigabrt-and-abort-methods

stweil:
Is returning to the program which called `ERRCODE::error` a good idea? Or would a typical signal handler use `longjmp`? In this 2nd case, `abort()` would not abort.

And yes, support for user registered error handlers would be the better solution.

amitdo:
Android has (had?) similar behaviour.
https://code.google.com/p/android/issues/detail?id=16672

amitdo:
@theraysmith, what's your take on that?

907231602:
**I hava the same question.why and how to resovle it........**
image

image

stweil:
@manisandro, is `raise(SIGABRT)` portable (working on Android and Windows, too)? If yes: could you please update this pull request, so that it can be merged?

amitdo:
https://code.google.com/p/android/issues/detail?id=16672

https://github.com/aosp-mirror/platform_system_core/blob/fc1cf90741e59d5615a7dcea1813f38bfa3a2eec/debuggerd/handler/debuggerd_fallback.cpp#L309

Bottom line: they moved to raise(SIGABRT), but they also handling multIthearding.

amitdo:
MS also supports this:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/raise?view=vs-2017

PR #344, opened 2016-06-12
File names in file header comments

StefRe:
Trying to familiarize with the source code I came across  some discrepancies between the file names in the file header comments and the actual file names. Maybe it's pedantic but I thought having the correct names might facilitate reading the code especially when using printouts.

Some files don't appear to have file header comments and boilerplate license statements as recommended by the [Style guide for Google-originated open-source projects](https://github.com/google/styleguide/blob/gh-pages/cppguide.html#L4537), but I don't feel that I were in a position to supply this information. This pertains to the following files:

```
ccutil\ccutil.cpp
neural_networks\runtime\*.cpp
opencl\openclwrapper.cpp
textord\gap_map.cpp
textord\tospace.cpp
training\commandlineflags.cpp
training\set_unicharset_properties.cpp
viewer\svpaint.cpp
```

### DIFF
--- a/ccstruct/coutln.h
+++ b/ccstruct/coutln.h
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:						coutln.c      (Formerly:  coutline.c)
+ * File:	coutln.h      (Formerly:  coutline.c)
  * Description: Code for the C_OUTLINE class.
  * Author:					Ray Smith
  * Created:					Mon Oct 07 16:01:57 BST 1991
--- a/ccutil/bits16.cpp
+++ b/ccutil/bits16.cpp
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        bits16.h  (Formerly bits8.h)
+ * File:        bits16.cpp  (Formerly bits8.h)
  * Description: Code for 8 bit field class.
  * Author:					Phil Cheatle
  * Created:					Thu Oct 17 10:10:05 BST 1991
--- a/ccutil/serialis.cpp
+++ b/ccutil/serialis.cpp
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        serialis.h  (Formerly serialmac.h)
+ * File:        serialis.cpp  (Formerly serialmac.h)
  * Description: Inline routines and macros for serialisation functions
  * Author:      Phil Cheatle
  * Created:     Tue Oct 08 08:33:12 BST 1991
--- a/cube/char_set.cpp
+++ b/cube/char_set.cpp
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        char_samp_enum.cpp
+ * File:        char_set.cpp (Formerly char_samp_enum.cpp)
  * Description: Implementation of a Character Set Class
  * Author:    Ahmad Abdulkader
  * Created:   2007
--- a/cube/char_set.h
+++ b/cube/char_set.h
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        char_samp_enum.h
+ * File:        char_set.h (Formerly char_samp_enum.h)
  * Description: Declaration of a Character Set Class
  * Author:    Ahmad Abdulkader
  * Created:   2007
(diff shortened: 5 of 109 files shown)


ccstruct/coutln.h
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:						coutln.c      (Formerly:  coutline.c)
+ * File:	coutln.h      (Formerly:  coutline.c)
```
amitdo: `Formerly coutline.c` should be removed.

jbreiden:
Tesseract predates Google.

ccutil/serialis.cpp
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        serialis.h  (Formerly serialmac.h)
+ * File:        serialis.cpp  (Formerly serialmac.h)
```
amitdo: `Formerly serialmac.h` should be replaced with `Formerly serialmac.c`

ccstruct/coutln.h
amitdo: `Formerly coutline.c` should be removed.

ccutil/bits16.cpp
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        bits16.h  (Formerly bits8.h)
+ * File:        bits16.cpp  (Formerly bits8.h)
```
amitdo: `Formerly bits8.h` should be replaced with `Formerly bits8.c`

amitdo:
>Trying to familiarize with the source code I came across some discrepancies between the file names in the file header comments and the actual file names. Maybe it's pedantic but I thought having the correct names might facilitate reading the code especially when using printouts.

+1 for the suggestion to fix it!

but because your patches touch too many files and for other reasons I prefer that a core developer from Google to take it.

cube/char_set.cpp
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        char_samp_enum.cpp
```
stweil: I'd omit the comment in parentheses (who cares for the former name?).

cube/char_set.h
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * File:        char_samp_enum.h
```
stweil: Same here.

stweil:
Fixing the wrong or out-dated names is good. I suggest to remove all modifications which are not related to such names and then apply this PR. Even if the PR were applied as it is, it would improve the current situation.

cube/feature_hybrid.h
```diff
@@ -19,7 +19,7 @@
 
 // The FeatureHybrid class implements a Bitmap feature extractor class. It
 // inherits from the FeatureBase class
-// This class describes the a hybrid feature vector composed by combining
```
stweil: This modification is unrelated to the commit comments.

cube/char_set.cpp
StefRe: I added these old file names just for uniformity's sake to be consistent with the other files.

cube/feature_hybrid.h
StefRe: OK, you're right. Just subconsciously corrected this obvious typo when reading the comment.

cube/char_set.cpp
stweil: Such comments might have been useful 20 years ago, but nowadays we have configuration management and can track all history in Git, so I think there is no need for such comments in the code any more. I'd prefer removing the existing comments (in a later PR of course) instead of adding more of them to reach uniformity.

zdenop:
this PR can not be merged to current code. Can you please update it?

amitdo:
I think @theraysmith should say if he is OK with these changes.
Ray, also see @stweil https://github.com/tesseract-ocr/tesseract/pull/344#issuecomment-239267937

 @jbreiden said something, but for some reason deleted his comment.

theraysmith:
The request is reasonable, but I would rather spend my limited time fixing actual bugs than up-integrating/merging such large changes (and asking colleagues to review them) that perform little real useful purpose, especially when there is a chance that a large chunk of the code could go away in the near future.

PR #351, opened 2016-06-16
Fix Cygwin compatibility

amitdo:


stweil:
I just tried to build with Cygwin and had not problems with the unpatched code.

configure.ac
```diff
@@ -94,7 +94,7 @@ case "${host_os}" in
         ;;
     cygwin*)
         AM_CONDITIONAL([ADD_RT], false)
-        AM_CONDITIONAL([T_WIN], true)
+        AM_CONDITIONAL([T_WIN], false)
```
stweil: `T_WIN` is set to `false` by default, so this line could simply be removed (if `false` is the correct value).

amitdo:
>I just tried to build with Cygwin (64 bit, tesseract + training) and had not problems with the unpatched code.

hmmm....

@matzeri?

matzeri:
On 16/06/2016 18:13, Amit Dovev wrote:
>     I just tried to build with Cygwin (64 bit, tesseract + training) and
>     had not problems with the unpatched code.
>
> hmmm....
>
> @matzeri <https://github.com/matzeri>?
>

It builds (now, it failed before) but the result is
a hybrid cygwin/windows program not a cygwin program.

Current tesseract binary

$ cygcheck /usr/bin/cygtesseract-3.dll |grep -v '^    '
E:\cygwin64\bin\cygtesseract-3.dll
   E:\cygwin64\bin\cygwin1.dll
   E:\cygwin64\bin\cyglept-5.dll
   E:\cygwin64\bin\cyggcc_s-seh-1.dll
   E:\cygwin64\bin\cygstdc++-6.dll

Building from HEAD without patches

$ cygcheck api/.libs/cygtesseract-3.dll |grep -v '^   '
E:\cyg_pub\devel\tesseract\prove2\tesseract-3.05.00dev_build\api\.libs\cygtesseract-3.dll
   E:\cygwin64\bin\cygwin1.dll
   E:\cygwin64\bin\cyglept-5.dll
   E:\cygwin64\bin\cyggcc_s-seh-1.dll
   E:\cygwin64\bin\cygstdc++-6.dll
   C:\Windows\system32\WS2_32.dll

amitdo:
@matzeri
The patch for `training/pango_font_info.cpp` is different from yours. Please have a look at it and test it.

amitdo:
@stweil
>T_WIN is set to false by default, so this line could simply be removed (if false is the correct value).

Fixed. Thanks.

amitdo:
@stweil
Please confirm that this PR compiles with Cygwin 64, including training tools.

stweil:
The build with Cygwin64 fails for the training tools:

    training/pango_font_info.cpp: In member function 'bool tesseract::PangoFontInfo::ParseFontDescription(const PangoFontDescription*)':
    training/pango_font_info.cpp:227:46: error: 'strcasestr' was not declared in this scope
       is_fraktur_ = (strcasestr(family, "Fraktur") != NULL);

stweil:
Defining `_GNU_SOURCE` fixes the build failure.

stweil:
It builds with this additional patch:

    diff --git a/training/pango_font_info.cpp b/training/pango_font_info.cpp
    index e9e6d73..8bc56a0 100644
    --- a/training/pango_font_info.cpp
    +++ b/training/pango_font_info.cpp
    @@ -25,10 +25,13 @@
     #if (defined __MINGW32__) || (defined __CYGWIN__)
     // workaround for stdlib.h and putenv
     #undef __STRICT_ANSI__
    -#endif
     
     #if (defined __MINGW32__)
     #include "strcasestr.h"
    +#else
    +/* needed for strcasestr in string.h */
    +#define _GNU_SOURCE
    +#endif
     #endif
     
     #include <stdlib.h>

amitdo:
@Shreeshrii reported that this patch https://github.com/amitdo/text2tif/commit/0473f14f, which is same as the the above patch works on Cygwin...

I'll change it anyway.

amitdo:
Something went wrong with the update. I'll try again soon.

PR #353, opened 2016-06-17
Remove redundant destructor

pnordhus:
Contrary to the comment, the destructor of the base class `GenericVector` calls `GenericVector::clear()`, which actually deletes `data_`, rendering `~DawgPositionVector()` redundant.

### DIFF
--- a/dict/dawg.h
+++ b/dict/dawg.h
@@ -368,14 +368,6 @@ struct DawgPosition {
 
 class DawgPositionVector : public GenericVector<DawgPosition> {
  public:
-  /// Overload destructor, since clear() does not delete data_[] any more.
-  ~DawgPositionVector() {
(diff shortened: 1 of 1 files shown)


amitdo:
This patch looks wrong to me. @stwell please check it too.

amitdo:
pinging @stweil

stweil:
Sorry, I'm still busy with other issues. This PR comes later.

pnordhus:
@amitdo Why do you think this patch is wrong?

amitdo:
@pnordhus

I'm sorry, I thought you were wrong by saying:
>Contrary to the comment, the destructor of the base class GenericVector calls GenericVector::clear(), which actually deletes data_, rendering ~DawgPositionVector() redundant.

But I now realize that you are right :smile:

Issue #233, opened 2016-02-19
Multi-page TIFF buffering is broken


tfmorris:
The current multi-page TIFF handling is seriously sub-optimal. First, it unnecessarily reads the entire file into memory which can tie up many MB of memory unnecessarily when pages are being processed in a streaming fashion. Second, I'm pretty sure that accessing by page number causes the entire buffer to be parsed from the beginning every time incurring both processing cost and memory thrashing.

jbreiden:
Please be careful with changes. It is very important not to break streaming support. 

https://github.com/tesseract-ocr/tesseract/wiki/FAQ#how-to-do-streaming

tfmorris:
Whoever makes the changes will have to make sure that all the tests in the test suite pass, just like any other change to the code base.

Of course this wouldn't be necessary if, when streaming support was added, it was done with some recognition of the performance impact rather than just leaving comments like "To keep code simple we will also buffer data coming from a file." https://github.com/tesseract-ocr/tesseract/blob/master/api/baseapi.cpp#L1119

amitdo:
https://groups.google.com/forum/?hl=en#!topic/tesseract-dev/LMo_igM4z90

jbreiden:
I plead guilty on all counts. I wrote the streaming feature, I failed to write a test, I introduced the performance regression, I wrote the comment, and I didn't seriously consider the importance of the performance regression.  I will attempt to write a test for streaming, and will work with you on TIFF. 

TIFF is historically tricky for two reasons. One is the duplicated functionality on the OpenCL path. Hard to synchronize and hard to test. I personally can't seem to run the OpenCL path at all without a segfault. Second, under win32 is it hard or impossible to pass a file descriptor between different DLLs. Which is awkward because the libtiff API prefers to work with file descriptors instead of file pointers.

tfmorris:
@jbreiden I don't see any particular reason that your feature should be the first to have a test. As far as I can tell, there are no tests at all for the entire program. Can you point to a more complete description of the file descriptor issue? Does TIFFClientOpen help at all?

Some random, possibly relevant, tidbits:
- "TIFF is not a streamable format" - http://www.awaresystems.be/imaging/tiff/faq.html#q7
- the "streaming" support doesn't stream - at least for anything other than a list of file names. All other formats are read entirely into memory before any processing starts. That's a pretty narrow definition of streaming.
- file format sniffing requires 12 bytes, no more
[...]

Issue #209, opened 2016-02-04
Binaries for windows


VanKurt:
Where can I download Tesseract for Windows?
All Links point to code.google.com, but since Tesseract has moved here all links are dead.

Also I would like to downoad different versions (2.x, 3.x) to see if any of them works.
The version I have right now (3.05 Setup) does not read any images but produces errors instead.

stweil:
That might be caused by missing permissions: all but the latest versions try to create temporary files in the root directory c:\.(that's caused by the code coming from Leptonica).

egorpugin:
Leptonica can be built on windows. But by default it is built without any image libraries (i.e. cannot load/save images). So, no png,gif,tiff functions available.
In such mode tesseract will only handle raw image input from api, not any exe calls.

VanKurt:
OK, thanks for the replies. Can I download a version somewhere that
a) Does not have the bug mentioned by stweil
b) Includes the image loading functionality

On https://github.com/tesseract-ocr/tesseract/wiki/Downloads it says "Currently, there is no official Windows installer for newer versions." :-(
The 3.05 setup from https://github.com/UB-Mannheim/tesseract/wiki does also not work (as mentioned above).

egorpugin:
Try these binaries compiled by me.
https://dl.dropboxusercontent.com/u/19622060/tesseract-3.05.00dev-win32-vc19.zip
You have to install VC2015 x86 redist from microsoft.com

amitdo:
> Also I would like to downoad different versions (2.x, 3.x) to see if any of them works.

http://sourceforge.net/projects/tesseract-ocr-alt/files/

bantilan:
Hi, I made a step by step guide how to compile Tesseract on Windows.. You can check my guide's here: http://gensanblog.com/2016/01/28/how-to-compile-tesseract-git-project-on-windows/

You can also download the compiled version..

zdenop:
@bantilan: please add it https://github.com/tesseract-ocr/tesseract/wiki/Compiling#windows (every github used can modify it)

egorpugin:
It's a bad guide. Do not add it anywhere.
1. Windows binaries should be made with MSVC, not cygwin. Why do you install Visual Studio and do not use it? You use only cygwin stuff. Which won't run if you distribute it without cygwin.
[...]

PR #367, opened 2016-07-16
win32: Show TIFF warnings on console

stweil:
Showing them in a window (default) is not acceptable for a console
application like Tesseract which must be able to work in batch mode.

### DIFF
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -33,6 +33,23 @@
 #include "openclwrapper.h"
 #include "osdetect.h"
 
+#if defined(_WIN32)
+
+#include <tiffio.h>
+#include <windows.h>
+
+static void Win32WarningHandler(const char* module, const char* fmt,
+                                va_list ap) {
+    if (module != NULL) {
+        fprintf(stderr, "%s: ", module);
+    }
+    fprintf(stderr, "Warning, ");
+    vfprintf(stderr, fmt, ap);
+    fprintf(stderr, ".\n");
+}
+
+#endif /* _WIN32 */
+
 void PrintVersionInfo() {
     char *versionStrP;
 
@@ -351,6 +368,11 @@ int main(int argc, char **argv) {
   GenericVector<STRING> vars_vec, vars_values;
   int arg_i = 1;
   tesseract::PageSegMode pagesegmode = tesseract::PSM_AUTO;
+
+#if defined(_WIN32)
+  /* Show libtiff warnings on console (not in GUI). */
+  TIFFSetWarningHandler(Win32WarningHandler);
+#endif /* _WIN32 */
 
   ParseArgs(argc, argv,
           &lang, &image, &outputbase, &datapath,
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -33,7 +33,7 @@
 #include "openclwrapper.h"
 #include "osdetect.h"
 
-#if defined(_WIN32)
+#if defined(HAVE_TIFFIO_H) && defined(_WIN32)
 
 #include <tiffio.h>
 #include <windows.h>
@@ -48,7 +48,7 @@ static void Win32WarningHandler(const char* module, const char* fmt,
     fprintf(stderr, ".\n");
 }
 
-#endif /* _WIN32 */
+#endif /* HAVE_TIFFIO_H &&  _WIN32 */
 
 void PrintVersionInfo() {
     char *versionStrP;
@@ -369,10 +369,10 @@ int main(int argc, char **argv) {
   int arg_i = 1;
   tesseract::PageSegMode pagesegmode = tesseract::PSM_AUTO;
 
-#if defined(_WIN32)
+#if defined(HAVE_TIFFIO_H) && defined(_WIN32)
   /* Show libtiff warnings on console (not in GUI). */
   TIFFSetWarningHandler(Win32WarningHandler);
-#endif /* _WIN32 */
+#endif /* HAVE_TIFFIO_H &&  _WIN32 */
 
   ParseArgs(argc, argv,
           &lang, &image, &outputbase, &datapath,


egorpugin:
TIFF gui warnings probably should be turned off during tiff library compilation with definition TIF_PLATFORM_CONSOLE.

stweil:
All precompiled TIFF libraries for Windows which I know (e .g. the mingw64-i686-tiff which is part of cygwin) don't define TIF_PLATFORM_CONSOLE.

I don't think that it would be a good idea to require using a special TIFF library for Tesseract.

So yes, in theory your suggestion is an alternative solution, but it is not feasible in practice.

egorpugin:
Your commit introduces dependency on tiff library for tesseract.
Now by default tesseract uses only leptonica library.

jbreiden:
I think there is another bug about better memory usage with multipage tiff, where the solution is also a tiff library dependency. Will try to find it.

jbreiden:
Yes, it was bug #223. There is an enhancement request to improve performance on multipage tiff reads. It is currently slow to read the latter images in multipage tiff. The only way to accomplish this is with a direct libtiff dependency. I didn't implement because I wasn't sure if it was worth it.

jbreiden:
I mean bug #233

stweil:
Some additional notes on the problem which is addressed by this PR:

It looks like many TIFF files (nearly all?) include some vendor specific data which trigger warnings from libtiff, so it is a real and very common problem for Windows users who want to do batch processing. I see these alternatives to handle the problem:

1. Change libtiff. Instead of defaulting to a message window for all warnings, it could do so for GUI applications but use stderr for console applications.

2. Change leptonica. It could offer an interface for setting the handler for warnings, and tesseract could use that interface to set the handler.  Or leptonica could set a reasonable default for GUI / console applications.

3. Change tesseract. That's what I did in my PR.

IMHO in the long run the first alternative would be the best solution, but i see no chance to get it quickly.

stweil:
@egorpugin, I added a commit which handles the libtiff dependency, so the code will not break if tiffio.h is unavailable.

egorpugin:
Yes, but the code won't be active almost always. Users should care about tiff library, download it, deploy, link to 'tesseractmain' or their own binary.
Also what is HAVE_TIFFIO_H? It is not defined in tesseract build system, so it will be turned off.

Better solution I provided when building tesseract with CPPAN.
See:
https://github.com/tesseract-ocr/tesseract/issues/209 (bottom comments)
https://github.com/tesseract-ocr/tesseract/wiki/Compiling#master-branch-305-and-later
and
https://github.com/DanBloomberg/leptonica/blob/master/cppan.yml#L62 this line will disable gui notifications from tiff library completely.

stweil:
HAVE_TIFFIO_H is set by the current tesseract code – both by configure and by cmake (I tested both variants before adding the last commit).

egorpugin:
Ah, I see now, sorry.
Well, it's ok then.

jbreiden:
Wait a second. Does this mean it is okay for me to use direct calls to libtiff if I 
put it inside HAVE_TIFFIO_H? If so, I can go make the multipage TIFF reader
vastly faster.

zdenop:
yes

egorpugin:
My vote for removing HAVE_TIFFIO_H (and removing this commit).
It is not used anywhere in tesseract except this commit.
Tesseract uses leptonica API, so maybe it's better to contribute there and use its API later.

egorpugin:
It seems HAVE_TIFFIO_H was a legacy check, that was automatically imported to cmake to save compatibility with autotools build. Now I see that it's not used.

jbreiden:
Well, this is a core decision. If and only if allow a direct TIFF dependency
for Tesseract, then I am going to use it to significantly speed up OCR of 
multipage TIFF as part of bug #233.

egorpugin:
Is it possible via leptonica API? I see some comments from Dan B. there.

jbreiden:
Yeah, Dan and I talked about this extensively and eventually came to agreement. It does not seem possible through Leptonica, even if we modify Leptonica. The issue is there is a libtiff data structure that must be held in memory, and must be iterated through to get successive images in multipage TIFF. Leptonica doesn't have the ability to keep this data structure around, because Leptonica is functional C program. So you can't stash it as a member variable as a Leptonica class or something like that. And we're sure as heck not going to put it in a global because that isn't threadsafe. Which means that, currently, Leptonica has to start over from the beginning every time. When someone asks image 2348 in a multipage TIFF, we have to start at the beginning and do 2348 seeks to get to data.

jbreiden:
I suppose it is possible we are wrong, that's just current thinking. I'll look one last time, starting ehre. http://www.libtiff.org/libtiff.html#Dirs

zdenop:
@egorpugin: libtiff is needed for [tesseract opencl](https://github.com/tesseract-ocr/tesseract/blob/master/opencl/openclwrapper.h#L5) implementation. So you need to remove much more than this commit.

egorpugin:
It's strange a bit. Does OpenCL really need this? I.e. leptonica can not provide required interfaces? Or opencl is only implemented for tiff images?

Also with CPPAN build I mentioned above tiff dependency can be added in the simplest possible way (add one line to cppan.yml).

---

I'm trying to understand what do you need to get custom pages (images, directories) from tiff image. Looks like it's `TIFFSetDirectory(TIFF *tif, int dir_n)` [1,2] call. It reads image file from the beginning, iterates over each directory until requested dir number and reads it (`TIFFReadDirectory(tif)` [3]). Then it can be read by leptonica with `pixReadFromTiffStream(tif)` like here [4]. I do not understand why leptonica has one more cycle here [5]. It gives us O(N^2) complexity instead O(N), isn't it?

So, for example if we want to read custom dirs (images) from tiff via leptonica, the complexity is O(N). But if we want to read tiff sequentially we can add a new function to lept. For example, `pixReadStreamTiffNext()` which will iterate to the next dir (image) in the tiff stream via one `TIFFReadDirectory()` call because it increases `tif->tif_nextdiroff` offset.

Correct me if I'm wrong somewhere.

[1] http://www.libtiff.org/man/TIFFSetDirectory.3t.html
[2] https://github.com/vadz/libtiff/blob/master/libtiff/tif_dir.c#L1531
[3] http://www.libtiff.org/man/TIFFReadDirectory.3t.html
[4] https://github.com/DanBloomberg/leptonica/blob/master/src/tiffio.c#L375
[5] https://github.com/DanBloomberg/leptonica/blob/master/src/tiffio.c#L409

jbreiden:
The OpenCl code (amongst other things) provides hardware accelerated alternatives to a very small subset of Leptonica calls.

You are right,  Leptonica's pixReadStreamTiff() is adding lots of unnecessary overhead. The only reason we didn't notice it was all this stuff gets cached in memory, hence the massive amount of seeking is a little bit hidden. I already have a patch out to Dan to call TIFFSetDirectory() only once.  However, even a single call to TIFFSetDirectory() itself is expensive. 

Your suggestion about calling TIFFReadDirectory(TIFF *tif) to get down to linear will not work unless someone is able to hold onto the TIFF *tif struct. Tesseract can do this, but only if it is allowed to directly link to libtiff. Leptonica cannot do this, because it has nowhere persistent to hold it. Everything is completely reset for every image read,  including the call to TIFFOpen. All that said, I now see that we can - potentially - work around this by keeping track of file offsets. Need to think this over.

```c
#include <stdio.h>
#include <tiffio.h>
#include <stdint.h>
#include <stdbool.h>

const char *testfile = "test.tiff";

size_t PrimeThePump() {
  TIFF *tiff = TIFFOpen(testfile, "r");
  TIFFSetDirectory(tiff, 0);
  size_t offset = TIFFCurrentDirOffset(tiff);
  TIFFClose(tiff);
  return offset;
}

size_t ThankYouMayIHaveAnother(size_t offset) {
  TIFF *tiff = TIFFOpen(testfile, "r");
  TIFFSetSubDirectory(tiff, offset);
  TIFFReadDirectory(tiff);
  offset = TIFFCurrentDirOffset(tiff);
  TIFFClose(tiff);
  return offset;
}

int main(void) {
  size_t offset = PrimeThePump();
  while (offset = ThankYouMayIHaveAnother(offset)) {
    printf("offset=%lu\n", offset);
  }
}
```

stweil:
Alternative 1 (see my comments from yesterday) is addressed by this libtiff issue: http://bugzilla.maptools.org/show_bug.cgi?id=2571.

I don't expect that a libtiff version with that modification will be available soon.

egorpugin:
So I see the following possible solution. We need to query all offsets in tiff file atleast once. We could do this in leptonica. Also we can add a function that will read tiff image by dir offset.

Code flow:
```
// tesseract side
FILE *fp = fopen(filename, rb);
int n_dirs = 0;

// call to leptonica
// fp is opened, leptonica will call fast? TIFFClientOpen()
uint64_t *offsets = pixReadTiffOffsets(fp, &n_dirs);

// read interesting image O(1)
// fp is opened, leptonica will call fast? tif = TIFFClientOpen()
// TIFFSetSubDirectory(tif, offset)
// TIFFReadDirectory(tif)
// pixReadFromTiffStream()
PIX *image = pixReadTiffDir(fp, offsets[interesting_image_number]);

// free() offsets as C does not have C++ delete[]
free(offsets);
```

What do you think?

jbreiden:
I dunno; we almost always read all the images in a loop, so putting functions in Leptonica somewhat similar to my example would also do the trick.  This is going ultimately be Dan's decision. But also consider that it's kind of a pain to put something new in Leptonica and then make Tesseract depend on it. So for Tesseract, we may be better off doing the direct libtiff calls.

amitdo:
This is the only place we use libtiff directly now.

Can we move it to Leptonica?

CC: @Danbloomberg

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/21e739ca2e1e/api/tesseractmain.cpp#L41
https://github.com/tesseract-ocr/tesseract/blob/21e739ca2e1e/api/tesseractmain.cpp#L411

amitdo:
stweil commented on Jul 17, 2016

>Change leptonica. It could offer an interface for setting the handler for warnings, and tesseract could use that interface to set the handler. Or leptonica could set a reasonable default for GUI / console applications.

amitdo:
https://github.com/DanBloomberg/leptonica/commit/d42a51d29eb8

amitdo:
1.72.2 is here, so we may able to drop this code soon.

We need to wait to an update for Mingw-w64's Leptonica PKGBUILD.

Egor will probably update cppan very soon.

egorpugin:
Added 1..74.2, tess will automatically use it now.

Issue #380, opened 2016-08-05
peculiarities when running text2image on windows


vidiecan:
(this is more of a comment than an issue but more issues can follow and the discussion might be useful; nevertheless, it might be closed after the PR for 1. )
1. At the moment, text2image expects `fc` backend e.g,:
   https://github.com/tesseract-ocr/tesseract/blob/ba2ea39caaa791b5e5f092953057cb8ffb094a82/training/pango_font_info.cpp#L356
   but if pango is compiled with win32 support, you get the win32 font map first

```
#if defined(HAVE_CAIRO_WIN32)
  if (!backend || 0 == strcmp (backend, "win32"))
    return g_object_new (PANGO_TYPE_CAIRO_WIN32_FONT_MAP, NULL);
#endif
#if defined(HAVE_CAIRO_FREETYPE)
  if (!backend || 0 == strcmp (backend, "fc")
           || 0 == strcmp (backend, "fontconfig"))
    return g_object_new (PANGO_TYPE_CAIRO_FC_FONT_MAP, NULL);
#endif 
```

and nasty crashes follow because of the wrong reinterpret cast.

Fast Solution: specify `fc` backend
Solution: a simple patch will follow that fixes the behaviour for, at least, the most important functionality.
1. If fontconfig is linked as dll, putenv does not get propagated to fontconfig
   https://github.com/tesseract-ocr/tesseract/blob/ba2ea39caaa791b5e5f092953057cb8ffb094a82/training/pango_font_info.cpp#L151

Solution: specify it as environmental variable
1. You cannot use disk paths (e.g., c:) in `FONTCONFIG_PATH` because fontconfig strips slashes from path (FcStrCanonAbsoluteFilename) and then uses 

```
GetFullPathNameW (dirname, 0, NULL, NULL)
```

without the slash and that function, interestingly, behaves like this

```
a file name begins with only a disk designator but not the backslash after the colon, it is interpreted as a relative path to the current directory on the drive with the specified letter.
```

Solution: specify a sane directory

amitdo:
None of these issues has been solved.

At least the first one probably affects Tesseract running in MinGW and Mac.

>Fast Solution: specify fc backend

https://github.com/GNOME/pango/blob/master/pango/pangocairo-fontmap.c#L48

Something like this should be put in `text2image.cpp`:

```
#ifdef _WIN32
 putenv("PANGOCAIRO_BACKEND=fc");
#else
[...]

PR #381, opened 2016-08-05
fixes some of the windows issue with text2image, see #380

vidiecan:


zdenop:
thanks. Fixes #380

amitdo:
The dotted circle fix might solve things for Mac users also (#195).

PR #400, opened 2016-08-29
ccutil: Fix and simplify implementation of variadic macro

stweil:
The implementation for MS C did not pass the variable arguments to
tprintf.

The standard is supported since C99 / C++11, so one implementation
is sufficient.

### DIFF
--- a/ccutil/errcode.h
+++ b/ccutil/errcode.h
@@ -87,21 +87,12 @@ const ERRCODE ASSERT_FAILED = "Assert failed";
                         __FILE__, __LINE__);                            \
   }
 
-#ifdef _MSC_VER
-#define ASSERT_HOST_MSG(x, msg, ...) if (!(x))                            \
+#define ASSERT_HOST_MSG(x, ...) if (!(x))                               \
   {                                                                     \
-    tprintf(msg);                                                       \
+    tprintf(__VA_ARGS__);                                               \
     ASSERT_FAILED.error(#x, ABORT, "in file %s, line %d",               \
                         __FILE__, __LINE__);                            \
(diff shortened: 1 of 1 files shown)


amitdo:
I believe that currently Tesseract still supports older C and C++ standards. This patch will change this...

zdenop:
Next version of tesseract (4.0) will need  C++11. 
3.x branch should be maintained that way it will be possible to build with VS2010...

stweil:
VS2010 should work with the new code, see the [Microsoft documentation](https://msdn.microsoft.com/en-us/library/ms177415(v=vs.100).aspx). Please note also that the old code already used C99 / C++11 standard for all Microsoft compilers and also for Windows gcc versions which set `_MSC_VER`, but that it was buggy for all those compilers because it did not pass the variable arguments. The patch fixes this bug for Microsoft compilers.
The patch changes the situation regarding compatibility for really old compilers (neither gcc, clang, msc which all support variadic macros since a long time). I think that is only a theoretical issue and that such compilers would fail compiling Tesseract anyway.
`opencl/oclkernels.h` already contains a variadic macro using `__VA_ARGS__`.

stweil:
gcc-3.1.1 already supported the variadic macro syntax used in my patch, see the [online documentation](https://gcc.gnu.org/onlinedocs/gcc-3.1.1/cpp/Variadic-Macros.html#Variadic%20Macros). It was release in [2002](https://gcc.gnu.org/gcc-3.1/).

zdenop:
@stweil : thanks for clarification. I was able to build tesseract with this patch and VS2010.

PR #402, opened 2016-08-30
fileio: Replace assert with tprintf() and exit(1)

stweil:
Assertions are good for programming errors, but not for wrong user input.

The new code no longer needs File::ReadFileToStringOrDie, so remove that
method.


zdenop:
thanks.

PR #404, opened 2016-08-30
configure: Don't use -Wno-uninitialized for debug builds

stweil:
There is no good reason to suppress useful compiler warnings.


zdenop:
thanks

PR #406, opened 2016-08-31
vs2010: Fix implementation of strcasestr

stweil:
A haystack which is shorter than the needle resulted in negative value
for length_haystack which was forced to a very large unsigned value.

The resulting buffer overflow while reading the haystack would crash
text2image when it was called with a short font name.

### DIFF
--- a/vs2010/port/strcasestr.cpp
+++ b/vs2010/port/strcasestr.cpp
@@ -49,9 +49,14 @@ char *strcasestr(const char *haystack, const char *needle) {
      return NULL;
 
    length_needle = strlen(needle);
-   length_haystack = strlen(haystack) - length_needle + 1;
+   length_haystack = strlen(haystack);
 
-   for (i = 0; i < length_haystack; i++)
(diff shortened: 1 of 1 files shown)


zdenop:
thanks

amitdo:
https://github.com/amitdo/text2tif/blob/master/port/strcasestr.cpp
This is based on a more updated version of the EFL code. You might want to compare it with your version.

stweil:
It's nearly identical to my implementation. Instead of

    length_haystack = length_haystack - length_needle + 1;
    for (i = 0; i < length_haystack; i++)

I used the shorter

    length_haystack -= length_needle;
    for (i = 0; i <= length_haystack; i++)

amitdo:
OK :-)

amitdo:
The lesson we should learn from this is: 
If you take a piece of code from another codebase and copy it into your codebase, you need to follow the other codebase for updates and [bug fixes](https://github.com/tasn/efl/commit/b457dff840ff1e).

PR #439, opened 2016-10-08
training: Remove unnecessary const qualifiers

stweil:
This fixes several gcc warnings:

warning:
 type qualifiers ignored on function return type [-Wignored-qualifiers]


egorpugin:
true

PR #442, opened 2016-10-17
helper script to generate dawg input files from text

jimregan:


jimregan:
Wrong branch, see #443 instead

PR #446, opened 2016-10-24
Fix some resource leaks reported by Coverity

stweil:


stweil:
@rmtheis, could you please test PR #470 which should fix the crash?

PR #459, opened 2016-10-30
opencl: Add tiff library needed by openclwrapper

stweil:
openclwrapper calls function TIFFReadRGBAImageOriented which is provided
by libtiff, so add that library to OPENCL_LIBS.

This fixes a linker error (unresolved symbol) when opencl is enabled.


amitdo:
@stweil,

Thanks for your work!

amitdo:
>openclwrapper calls function TIFFReadRGBAImageOriented

Not anymore.

So, this commit should be reverted.

Same for the backport to 3.05.

PR #463, opened 2016-11-01
opencl: Move declaration of MORPH_BC from .h to .cpp file

stweil:
It is only used locally in opencl/openclwrapper.cpp.

For all other files which include openclwrapper.h, the compiler
complained about an unused static variable:

opencl/openclwrapper.h:175:16: warning:
 ‘MORPH_BC’ defined but not used [-Wunused-variable]


stweil:
It is still not clear for me why that variable `MORPH_BC` is needed at all.

zdenop:
AFAIK this should be initial support for OpenCL. Unfortunately it was not extended until this time.

PR #470, opened 2016-11-07
Fix crash caused by undefined value of local variable

stweil:
Commit b1f03cb6975ee4bde6d8c9450ea55f111cc36ac1 added a call of function
FreeFeatureSet to fix a memory leak, but introduced a new bug because the
local variable FloatFeatures was not always assigned a value.

Now FloatFeatures is always assigned a value, and we only need a single
place where FreeFeatureSet is called.


rmtheis:
Looks good to me. Thanks. In my testing, this revision fixes the crash issue.

PR #471, opened 2016-11-08
Fix a typo in tesseract(1) man page

ciupicri:
`C++` needs to escaped as `C\+\+` in the AsciiDoc source code.


amitdo:
Also, stl => STL

Issue #435, opened 2016-09-30
leptonica library with pdf support (>= 1.71) is missing


canny07:
![leptonica](https://cloud.githubusercontent.com/assets/17674215/18988851/44796cb8-8726-11e6-8854-9ff87c11ad96.png)
I followed the steps of installation of `tesseract-ocr`. After `autogen.sh` , I tried `./configure` but terminal is showing `leptonica library with pdf support (>= 1.71) is missing` and no further installation is happening.
I searched for this problem extensively. I installed `leptonica-17.3` from source. Then I tried `./configure` again. The problem persisted. I again searched. In one closed issue, there was suggestion for `./configure --with-extra-includes=/usr/local/include --with-extra-libaries=/usr/local/lib
` instead of `./configure`. I didn't work.
The problem is still the same. 
What to do? 
Any suggestion is welcome.

zdenop:
Please use tesseract user forum for asking support.

Neppord:
I have the same bug. Since it is telling you that you have the wrong version, but that is not the real issue.

 this is bug and not a "asking support".

Neppord:
@canny07 i made it work by setting the library path diffierntly. It is probably a old version of leptonica that is "blocking the sight"  for the config script .

canny07:
@Neppord Yeah It seems so. For my case, I found that the older version was the reason of trouble. So, I removed it and downloaded and installed the latest version from the source. 
Thanks for answering :)

amitdo:
First, it is recommended to uninstall an older Leptonica before install new one.

Running `sudo ldconfig` after `sudo make install` in both Leptonica and Tesseract installation should prevent the issue you had.

Anyway, it's not a bug in Tesseract.

stweil:
> Anyway, it's not a bug in Tesseract.

In some way it is. I had to fix my Tesseract configuration to get PDF support for the Windows version. Maybe I should have sent a PR earlier – I'll do it now.

stweil:
Please test and review pull request #473 which should fix this issue.

amitdo:
The cause of their issue is not related to libpng.

PR #473, opened 2016-11-10
configure: Fix detection of PDF support in leptonica

stweil:
The check needs libpng, otherwise it always fails with unresolved symbols.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -442,7 +442,8 @@ done
 if test "$have_lept" = yes ; then
   AC_MSG_RESULT(yes)
   AC_CHECK_LIB([lept], [l_generateCIDataForPdf], [],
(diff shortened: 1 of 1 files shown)


stweil:
This should fix issue #435.

zdenop:
AC_CHECK_LIB macro tests whether the library (leptonica) is providing function (l_generateCIDataForPdf). if l_generateCIDataForPdf is not present in leptonica configure will fail (because tesseract compilation would fail for the same reason). So adding linking to libpng is IMO wrong - png library is not required by tesseract. configure is just testing if function l_generateCIDataForPdf is present (e.g. we have correct version of leptonica) or not (we have wrong version of leptonica).
And this is not fixing issue #435 - if you use wrong (old) library it does not matter if libpng is present or not.

PR #496, opened 2016-11-23
Fix broken cmake builds

stweil:
Cmake builds need the additional directories `arch` and `lstm`
to find all include files and also for linking.


stweil:
This PR also fixes Travis CI builds. AppVeyor builds are still broken for the MS compiler.

Issue #493, opened 2016-11-22
Build broken in mingw64-i686 (32-bits)


wiiaboo:
`_mm256_extract_epi64` isn't declared for 32-bit.

```
$ ../configure --host=i686-w64-mingw32 --build=i686-w64-mingw32 --target=i686-w64-mingw32 --prefix=/local32 --disable-shared --enable-static --bindir=/local32/bin-global --disable-graphics
$ make
make  all-recursive
make[1]: Entering directory '/build/tesseract-git/build-32bit'
Making all in arch
make[2]: Entering directory '/build/tesseract-git/build-32bit/arch'
make[3]: Entering directory '/build/tesseract-git/build-32bit/arch'
/bin/sh ../libtool  --tag=CXX   --mode=compile i686-w64-mingw32-g++ -DHAVE_CONFIG_H -I. -I../../arch -I..  -O2 -DNDEBUG -I../../ccutil  -D_FORTIFY_SOURCE=2 -D__USE_MINGW_ANSI_STDIO=1 -I/local32/include/leptonica -mavx -mthreads -mtune=generic -O2 -pipe -std=c++11 -MT libtesseract_avx_la-dotproductavx.lo -MD -MP -MF .deps/libtesseract_avx_la-dotproductavx.Tpo -c -o libtesseract_avx_la-dotproductavx.lo `test -f 'dotproductavx.cpp' || echo '../../arch/'`dotproductavx.cpp
/bin/sh ../libtool  --tag=CXX   --mode=compile i686-w64-mingw32-g++ -DHAVE_CONFIG_H -I. -I../../arch -I..  -O2 -DNDEBUG -I../../ccutil  -D_FORTIFY_SOURCE=2 -D__USE_MINGW_ANSI_STDIO=1 -I/local32/include/leptonica -msse4.1 -mthreads -mtune=generic -O2 -pipe -std=c++11 -MT libtesseract_sse_la-dotproductsse.lo -MD -MP -MF .deps/libtesseract_sse_la-dotproductsse.Tpo -c -o libtesseract_sse_la-dotproductsse.lo `test -f 'dotproductsse.cpp' || echo '../../arch/'`dotproductsse.cpp
[...]

PR #498, opened 2016-11-24
Fix 32 bit builds (missing _mm256_extract_epi64)

stweil:


stweil:
This fixes issue #493, but also builds on 32 bit Linux.

PR #506, opened 2016-11-28
lstm: Fix issues reported by Coverity scan

stweil:

### DIFF
--- a/lstm/lstmtrainer.cpp
+++ b/lstm/lstmtrainer.cpp
@@ -1201,6 +1201,9 @@ double LSTMTrainer::ComputeCharError(const GenericVector<int>& truth_str,
   for (int i = 0; i < label_counts.size(); ++i) {
     char_errors += abs(label_counts[i]);
   }
+  if (truth_size == 0) {
+    return (char_errors == 0) ? 0.0 : 1.0;
+  }
   return static_cast<double>(char_errors) / truth_size;
 }
 
--- a/lstm/lstmtrainer.cpp
+++ b/lstm/lstmtrainer.cpp
@@ -147,7 +147,7 @@ void LSTMTrainer::InitCharSet(const UNICHARSET& unicharset,
 // ctc_mode controls how the truth text is mapped to the network targets.
 // Note: Call before InitNetwork!
 void LSTMTrainer::InitCharSet(const UNICHARSET& unicharset,
-                              const UnicharCompress recoder) {
+                              const UnicharCompress& recoder) {
   EmptyConstructor();
   int flags = TF_COMPRESS_UNICHARSET;
   training_flags_ = static_cast<TrainingFlags>(flags);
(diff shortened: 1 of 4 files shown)


lstm/lstmtrainer.cpp
```diff
@@ -149,7 +149,7 @@ void LSTMTrainer::InitCharSet(const UNICHARSET& unicharset,
 // ctc_mode controls how the truth text is mapped to the network targets.
 // Note: Call before InitNetwork!
 void LSTMTrainer::InitCharSet(const UNICHARSET& unicharset,
-                              const UnicharCompress recoder) {
+                              const UnicharCompress& recoder) {
```
amitdo: `UnicharCompress` can be `NULL`.

In the same file you have this:
`LSTMTrainer::EncodeString(const STRING& str, const UNICHARSET& unicharset, ...`
and
`if (recoder != NULL) {`

stweil:
That's a different function which either gets called with `&recoder_` or with `NULL` as value for parameter `recoder`. It is unrelated to my modification.

amitdo:
My point is that recoder may be NULL and you should use a pointer instead of a reference.

stweil:
I don't think so. The old code passed an `UnicharCompress` object. Passing a reference is the optimized equivalent. How would you assign a NULL pointer to a `UnicharCompress` object (not a pointer) as it is done in the function?

amitdo:
I will recheck it after a good sleep :)

amitdo:
Stefan,
I tried to examine the call hierarchy, but I didn't find a call to this variant of `InitCharSet` overloaded method.

amitdo:
@stweil?

stweil:
Your analysis is correct: one of the `InitCharSet` methods is unused and can be removed.

@theraysmith, are there plans to use that interface in the future, or can `void InitCharSet(const UNICHARSET& unicharset, const UnicharCompress recoder)` be removed?

@zdenop, I think this PR can be applied anyway – we can also remove a fixed method in a later PR.

theraysmith:
The alternative variant of InitCharset is used internally at Google to save accessing the script-dir directory from a trainer running in a Google data center. It is easier to pack the unicharset and recoder in a proto buffer in a separate step. If anything the other variant could prepare the recoder and then call the one that takes a recoder.

stweil:
I rebased the PR now to fix a merge conflict in `lstm/lstmtrainer.cpp`.

stweil:
Continuous integration fails because of missing `tesseract::LSTMTester::LSTMTester(long)`. This is unrelated to this PR.

stweil:
I rebased to fix the build with CMake in the continuous integration.

amitdo:
TIP: You can also close the PR and reopen it a the CI builds will rerun.

stweil:
Thanks, I knew that. In this case the trick would not have worked because the CI problem was only fixed in git master.

amitdo:
You are right :)

Issue #97, opened 2015-09-16
Remove unnecessary null pointer checks


elfring:
[Extra null pointer checks are not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null) like the following.
- [BeamSearch::Cleanup](https://github.com/tesseract-ocr/tesseract/blob/5378679dce5b9604e56657a466800db5fb6a390d/cube/beam_search.cpp#L35)
- [Bmp8::FreeBmpBuffer](https://github.com/tesseract-ocr/tesseract/blob/10ea4f0636dc997054bda3a0cd463d39e1e13a8d/cube/bmp_8.cpp#L48)
- [CharClassifier](https://github.com/tesseract-ocr/tesseract/blob/10ea4f0636dc997054bda3a0cd463d39e1e13a8d/cube/classifier_base.h#L49)
- [CubeObject::Cleanup](https://github.com/tesseract-ocr/tesseract/blob/524a61452d2cc1ed152ebd7e759713ec68e1e953/cube/cube_object.cpp#L55)
- [CubeRecoContext](https://github.com/tesseract-ocr/tesseract/blob/524a61452d2cc1ed152ebd7e759713ec68e1e953/ccmain/cube_reco_context.cpp#L57)

amitdo:
Fixed.

PR #507, opened 2016-11-28
cube: Simplify delete operations

stweil:
It is not necessary to check for null pointers.

Remove also unneeded delete operations and add missing delete operations
in cube/bmp_8.cpp.

Simplify also a conditional statement in cube/cube_object.cpp.

### DIFF
--- a/cube/bmp_8.cpp
+++ b/cube/bmp_8.cpp
@@ -48,18 +48,14 @@ Bmp8::~Bmp8() {
 // free buffer
 void Bmp8::FreeBmpBuffer(unsigned char **buff) {
   if (buff != NULL) {
-    if (buff[0] != NULL) {
-      delete []buff[0];
-    }
+    delete []buff[0];
     delete []buff;
   }
 }
 
 void Bmp8::FreeBmpBuffer(unsigned int **buff) {
   if (buff != NULL) {
-    if (buff[0] != NULL) {
-      delete []buff[0];
-    }
+    delete []buff[0];
     delete []buff;
   }
 }
@@ -77,7 +73,6 @@ unsigned char **Bmp8::CreateBmpBuffer(unsigned char init_val) {
 
   buff = (unsigned char **) new unsigned char *[hgt_ * sizeof(*buff)];
   if (!buff) {
-    delete []buff;
     return NULL;
   }
 
@@ -85,6 +80,7 @@ unsigned char **Bmp8::CreateBmpBuffer(unsigned char init_val) {
   buff[0] = (unsigned char *)
       new unsigned char[stride_ * hgt_ * sizeof(*buff[0])];
   if (!buff[0]) {
+    delete []buff;
     return NULL;
   }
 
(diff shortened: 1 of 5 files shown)


stweil:
This fixes issue #97.

cube/bmp_8.cpp
```diff
@@ -77,14 +73,14 @@ unsigned char **Bmp8::CreateBmpBuffer(unsigned char init_val) {
 
   buff = (unsigned char **) new unsigned char *[hgt_ * sizeof(*buff)];
   if (!buff) {
```
elfring: I find this null pointer check after the use of the C++ new operator questionable.
Can it be omitted here?

elfring:
Thanks for [your small source code improvement](https://github.com/tesseract-ocr/tesseract/pull/507/commits/c46e773758f1940d8bc4733ee6d119cc60d5caaa "Simplify delete operations").

cube/bmp_8.cpp
stweil: Yes, indeed. Do you want to send a pull request?

cube/bmp_8.cpp
```diff
@@ -77,14 +73,14 @@ unsigned char **Bmp8::CreateBmpBuffer(unsigned char init_val) {
...
-    delete []buff;
     return NULL;
   }
 
   // alloc and init memory for buffer and line buffer
   buff[0] = (unsigned char *)
       new unsigned char[stride_ * hgt_ * sizeof(*buff[0])];
   if (!buff[0]) {
```
stweil: @elfring, this test can also be removed completely.

cube/bmp_8.cpp
```diff
@@ -105,13 +101,13 @@ unsigned int ** Bmp8::CreateBmpBuffer(int wid, int hgt,
   // compute stride (align on 4 byte boundries)
   buff = (unsigned int **) new unsigned int *[hgt * sizeof(*buff)];
   if (!buff) {
-    delete []buff;
     return NULL;
   }
 
   // alloc and init memory for buffer and line buffer
   buff[0] = (unsigned int *) new unsigned int[wid * hgt * sizeof(*buff[0])];
   if (!buff[0]) {
```
stweil: Should be removed completely as well.

cube/bmp_8.cpp
elfring: * Do any special build options matter for the handling of C++ exceptions by such a software component?
* Can any smart pointers help to improve the source code further?

amitdo: The codebase doesn't have exceptions handling (Goole have a C++ policy that forbid their use).

amitdo: >Can any smart pointers help to improve the source code further?
There is a disscution about thos topic. See https://groups.google.com/forum/#!topic/tesseract-dev/mZ2IUsvWgbY

zdenop: cube will be removed from source code... I would like to keep it for 3.05 at least. 
So only bugfixes for cube make sense for me..

PR #510, opened 2016-11-28
classify: Restore previous working state

stweil:
This restores classify/adaptmatch.cpp to commit 1e60a8d71c591ab9562983778d6ce2541b75abc0
(+ two deleted empty lines).


theraysmith:
I've now fixed it properly.
When I did my original pull, all I got was the double free.
The solution that frees outside of GetAdaptiveFeatures is not as clean as freeing properly inside.

PR #515, opened 2016-11-30
Support standard parameter form --oem, --psm for tesseract executable

stweil:


stweil:
We could also consider printing a warning message if `-psm` is still used, but maybe this should not be done now.

Issue #96, opened 2015-09-16
reserved identifier violation


elfring:
I would like to point out that identifiers like "[`TESSERACT_API_BASEAPI_H__`](https://github.com/tesseract-ocr/tesseract/blob/545a0634dab4c0c168d37bf2b757f246d2aeef12/api/baseapi.h#L20)" and "[`TESSERACT_CCUTIL_PLATFORM_H__`](https://github.com/tesseract-ocr/tesseract/blob/ae6f195b1f3458fed31f8d0dbf97ac2ee9ee381b/ccutil/platform.h#L20)" [do not fit](https://www.securecoding.cert.org/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51-CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29) to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

zdenop:
Yes, you can make a pull request.

zdenop:
I did one more check with google [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint):
it ask to use API_BASEAPI_H_, while your doc suggest to use API_BASEAPI_H... 
I think we should fit to google-style (and there should be more style fixes)

amitdo:
According to their [C++ Style Guide](https://google.github.io/styleguide/cppguide.html#The__define_Guard) it should be: `TESSERACT_API_BASEAPI_H_`

PR #531, opened 2016-12-04
Fix header file guards and replace reserved identifiers

stweil:


stweil:
This also fixes issue #96.

egorpugin:
What about moving to `#pragma once`?

amitdo:
>What about moving to #pragma once?

https://google.github.io/styleguide/cppguide.html

>Do not use #pragma once; instead use the standard Google include guards. The path in the include guards should be relative to the top of your project tree.

egorpugin:
Ah,ok. No objections in this case.

stweil:
`#pragma once` is widely supported, but not a standard, while the include guards using macros only use standard code. Modern compilers like gcc or clang know include guards, so for those compilers `#pragma once` has no performance advantage.

egorpugin:
But less to write :)

stweil:
True.

stweil:
@zdenop, I think this PR is ready to get merged. Issue #96 can be closed then.

PR #532, opened 2016-12-04
openmp: Fix build with clang++ and compilers without OpenMP support

stweil:
Builds without support for OpenMP failed with the old code. Fix this:

* Add OPENMP_CXXFLAGS for ccmain.
* Replace unconditional -fopenmp by OPENMP_CXXFLAGS for lstm.
* Always use _OPENMP for conditional compilation.
* Remove OPENMP as there is already _OPENMP.
* Include omp.h conditionally.

### DIFF
--- a/api/Makefile.am
+++ b/api/Makefile.am
@@ -88,9 +88,7 @@ tesseract_LDADD = libtesseract.la
 
 tesseract_LDFLAGS = $(OPENCL_LDFLAGS)
 
-if OPENMP
 tesseract_LDADD += $(OPENMP_CFLAGS)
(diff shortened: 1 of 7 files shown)


stweil:
I tested these three test cases on Debian GNU Linux:

* default `configure && make` (OpenMP supported, wrong compiler options for `ccmain/par_control.cpp` without PR?)
* `configure --disable-openmp && make` (OpenMP disabled, bad build without PR)
* `configure CXX=clang++-3.8 && make` (OpenMP unsupported, build fails without PR)

Please review and test.

Issue #522, opened 2016-12-02
LSTM: Warnings on console when using -oem 4 on command line on windows


Shreeshrii:
While using png files as input and using oem 4 - LSTM, tesseract gives warnings on console in command mode. oem 0, 1, 2 and 3 gives no warnings.

On Windows 10, using 4.0 Alpha binaries provided by @stweil  and 4.0 alpha traineddata 

>>tesseract hin001.png hin001-hin-3 -oem 3 -l hin
Tesseract Open Source OCR Engine v4.00.00alpha with Leptonica

>>tesseract hin001.png hin001-hin-4 -oem 4 -l hin
Tesseract Open Source OCR Engine v4.00.00alpha with Leptonica
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
...

The warnings seem to be given per line, so a page with 20 lines of text gets about 40 lines of messages.

I have not tested this in the linux environment. The warnings are probably from leptonica.

zdenop:
What is an issue?

Shreeshrii:
There are a large number of warning messages when using -oem 4, not with other oem modes.

Is there a way to eliminate these? 

C:\shree>tesseract san001.png san001 -oem 4 -l san
Tesseract Open Source OCR Engine v4.00.00alpha with Leptonica
Detected 18 diacritics
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
Warning in pixWriteMemPng: work-around: writing to a temp file
Warning in fopenReadFromMemory: work-around: writing to a temp file
[...]

PR #543, opened 2016-12-07
tesseract: Disable Leptonica messages

stweil:
Disable debugging and informational messages from Leptonica
for release builds.

### DIFF
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -388,6 +388,11 @@ int main(int argc, char** argv) {
   static GenericVector<STRING> vars_vec;
   static GenericVector<STRING> vars_values;
 
+#if !defined(DEBUG)
+  // Disable debugging and informational messages from Leptonica.
+  setMsgSeverity(L_SEVERITY_WARNING);
+#endif
+
 #if defined(HAVE_TIFFIO_H) && defined(_WIN32)
   /* Show libtiff warnings on console (not in GUI). */
   TIFFSetWarningHandler(Win32WarningHandler);


amitdo:
https://github.com/tesseract-ocr/tesseract/issues/522

stweil commented

>That would be possible, yes. But without those nagging messages, I'd never have had a look on that part of Leptonica.


 DanBloomberg commented

>Alternatively, for a bit more flexibility, you can define the environmental variable
#define LEPT_MSG_SEVERITY L_SEVERITY_WARNING
and use
setMsgSeverity(L_SEVERITY_EXTERNAL);

amitdo:
I want the above option in the release build.
@stweil?

stweil:
Using the environment variable would imply that you always get at least one message from Leptonica because the code which enables / disables warnings also writes messages. :-( That's why I decided against that feature for the moment.
Leptonica needs a modification which allows suppressing messages from setMsgSeverity, no matter whether the environment variable exists or not and no matter which message severity level is set.

amitdo:
cc: @Danbloomberg

DanBloomberg:
First, some more of the basics, for disabling at least all info and warning messages.

You can define the external env variable to L_SEVERITY_ERROR.  That will disable warnings, which I believe makes sense for tesseract.

If you set this severity level at compile time, you guarantee that it can't be over-ridden at run-time.

If you want to go further and suppress all messages, including those from the ERROR_*  macros (which are different from the informational L_* macros), define NO_CONSOLE_IO on the compile line (or use L_SEVERITY_NONE).

As for the suggestion that setMsgSeverity() tells you what it does, I can easily add a 2nd arg (debug) and only emit those messages when debug = TRUE.  Stefan, is that what you want?

DanBloomberg:
Better, I'll just comment out the comments, leaving the interface unchanged.

stweil:
Yes, I'd prefer keeping the interface unchanged and disabling the messages (at least for arguments in the valid range).

DanBloomberg:
this change is pushed to leptonica master on github

DanBloomberg:
and modified to keep the warning(error) message but not the info(OK) messages

stweil:
An unset environment variable should not result in a warning, as this is the normal case. With the latest Leptonica code this requires unnecessarily tricky code.

DanBloomberg:
I was thinking that if you try to set it to the external value, and that value is not set, it should issue a warning because the attempt failed.  I am already commenting the L_INFOs out there, and I can comment the warning out as well.

DanBloomberg:
All leptonica messages in setMsgSeverity() are now disabled, as requested.

api/tesseractmain.cpp
```diff
@@ -388,6 +388,11 @@ int main(int argc, char** argv) {
   static GenericVector<STRING> vars_vec;
   static GenericVector<STRING> vars_values;
 
+#if !defined(DEBUG)
```
egorpugin: Can we change this to `#ifndef NDEBUG` ? `NDEBUG` is more common than `DEBUG`.

stweil:
Either `DEBUG` or `NDEBUG` is set by `configure` (depending on --enable-debug). It looks like cmake currently does not set `DEBUG`. Maybe that's the reason why `api/baseapi.cpp` also tests for `_DEBUG` (set my the MS compiler).

`NDEBUG` is a standard macro which changes the behavior of `assert`. That's also the reason why it is used rather often.

If we want to separately control standard features and Tesseract features, we need both macros. Otherwise we could stick to using `NDEBUG`.

Personally I prefer keeping both macros, but I think that removing `DEBUG` would also work.

PR #544, opened 2016-12-07
Add new C API for detecting orientation and script, remove old one (4.00)

jbarlow83:
I suggest for 4.00 removing the old API entirely because it's not really usable without a lot of fiddling, and at a major version jump there are probably other API changes anyway, not that I'd know.

For 3.x I suggest cherrypicking the first commit to add the new change and leaving DetectOS() intact but making it return FALSE instead of risk corrupting memory in the caller if there is disagreement about the OSResults results structure. Personally I think it's better to gently break the caller in this case without breaking their compilation, but opinions may differ so I am leaving 3.x untouched.

### DIFF
--- a/api/baseapi.cpp
+++ b/api/baseapi.cpp
@@ -1891,31 +1891,44 @@ char* TessBaseAPI::GetUNLVText() {
   return result;
 }
 
+bool TessBaseAPI::DetectOrientationScript(int& orient_deg, float& orient_conf, std::string& script, float& script_conf) {
+  OSResults osr;
+
+  bool osd = DetectOS(&osr);
+  if (!osd) {
+    return false;
+  }
+
+  int orient_id = osr.best_result.orientation_id;
+  int script_id = osr.get_best_script(orient_id);
+  orient_conf = osr.best_result.oconfidence;
+  script_conf = osr.best_result.sconfidence;
+  const char* script_name =
+      osr.unicharset->get_script_from_script_id(script_id);
+
+  // clockwise orientation of the input image, in degrees
+  orient_deg = orient_id * 90;
+
+  script = script_name;
+  return true;
+}
+
 /**
  * The recognized text is returned as a char* which is coded
  * as UTF8 and must be freed with the delete [] operator.
  * page_number is a 0-based page index that will appear in the osd file.
  */
 char* TessBaseAPI::GetOsdText(int page_number) {
-  OSResults osr;
+  int orient_deg;
+  float orient_conf;
+  std::string script_name;
+  float script_conf;
 
-  bool osd = DetectOS(&osr);
-  if (!osd) {
+  if (!DetectOrientationScript(orient_deg, orient_conf, script_name, script_conf))
     return NULL;
-  }
-
-  int orient_id = osr.best_result.orientation_id;
-  int script_id = osr.get_best_script(orient_id);
-  float orient_conf = osr.best_result.oconfidence;
-  float script_conf = osr.best_result.sconfidence;
-  const char* script_name =
-      osr.unicharset->get_script_from_script_id(script_id);
-
-  // clockwise orientation of the input image, in degrees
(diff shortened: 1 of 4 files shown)


jbarlow83:
Python script to test the new API. Requires `pip install cffi`. Only checked on Python3.

```python

# coding: utf-8

# In[2]:

PATH_TO_LIBTESS = '/path/to/development/libtesseract.so'

import cffi


# In[12]:

ffi = cffi.FFI()
ffi.cdef("""
struct Pix;
typedef struct Pix PIX;
PIX * pixRead ( const char *filename );
char * getLeptonicaVersion (  );


typedef struct TessBaseAPI TessBaseAPI;
typedef int BOOL;

const char* TessVersion();

TessBaseAPI* TessBaseAPICreate();
int TessBaseAPIInit3(TessBaseAPI* handle, const char* datapath, const char* language);

void TessBaseAPISetImage2(TessBaseAPI* handle, struct Pix* pix);


BOOL   TessBaseAPIDetectOrientationScript(TessBaseAPI* handle, char** best_script_name, 
                                                            int* best_orientation_deg, float* script_confidence, 
                                                            float* orientation_confidence);
""")


# In[13]:

libtess = ffi.dlopen(PATH_TO_LIBTESS)

from ctypes.util import find_library
liblept = ffi.dlopen(find_library('lept'))


# In[15]:

ffi.string(libtess.TessVersion())


# In[16]:

ffi.string(liblept.getLeptonicaVersion())


# In[17]:

api = libtess.TessBaseAPICreate()


# In[18]:

libtess.TessBaseAPIInit3(api, ffi.NULL, ffi.NULL)


# In[20]:

pix = liblept.pixRead('mono.png'.encode())


# In[21]:

pix


# In[22]:

libtess.TessBaseAPISetImage2(api, pix)


# In[24]:

script_name = ffi.new('char **')
orient_deg = ffi.new('int *')
script_conf = ffi.new('float *')
orient_conf = ffi.new('float *')
libtess.TessBaseAPIDetectOrientationScript(api, script_name, orient_deg, script_conf, orient_conf)


# In[25]:

ffi.string(script_name[0])


# In[26]:

orient_deg[0], script_conf[0], orient_conf[0]


```

amitdo:
It night here and I am going to bed...
Please don't commit until I'll review it tomorrow

api/baseapi.cpp
```diff
@@ -1891,31 +1891,44 @@ char* TessBaseAPI::GetUNLVText() {
   return result;
 }
 
-/**
- * The recognized text is returned as a char* which is coded
- * as UTF8 and must be freed with the delete [] operator.
- * page_number is a 0-based page index that will appear in the osd file.
- */
-char* TessBaseAPI::GetOsdText(int page_number) {
+bool TessBaseAPI::DetectOrientationScript(int& orient_deg, float& orient_conf, std::string& script, float& script_conf) {
```
amitdo: You should use plain pointers instead of references.
This is what all(?) the other methods which are part of the API do.
https://google.github.io/styleguide/cppguide.html#Reference_Arguments

api/baseapi.h
```diff
@@ -26,6 +26,7 @@
                                             (patch))
 
 #include <stdio.h>
+#include <string>
```
amitdo: There is a #include <string.h> in `#else` block above.

api/baseapi.cpp
amitdo: also, other methods use STRING not std::string ...

amitdo: The only reason is consistency with the rest of the code.
But if you insist, it's fine by me.

Please not that I'm just a community contributor like you, so you don't **have** to take my advice.

jbarlow83:
Updated. It ended up working out more nicely to drop std::string/STRING entirely and return a pointer to constant memory.

amitdo:
Thanks @jbarlow83

PR #551, opened 2016-12-09
Remove unused code

amitdo:
Remove 'cutil/listio.cpp' and 'cutil/listio.h'. Also remove 'strsave' and 'strfree' defines from 'cutil/cutil.h'.


amitdo:
This is one part of #11.

PR #568, opened 2016-12-12
opencl: Fix runtime crash when no device was found

stweil:
The native CPU is always available and needs an initialized profile.

Signed-off-by: Stefan Weil <stefan.weil@bib.uni-mannheim.de>


stweil:
AppVeyor did not run for win64. I restart continuous integration...

PR #572, opened 2016-12-13
Address some build issues (mainly for Mingw-w64)

stweil:


amitdo:
What about using pkg-config for detecting Leptonica version?

stweil:
Yes, that could be done as well. Maybe we should require a Leptonica installations with pkg-config. That would simplify that part of the configuration a lot.

amitdo:
Leptonica provides pkg-config pc file since 1.70.

stweil:
Then let's remove the unneeded code. :-)

amitdo:
Any volunteer? :laughing:

PR #586, opened 2016-12-15
Remove more cube related code

stweil:


zdenop:
IMO support for VS2010 should be dropped in 4.00 because of  [c++11](https://msdn.microsoft.com/en-us/library/hh567368.aspx)...

stweil:
That would simplify things.

amitdo:
Don't remove the 'port' dir.

egorpugin:
That dir (`port`) should be just reorganized. VS solutions could be removed because we have cmake.

By the way, did you decide how to version 4 branch? 4.00 or 4.0.0 ?

stweil:
4.0.0 could be a start for [semantic versioning](http://semver.org/) (like it is used by many other projects).

egorpugin:
That's why I'm asking. My vote for 4.0.0.

zdenop:
There is [4.00.00alpha](https://github.com/tesseract-ocr/tesseract/blob/master/api/baseapi.h#L23) and  [4.00.00dev](https://github.com/tesseract-ocr/tesseract/blob/master/configure.ac#L9) version  info. If there is change reqeuest - create PR and ask Ray for merge...

zdenop:
@stweil: will you update this PR according comments (VS removal, port reorganized)?

egorpugin:
> If there is change request - create PR and ask Ray for merge...

Sure. Will do it if no serious objections. Also following tags must be also named as 4.0.0-smth.
https://github.com/tesseract-ocr/tesseract/releases

stweil:
@zdenop, I removed the VS related commit from this PR now. Removing VS and moving port can be handled in a new PR.

zdenop:
thanks

egorpugin:
@zdenop 
By the way, maybe you'd like to reorganize sources also?
1. move all sources to `src` dir.
2. (optional) add `include` dir with all public stuff - headers.

egorpugin:
I can do this, but someone should fix autotools after the changes.

zdenop:
@egorpugin : reorganization of sources should be discussed with Ray...

egorpugin:
Ok, I'll ask him a bit later when fixing version numbers.

PR #587, opened 2016-12-16
Remove support for VS2010

stweil:
It is dropped because of missing C++11 support.
CMake can be used instead.


stweil:
Directory `port` and some files which are still used by CMake were kept. They can be re-organized. Any suggestions for the new directory structure? Maybe `win/ports`?

PR #593, opened 2016-12-19
Switch to semantic versioning

egorpugin:
@theraysmith @zdenop 

This PR replaces all 4.00.00 version numbers to 4.0.0.
The idea is to follow semantic versioning since 4.0.0 branch. https://semver.org
Following git tags should be also named as `X.Y.Z[-name]`, e.g. `4.0.0`, `4.0.1`, `4.0.0-beta`, `4.0.0-rc.1` etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,4 +75,7 @@ kernel*.bin
 /.cppan
 /cppan
 /*.dll
+/*.lib
+/*.exe
+/*.lnk
 /win*
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,14 +27,14 @@ before_install:
   
   - if [[ $OSX ]]; then brew update; fi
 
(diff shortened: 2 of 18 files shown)


Shreeshrii:
@egorpugin Isn't it better to change to the new versioning scheme while 4.0 is still in development branch?

Does @zdenop need to change the git tag?

egorpugin:
No need to change old git tags (even 4.00.00alphas), but we need to decide versioning until release.

Shreeshrii:
@jbreiden @AlexanderP 

Should this be implemented before Ubuntu freeze?

AlexanderP:
@Shreeshrii  
I think not.

egorpugin:
This was applied with other commits/PRs.

PR #595, opened 2016-12-19
Add missing Leptonica linker flags (needed for macOS)

stweil:
This fixes a build regression caused by commit
d70f3c36635c251ac3286063c6c29d10b815a565.

### DIFF
--- a/api/Makefile.am
+++ b/api/Makefile.am
@@ -45,7 +45,7 @@ endif
 libtesseract_api_la_SOURCES = baseapi.cpp capi.cpp renderer.cpp pdfrenderer.cpp
 
 lib_LTLIBRARIES += libtesseract.la
-libtesseract_la_LDFLAGS = 
+libtesseract_la_LDFLAGS = $(LEPTONICA_LIBS)
 libtesseract_la_SOURCES =
 # Dummy C++ source to cause C++ linking.
 # see http://www.gnu.org/s/hello/manual/automake/Libtool-Convenience-Libraries.html#Libtool-Convenience-Libraries


amitdo:
>needed for macOS

I'm on Linux...

egorpugin:
Do we still need autotools?
What do you think? Maybe elaborate more on cmake?

stweil:
Personally I'm more used to autotools (> 10 years) than to cmake (< 1 year). Therefore I usually build with configure && make. But of course supporting both autotools and cmake wastes resources and requires more testing.

amitdo:
Personally, I like Meson
https://github.com/mesonbuild/meson

egorpugin:
Nice, but too heavy. It uses python.
Build system (and other technical tools) for a general-purpose language must be written in the same language or more low level language, but not higher level lang. than this.

egorpugin:
Examples:
autotools is written in C
cmake in C++
We use c++, so we need something in C++,C,ASM. :)

amitdo:
>autotools is written in C

and  perl, shell, m4 and make ...

egorpugin:
And that's not good.
CMake is C++ (with some C libs) and no other deps.

zdenop:
> And that's not good.
it depends on criteria ;-) On linux that are (almost) always installed smoothly. 
With cmake I had issues - e.g. when I build and installed some basic lib cmake on windows (AFAIR png) - cmake was not able to find it after installation. It never happens with autotools on linux. I several times started to use cmake, but...

I am glad for cmake support in tesseract because of windows, but I would keep autotools because of linux. And I am willing to maintained it ;-)

egorpugin:
> And I am willing to maintained it ;-)

Ok, not a problem.

> With cmake I had issues 

Me too. That's why I'm developing CPPAN. We already can build the same training tools as with cmake with cppan on windows. So, we have parity at the moment. And in the future I hope will be able to build the last training tool - text2image on windows.

stweil:
The [UB Mannheim Windows binaries](https://github.com/UB-Mannheim/tesseract/wiki) include `text2image.exe` and were built using autotools.

egorpugin:
They are built with gcc (mingw) which is not a big advantage.
We don't need mingw (cygwin/other linux port stuff) build. We need msvc build.

amitdo:
>in the future I hope will be able to build the last training tool - text2image on windows - which is very hard with pure cmake.

What the issues with text2image and cmake?

egorpugin:
Actually, it requires pango, cairo and some other libs. See https://github.com/tesseract-ocr/tesseract/blob/master/training/CMakeLists.txt#L265
And in total all their deps contain about 15 indirect deps.

You may notice that building deps on windows (msvc) is not so easy task.
You should find every lib, make it compile on windows, include to your project and to each other etc.

PR #613, opened 2016-12-26
Fix Leptonica version check for cmake

stweil:
* The check was missing.
* The check did not work with the given Leptonica version.


egorpugin:
Cmake works with such numbers (`1.74`) too. Let's keep it.

Issue #517, opened 2016-12-01
LSTM: Windows detection of SSE and AVX is missing


amitdo:
https://github.com/tesseract-ocr/tesseract/wiki/NeuralNetsInTesseract4.00#for-open-source-contributors

>but since Windows supports the Intel intrinsics, support is just a matter of #ifdefs around the detection function.

amitdo:
Fixed. Thanks @stweil.

PR #616, opened 2016-12-27
lstm: Update AVX / SSE support

stweil:


stweil:
This PR fixes issue #517. The detection code was tested with Mingw-w64 (which supports both the gcc and the MSVC specific code), but I cannot test builds with MSVC.

stweil:
@zdenop, https://ci.appveyor.com/project/zdenop/tesseract/build/3.05.560 still includes 3.05 in the URL.

zdenop:
@stweil I change build version format to "{branch}.{build}" (from "3.05..{build}". Lets see if it helps.

stweil:
Looks good, thank you.

amitdo:
Nice!
What about moving the SIMD detection code to a separate file?
(maybe to `arch/simddetection.h`)

stweil:
Good idea. Done in PR #618.

PR #618, opened 2016-12-27
lstm: Move class SIMDDetect to new source file and improve code

stweil:
Modify also the code to use a singleton. This simplifies the code as
no locking is needed. It also slightly improves the performance because
no check whether the architecture was tested is needed.


amitdo:
Thanks again.

PR #625, opened 2016-12-28
Move AVX / SSE messages to function PrintVersionInfo (crash fix)

stweil:
This information is not needed for normal runs, so it is sufficient
to show it on request (like versions and OpenCL information).

This also fixes a crash caused by undefined order of global constructors:

When the global variable SIMDDetect::detector is initialized before the
global variable debug_file, the first tprintf call in simddetect.cpp
crashes because of a NULL pointer in debug_file. This was only seen when
running with a shared library (libtesseract.so).

### DIFF
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -28,6 +28,7 @@
 #include "baseapi.h"
 #include "basedir.h"
 #include "renderer.h"
+#include "simddetect.h"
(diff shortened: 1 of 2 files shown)


egorpugin:
This broke windows build.

egorpugin:
More precisely - dll build. 
I've fixed it. Please, be careful with static variables.

PR #661, opened 2017-01-14
Implement SIMD detection on macOS-gcc; complain if no SIMD detection

jbarlow83:
Verified that this compiles and runs on macOS El Capitan with homebrew gcc. I don't believe clang compiles at the moment so not sure if it works for that.

Also added an #error if SIMD checking is missing. Autotools already causes configure to fail if there is no SIMD implementation, so if Makefiles are generated something is wrong.

### DIFF
--- a/arch/simddetect.cpp
+++ b/arch/simddetect.cpp
@@ -26,7 +26,7 @@
 #endif // x86 target
 
 #if defined(X86_BUILD)
-# if defined(__linux__) || defined(__MINGW32__) || defined(__APPLE__)
+# if defined(__GNUC__) || defined(__clang__) || defined(__MINGW32__) || defined(__MINGW64__) 
 #  include <cpuid.h>
 # elif defined(_WIN32)
 #  include <intrin.h>
@@ -45,7 +45,7 @@ bool SIMDDetect::sse_available_;
 // any other available SIMD equipment.
 SIMDDetect::SIMDDetect() {
 #if defined(X86_BUILD)
-# if defined(__linux__) || defined(__MINGW32__) || defined(__APPLE__)
+# if defined(__GNUC__) || defined(__clang__) || defined(__MINGW32__) || defined(__MINGW64__) 
   unsigned int eax, ebx, ecx, edx;
   if (__get_cpuid(1, &eax, &ebx, &ecx, &edx) != 0) {
     sse_available_ = (ecx & 0x00080000) != 0;
@@ -60,7 +60,7 @@ SIMDDetect::SIMDDetect() {
     avx_available_ = (cpuInfo[2] & 0x10000000) != 0;
   }
 # else
-#  error "I don't know how to test for SIMD on this platform"
+#  error "I don't know how to test for SIMD with this compiler"
 # endif
 #endif // X86_BUILD
 }
--- a/arch/simddetect.cpp
+++ b/arch/simddetect.cpp
@@ -26,7 +26,7 @@
 #endif // x86 target
 
 #if defined(X86_BUILD)
-# if defined(__GNUC__) || defined(__clang__) || defined(__MINGW32__) || defined(__MINGW64__) 
+# if defined(__GNUC__) || defined(__MINGW32__)
 #  include <cpuid.h>
 # elif defined(_WIN32)
 #  include <intrin.h>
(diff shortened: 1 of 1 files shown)


amitdo:
>Also added an #error if SIMD checking is missing.

This is not the desired behavior. There is (slow) fallback function that runs when no simd is found. This `#else` code will prevent its execution.

>Autotools already causes configure to fail if there is no SIMD implementation

Are you sure? This should not happen.

arch/simddetect.cpp
```diff
@@ -59,6 +59,8 @@ SIMDDetect::SIMDDetect() {
     sse_available_ = (cpuInfo[2] & 0x00080000) != 0;
     avx_available_ = (cpuInfo[2] & 0x10000000) != 0;
   }
+# else
```
amitdo: Suggested alternative:

```
...

// If true, then AVX has been detected.
bool SIMDDetect::avx_available_ = false;
// If true, then SSe4.1 has been detected.
bool SIMDDetect::sse_available_ = false;

...

SIMDDetect::SIMDDetect() {
...
```

amitdo:
>I don't believe clang compiles at the moment so not sure if it works for that.

If you can test with clang on macOS and report issues, if any, please do it.

jbarlow83:
I looked into what compilers provide the __get_cpuid() built-in intrinsic - GCC (including mingw) and clang. Microsoft's compiler has an equivalent named __cpuid(). It doesn't matter what OS we're running on, so after establishing we're on x86, it's better to check for the compiler than exhaustive list all platforms. I changed it to reflect this.

I believe the #error should stand. Code that lacks an implementation for a specific compiler should be a build error. This should be better than silently building an inferior tesseract. In any case only compilers outside of GCC, clang, mingw, Microsoft C++ and Intel C++ (which behaves like GCC on Linux and Microsoft on Windows) should hit the #error so anything left is quite rare.

arch/simddetect.cpp
```diff
@@ -26,7 +26,7 @@
 #endif // x86 target
 
 #if defined(X86_BUILD)
-# if defined(__linux__) || defined(__MINGW32__) || defined(__APPLE__)
+# if defined(__GNUC__) || defined(__clang__) || defined(__MINGW32__) || defined(__MINGW64__)
```
amitdo: MinGW-w64 64  defines `__MINGW32__` so ` || defined(__MINGW64__) ` is not necessary.

` (__GNUC__)` is defined not only by GCC, but also by Clang and ICC.

jbarlow83: clang also provides __get_cpuid(). Not sure about ICC.

I don't think it's a bad thing for the list to be inclusive of various compilers expected to work even if some defines are redundant. What do you think?

amitdo: Personally, I don't like redundancy in my code.

My point was that if `__GNUC__` does not have the meaning you thought it has.
It does not mean 'compiled with GCC'.

https://sourceforge.net/p/predef/wiki/Compilers

amitdo: We don't use `__MINGW64__` elsewhere in the codebase, we only use `__MINGW32__`.

jbarlow83: I'm aware of what GNUC means but it's not obvious and unfortunate that other compilers use it as well. I think  `defined(__GNUC__) || defined(__clang__)` is more explicit.

In the interest of moving forward I made these change with some comments to note the implications for Mingw64 and clang. 

Checked with gcc and clang.

amitdo:
@stweil, what are your thoughts on this?

arch/simddetect.cpp
```diff
@@ -26,7 +26,7 @@
 #endif // x86 target
 
 #if defined(X86_BUILD)
-# if defined(__linux__) || defined(__MINGW32__)
+# if defined(__GNUC__) || defined(__MINGW32__)
```
stweil: Testing `__GNUC__` is sufficient, as MinGW also uses gcc, so the test for `__MINGW32__` can be removed.

arch/simddetect.cpp
```diff
@@ -43,9 +43,11 @@ bool SIMDDetect::sse_available_;
 // Constructor.
 // Tests the architecture in a system-dependent way to detect AVX, SSE and
 // any other available SIMD equipment.
+// __GNUC__ is also defined by compilers that include GNU extensions such as clang.
+// __MINGW32__ includes 64-bit mingw
 SIMDDetect::SIMDDetect() {
 #if defined(X86_BUILD)
-# if defined(__linux__) || defined(__MINGW32__)
+# if defined(__GNUC__) || defined(__MINGW32__)
```
stweil: Here the test for `__MINGW32__` can also be removed.

stweil:
I think the patch is good and only requires a small clean-up (removing the `__MINGW32__` test).
It builds fine with clang on macOS.

I also suggest to squash all commits before applying them.

jbarlow83:
I'm stuck without a full dev environment for a few days and can't do this
easily. I support the change @stweil proposed if anyone else wants to do
it.

stweil:
@jbarlow83, is [this commit](https://github.com/stweil/tesseract/commit/fa677f1ba183c6265fb11a14328dca6b8dd34368) fine for you? If yes, I'd make a pull request for it. I kept the original date and you as the author.

jbarlow83:
Looks good. Thanks.

PR #662, opened 2017-01-16
Remove code that is no longer needed

amitdo:
The code in ccutil/hashfn.h was needed for some old compilers. Now that we support MSVC >= 2010 and compilers that have good support for C++11, we can drop this code.

As a result of this file removal, we now use:
  std::unordered_map
  std::unordered_set
  std::unique_ptr
directly in the codebase with '#include' for the needed headers.


amitdo:
>Maybe you can order the system include statements in alphabetic order.

You can send a new PR :-)

PR #667, opened 2017-01-20
Fix PDF syntax error: "XObject" instead of "/XObject" when textonly_pdf=false

jbarlow83:


amitdo:
Thanks to @jbreiden, @jbarlow83 and @zdenop.

Thanks for

amitdo:
... and @Wikinaut whuich triggered this PR with his feature request.

PR #668, opened 2017-01-21
Improve textonly_pdf parameter description

Wikinaut:
I changed the parameter description because the older text was ambiguous.


Wikinaut:
@zdenop text is now "Create PDF with only one invisible text layer"

PR #675, opened 2017-01-23
Implement SIMD detection on macOS-gcc; complain if no SIMD detection

stweil:
Change SIMD detection to instead check for compilers with __get_cpuid.
Verified that this compiles and runs on macOS El Capitan with homebrew gcc.

Also added an #error if SIMD checking is missing. Compilers that have no
SIMD should make that explicit.

[sw: Merged original commits and slightly polished the commit message]
[sw: Removed test for __MINGW32__ which is no longer needed]

### DIFF
--- a/arch/simddetect.cpp
+++ b/arch/simddetect.cpp
@@ -26,7 +26,7 @@
 #endif // x86 target
 
 #if defined(X86_BUILD)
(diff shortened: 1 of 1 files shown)


stweil:
This PR‌ supersedes PR #661.

Issue #679, opened 2017-01-26
Linking failure with --disable-graphics


voyageur:
Tested with 4.00.00alpha and current head (f566a45b30cd4dd448ec6419160c17310d898aa2) on Linux,tesseract can not be compiled if I disable scrollview:
/bin/sh ../libtool  --tag=CXX   --mode=link x86_64-pc-linux-gnu-g++  -march=native -O2 -pipe -std=c++11  -Wl,-O1 -Wl,--as-needed -o tesseract tesseract-tesseractmain.o libtesseract.la   -lrt -llept -lpthread 
```
libtool: link: x86_64-pc-linux-gnu-g++ -march=native -O2 -pipe -std=c++11 -Wl,-O1 -o .libs/tesseract tesseract-tesseractmain.o  -Wl,--as-needed ./.libs/libtesseract.so -lrt -llept -lpthread -fopenmp
./.libs/libtesseract.so: undefined reference to `ScrollView::AwaitEvent(SVEventType)'
./.libs/libtesseract.so: undefined reference to `ScrollView::Pen(ScrollView::Color)'
./.libs/libtesseract.so: undefined reference to `ScrollView::Update()'
./.libs/libtesseract.so: undefined reference to `ScrollView::SetCursor(int, int)'
./.libs/libtesseract.so: undefined reference to `ScrollView::Image(Pix*, int, int)'
./.libs/libtesseract.so: undefined reference to `ScrollView::ScrollView(char const*, int, int, int, int, int, int, bool)'
./.libs/libtesseract.so: undefined reference to `ScrollView::Clear()'
./.libs/libtesseract.so: undefined reference to `SVSync::StartThread(void* (*)(void*), void*)'
./.libs/libtesseract.so: undefined reference to `ScrollView::DrawTo(int, int)'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:605: tesseract] Error 1
```

Some of these can be fixed by including config_auto.h (mostly in lstm), but I am less sure about the SVSync::StartThread() call in ccstruct/imagedata.cpp

zdenop:
Please try master and report if problem is fixed.

voyageur:
master is OK now, thanks for the fast merge!

PR #680, opened 2017-01-26
Fix GRAPHICS_DISABLED build, issue #679

voyageur:
Include automatically generated configuration file if running autoconf


amitdo:
Thanks!

PR #700, opened 2017-02-04
Add the packaging metadata to build the tesseract snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,34 @@
+name: tesseract
+version: master
+summary: open source optical character recognition engine
+description: |
+  Tesseract has unicode (UTF-8) support, and can recognize more than 100
+  languages "out of the box". It can be trained to recognize other languages.
+  Tesseract supports various output formats: plain-text, hocr(html), pdf.
+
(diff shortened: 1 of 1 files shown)


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/tesseract
    $ cd tesseract
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(dangerous because the snap you build yourself doesn't come signed from the trusted store. Once you push it to the store, your users will just have to do $ sudo snap install tesseract)

The tessdata has to be stored in ~/snap/tesseract/common/.

If you have any questions or comments, please let me know.

elopio:
thanks for the merge @zdenop!

With a few commands you can make your first release to the store, and then maybe you could be interested in automating releases after the travis tests pass: https://snapcraft.io/docs/build-snaps/publish

Any feedback will be appreciated.

PR #703, opened 2017-02-05
Add experimental support for big endian hosts (read part only)

stweil:
This PR is not ready for getting pulled. I put it here only for further discussion. See also issue #518.

Writing training data on big endian machines is currently not addressed.


ccutil/serialis.h
```diff
@@ -69,6 +72,19 @@ class TFile {
   char* FGets(char* buffer, int buffer_size);
   // Replicates fread, returning the number of items read.
   int FRead(void* buffer, int size, int count);
+  int FRead(char* buffer, unsigned count);
```
egorpugin: Is it possible to implemente this using templates?

ccutil/tessio.h
```diff
@@ -0,0 +1,127 @@
...
+static inline bool fread(char* data, FILE* f, size_t n = 1) {
+  return fread(data, sizeof(*data), n, f) == n;
+}
+
+static inline bool fread(float* data, FILE* f, size_t n = 1) {
+  return fread(data, sizeof(*data), n, f) == n;
+}
+#else
+
+bool fread(uint8_t* data, FILE* f, size_t n = 1);
```
egorpugin: templates?

stweil:
Yes, I think it would be possible to use templates in both cases.

I did not do that because I wanted to see which cases occur in practice. Some of those cases are problematic, because they don't use portable data types: the current code reads and writes `enum` and `int` data which happen to be 32 bit on Intel, but maybe not everywhere.

PR #706, opened 2017-02-06
Remove swap parameter and related code

stweil:
Remove all code for auto detection of endianness used in data files.
This prepares the switch to data files with fixed endianness.

### DIFF
--- a/ccstruct/matrix.h
+++ b/ccstruct/matrix.h
@@ -151,29 +151,18 @@ class GENERIC_2D_ARRAY {
 
   // Reads from the given file. Returns false in case of error.
   // Only works with bitwise-serializeable types!
-  // If swap is true, assumes a big/little-endian swap is needed.
-  bool DeSerialize(bool swap, FILE* fp) {
-    if (!DeSerializeSize(swap, fp)) return false;
+  bool DeSerialize(FILE* fp) {
+    if (!DeSerializeSize(fp)) return false;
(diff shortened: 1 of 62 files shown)


stweil:
@theraysmith, this PR is ready to get pulled as soon as there is a decision to switch to fixed endianness (little endian) for the training data files.

ccstruct/matrix.h
```diff
@@ -475,26 +463,17 @@ class GENERIC_2D_ARRAY {
...
-  bool DeSerializeSize(bool swap, FILE* fp) {
+  bool DeSerializeSize(FILE* fp) {
     inT32 size1, size2;
     if (fread(&size1, sizeof(size1), 1, fp) != 1) return false;
     if (fread(&size2, sizeof(size2), 1, fp) != 1) return false;
-    if (swap) {
-      ReverseN(&size1, sizeof(size1));
-      ReverseN(&size2, sizeof(size2));
-    }
     Resize(size1, size2, empty_);
```
theraysmith: There is a malloc failure and a crash right here if you don't reverse the size to be allocated!

theraysmith:
Rejected. I don't see how this can possibly work.
See comment on matrix.h.

stweil:
I'm afraid there is a misunderstanding. This patch does not implement big endian support. It removes the current effort to support big endian as a first step for a different implementation which I think is more promising.

PR #709, opened 2017-02-07
Use portable data types

stweil:
Any variables which are read from a file must use a portable data type
with a well defined size.


stweil:
This is currently only a theoretical problem, because all relevant hosts which I know use a 32 bit `int` data type.

amitdo:
These files will be removed soon with the rest of the legacy ocr engine, unless you'll persuade Ray to keep that engine...
.

zdenop:
Lets keep it in 3.05 branch...

Issue #636, opened 2017-01-04
Need better escaping for PDF title


jbreiden:
We're going to get corrupt output if an open paren is passed as a title.
Proper escaping look like this. The leading FEFF is boilerplate that 
signifies the byte order, everything else is UTF-16BE. The title in this
case is "ru"

```
<< /CreationDate (D:20170103154208-08'00') /Producer (Tesseract 2000) /Title <FEFF00720075> >>
```

https://github.com/tesseract-ocr/tesseract/blob/master/api/pdfrenderer.cpp#L963

amitdo:
https://github.com/DanBloomberg/leptonica/commit/792db025518a
>Encode pdf title in escape 4-byte hex for safety.

jbreiden:
Dan's Leptonica change is for a different code path. The Tesseract fix has not yet been written.

jbreiden:
Fix written, under review.

jbreiden:
```c++
--- tesseract/api/pdfrenderer.cpp	2017-01-20 14:57:59.000000000 -0800
+++ tesseract/api/pdfrenderer.cpp	2017-03-30 14:58:23.000000000 -0700
@@ -174,6 +174,11 @@
 // If the font is 10 pts, nominal character width is 5 pts
 const int kCharWidth = 2;
 
+// Used for memory allocation. A codepoint must take no more than this
+// many bytes, when written in the PDF way. e.g. "<0063>" for the
+// letter 'c'
+const int kMaxBytesPerCodepoint = 20;
+
 /**********************************************************************
  * PDF Renderer interface implementation
  **********************************************************************/
@@ -303,6 +308,22 @@
   if (rise < 2.0 && 2.0 < run)
     *line_y1 = *line_y2 = (y1 + y2) / 2;
 }
+
+bool CodepointToUtf16be(int code, char *utf16) {
+  if ((code > 0xD7FF && code < 0xE000) || code > 0x10FFFF) {
+    tprintf("Dropping invalid codepoint %d\n", code);
+    return false;
+  }
+  if (code < 0x10000) {
+    snprintf(utf16, sizeof(utf16), "%04X", code);
+  } else {
+    int a = code - 0x010000;
+    int high_surrogate = (0x03FF & (a >> 10)) + 0xD800;
+    int low_surrogate = (0x03FF & a) + 0xDC00;
+    snprintf(utf16, sizeof(utf16), "%04X%04X", high_surrogate, low_surrogate);
+  }
[...]

PR #731, opened 2017-02-20
new API method for setting producer for TessPDFRenderer

mato7d5:
- enable to set up producer property in Info dict.

### DIFF
--- a/api/pdfrenderer.cpp
+++ b/api/pdfrenderer.cpp
@@ -185,6 +185,8 @@ TessPDFRenderer::TessPDFRenderer(const char *outputbase, const char *datadir,
   datadir_ = datadir;
   textonly_ = textonly;
   offsets_.push_back(0);
+  producer_ = "Tesseract ";
+  producer_.append(TESSERACT_VERSION_STR);
 }
 
 void TessPDFRenderer::AppendPDFObjectDIY(size_t objectsize) {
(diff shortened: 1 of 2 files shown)


jbreiden:
1. What is the reason for this change?  Shouldn't Tesseract get credit for its PDF output?
2. As written, this creates a new security vulnerability. See issue #636.

mato7d5:
Ok, I understand. Please reject this PR.

amitdo:
>What is the reason for this change? Shouldn't Tesseract get credit for its PDF output?

If someone embeds libtesseract in his his product 'myOwnOCR', he probably prefer that any output will use this name instead of 'Tesseract'...

Maybe it will be more fair if 'libtesseract x.yz' will be kept in the output, especially if the developer doesn't apply any change to it, or apply only minor changes.

mato7d5:
> If someone embeds libtesseract in his his product 'myOwnOCR', he probably prefer that any output will use this name instead of 'Tesseract'...

Exactly, that is my case. I use libtesseract and in"about" of my app is mentioned that's is based on tesseract. I would like to set another producer as "tesseract <version>"

jbreiden:
Aside from any credit issues, there is also a matter of compatibility. We know that 3.0.4 Tesseract PDF does better with Firefox, and 3.0.5 Tesseract PDF does better with ghostscript. Maybe 3.0.6 will finally do better with Arabic / Hebrew. That sort of information gets lost when we override Producer. That said, reasonable people can probably disagree here.

I think the escaping issue is very clear. We don't want to create opportunities for creating corrupt or malicious PDF files, and therefore issue #636 is a hard requirement.

jbreiden:
There is now a fix for issue #636 which demonstrates how to safely set a metadata field.

jbarlow83:
I'm against this as well. It's too important to be able to trace Tesseract-produced PDFs back to the version of Tesseract that produced them. Tesseract should not allow changing the /Producer field.

The /Creator field, which Tesseract does not set, is the appropriate one for this use case.

mato7d5:
OK, I think you can close/deny this PR.

PR #738, opened 2017-02-27
Support AVX for 32 bit platforms

stweil:
_mm256_extract_epi64 is not available for 32 bit platforms,
but it can be replaced by "a very simple workaround".


stweil:
I have run some tests on 32 bit Linux. AVX works, produces identical results as SSE and is a little bit faster. The "very simple workaround" introduced by commit e376ffc5fe9c87d2779bf839278e08c2c2dffe70 could be used for all platforms, but I decided to be conservative in this pull request and kept the code for 64 bit platforms.

PR #740, opened 2017-02-28
Fix reading carriage return in config files

solomennikm:


egorpugin:
Could you squash or rebase or do whatever to get a single commit with your change?

PR #752, opened 2017-03-09
Remove some unused local variables and functions

stweil:
This fixes several compiler warnings.

### DIFF
--- a/classify/trainingsampleset.cpp
+++ b/classify/trainingsampleset.cpp
@@ -507,22 +507,6 @@ bool TrainingSampleSet::DeleteableSample(const TrainingSample* sample) {
   return sample == NULL || sample->class_id() < 0;
 }
 
-static Pix* DebugSample(const UNICHARSET& unicharset,
-                        TrainingSample* sample) {
-  tprintf("\nOriginal features:\n");
-  for (int i = 0; i < sample->num_features(); ++i) {
-    sample->features()[i].print();
-  }
-  if (sample->features_are_mapped()) {
-    tprintf("\nMapped features:\n");
-    for (int i = 0; i < sample->mapped_features().size(); ++i) {
-      tprintf("%d ", sample->mapped_features()[i]);
-    }
-    tprintf("\n");
-  }
-  return sample->RenderToPix(&unicharset);
-}
-
 // Construct an array to access the samples by font,class pair.
 void TrainingSampleSet::OrganizeByFontAndClass() {
   // Font indexes are sparse, so we used a map to compact them, so we can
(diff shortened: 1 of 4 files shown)


classify/trainingsampleset.cpp
```diff
@@ -507,22 +507,6 @@ bool TrainingSampleSet::DeleteableSample(const TrainingSample* sample) {
   return sample == NULL || sample->class_id() < 0;
 }
 
-static Pix* DebugSample(const UNICHARSET& unicharset,
```
amitdo: The all 'classify' directory will most likely be removed by Ray (it is part of the old engine). There is no point in applying any change to files under this directory.

stweil:
> The whole 'classify' directory will most likely be removed by Ray (it is part of the old engine). There is no point in applying any change to files under this directory.

I disagree. Maybe that directory will be removed in the future. That's not a good reason why unused code should not be removed now. Maybe we could support a Tesseract 4 with both old and new recognizer engine at least for the next year (thus eliminating the need to support Tesseract 3 for a longer time span). I'd appreciate that scenario, but if we prevent any code maintenance of the old engine in Tesseract 4, we'll never know whether that might be possible.

See also #744 which I'd apply for the same reason. My code for big endian support (#703) also changes both recognizers.

amitdo:
If you really wants to piss @theraysmith off, then fine...

stweil:
I don't. As far as I have understood Ray, he wants a Tesseract which works really good. LSTM in Tesseract 4 is a large step forward. Nevertheless, there is obviously still some need for the old engine until we can be sure that LSTM outperforms it in all OCR applications (see issue #707). Based on my experience I'd try to keep both engines for some time to allow easy verification of the LSTM performance. In issue #518, Ray said that he wants to remove the old engine, but is open for a discussion based on facts, and he also said what would have to be done if the old engine were kept.

amitdo:
Zdenko will decide...

Peace :-)

amitdo:
@egorpugin 
Is there no good alternative to AppVeyor?

egorpugin:
I don't know.
The issue is currently in cppan. I've just found it, probably will fix it tomorrow or till Monday.

zdenop:
IMO removing unused function from directory, that _should be removed_, is not big issue.

theraysmith:
OK, I only skimmed this to get the gist.
Yes I'm open to discussion on the future of the legacy engine in 4.00 and beyond, but my aim is to prove to the community that it can be safely removed, so my favorite kind of discussion is "it doesn't work well for this image" with example.
I'm not against removal of unused code from the dead code.
What I'm not keen on is major edits to the dead code that don't remove dead code.

I have a commit coming (next week?) that fulfills a long-standing request - to be able to load from a trainneddata file in memory (or via a file-reader function pointer). This involves a major edit to the dead code, but I had to do it to keep Android portability.
This will also showcase my clean and simple fix to the endian issue that Stefan raised.
I haven't yet done the endian fix to the LSTM code, but it is *easy* now.

Issue #739, opened 2017-02-27
libtiff underlinking in 3.0x.y


mkszuba:
Hello,

Tesseract-3.04.01 (and having briefly glanced at the code, likely so does 3.05.00) does not explicitly link libtesseract against libtiff even when the former references symbols from the latter. This results in the following error when one attempts to build tesseract using a linker that dislikes implicit linking, e.g. gold:

/bin/sh ../libtool  --tag=CXX   --mode=link x86_64-pc-linux-gnu-g++  -O2 -pipe -march=native -std=c++11 -lOpenCL -Wl,-O1 -Wl,--as-needed -Wl,--as-needed -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common -O2 -pipe -march=native -o tesseract tesseract-tesseractmain.o libtesseract.la   -lrt -llept -lpthread 
libtool: link: x86_64-pc-linux-gnu-g++ -O2 -pipe -march=native -std=c++11 -Wl,-O1 -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common -O2 -pipe -march=native -o .libs/tesseract tesseract-tesseractmain.o  -lOpenCL -Wl,--as-needed ./.libs/libtesseract.so -lrt -llept -lpthread
./.libs/libtesseract.so: error: undefined reference to 'TIFFFdOpen'
./.libs/libtesseract.so: error: undefined reference to 'TIFFGetFieldDefaulted'
./.libs/libtesseract.so: error: undefined reference to 'TIFFGetField'
./.libs/libtesseract.so: error: undefined reference to 'TIFFScanlineSize'
./.libs/libtesseract.so: error: undefined reference to 'TIFFReadRGBAImageOriented'
./.libs/libtesseract.so: error: undefined reference to 'TIFFReadScanline'
./.libs/libtesseract.so: error: undefined reference to 'TIFFClientOpen'
./.libs/libtesseract.so: error: undefined reference to 'TIFFReadDirectory'
./.libs/libtesseract.so: error: undefined reference to 'TIFFCleanup'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:591: tesseract] Error 1
make[2]: Leaving directory '/tmp/tesseract-3.04.01/api'
make[1]: *** [Makefile:480: all-recursive] Error 1
make[1]: Leaving directory '/tmp/tesseract-3.04.01'
make: *** [Makefile:389: all] Error 2

[...]

PR #760, opened 2017-03-14
Backports for GRAPHICS_DISABLED build/libtiff underlinking

voyageur:
SVSync::StartThread() is used in api/ , backporting the GRAPHICS_DISABLED fix
Backport api/ libtfiff missing dependency for win32 platform

### DIFF
--- a/api/Makefile.am
+++ b/api/Makefile.am
@@ -81,6 +81,8 @@ tesseract_LDADD = libtesseract.la
 
 tesseract_LDFLAGS = $(OPENCL_LDFLAGS)
 
+tesseract_LDADD += -ltiff
+
 if T_WIN
 tesseract_LDADD += -lws2_32
 libtesseract_la_LDFLAGS += -no-undefined -Wl,--as-needed -lws2_32
--- a/viewer/svutil.cpp
+++ b/viewer/svutil.cpp
@@ -83,6 +83,27 @@ void SVMutex::Unlock() {
 #endif
 }
 
+// Create new thread.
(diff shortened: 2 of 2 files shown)


RandomDSdevel:
@voyageur:  @zdenop said in [this comment](https://github.com/tesseract-ocr/tesseract/issues/739#issuecomment-286849254) on #739 that [your second commit](https://github.com/tesseract-ocr/tesseract/pull/760/commits/7e0352f5276a68d7c58b750279273f25a868efaa) to this PR's staging branch isn't really needed for branch `3.05` since `configure` _should_ add the relevant linker flag automatically when Tesseract is being built with OpenCL support&thinsp;—&thinsp;at least, I _think_ that was his reasoning…

voyageur:
@RandomDSdevel oh you're right, this is not needed for #739. Windows builds may still need as there is a direct call to tiff library in api/ for win32.
I will update the commit message, but can drop this commit of course if not really desired for 3.05 branch

zdenop:
@voyageur: How do you build tesseract on windows? AFAIK autotools (should) do the same jobs regardless platform....

voyageur:
@zdenop I don't ;) the (outside of opencl) TIFFSetWarningHandler call at https://github.com/tesseract-ocr/tesseract/blob/3.05/api/tesseractmain.cpp#L396 is wrapped in  "defined(_WIN32)" so probably needs the -ltiff linking

zdenop:
if libtiff is needed only for windows build than it should be fixed on correct place - I did it in f30cac479d1e1f857528da57ba82977b830db87a

PR #778, opened 2017-03-21
tidy tesseract(1) adding missing options

cjmayo:
Also fix "C\\++" in the manpage. I've put that in a separate commit, not least because asciidoc is adding an additional escape character for --psm in the manpage.

### DIFF
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -169,7 +169,7 @@ void PrintHelpMessage(const char* program) {
       "  --help-oem            Show OCR Engine modes.\n"
       "  -v, --version         Show version information.\n"
       "  --list-langs          List available languages for tesseract engine.\n"
-      "  --print-parameters    Print tesseract parameters to stdout.\n";
+      "  --print-parameters    Print tesseract parameters.\n";
 
   printf("\n%s", single_options);
 }
--- a/doc/tesseract.1.asc
+++ b/doc/tesseract.1.asc
@@ -70,6 +70,14 @@ OPTIONS
 	9 = Treat the image as a single word in a circle.
 	10 = Treat the image as a single character.
 
+'--oem N'::
+	Specify OCR Engine mode. The options for *N* are:
+
+	0 = Original Tesseract only.
+	1 = Neural nets LSTM only.
+	2 = Tesseract + LSTM.
+	3 = Default, based on what is available.
+
 'configfile'::
 	The name of a config to use. A config is a plaintext file which
 	contains a list of variables and their values, one per line, with a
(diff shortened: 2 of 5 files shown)


Shreeshrii:
Options maybe different between 3.05 branch and master. Eg. --psm vs -psm, --oem etc. Also 3.05 may have support for cube engine.

It maybe useful in commits to mark if they also apply for 3.05.

doc/tesseract.1.asc
```diff
@@ -268,7 +277,7 @@ The engine was developed at Hewlett Packard Laboratories Bristol and at
 Hewlett Packard Co, Greeley Colorado between 1985 and 1994, with some more
 changes made in 1996 to port to Windows, and some C\+\+izing in 1998. A
 lot of the code was written in C, and then some more was written in C\+\+.
-The C\+\+ code makes heavy use of a list system using macros. This predates
+The C++ code makes heavy use of a list system using macros. This predates
```
stweil: What about the C++ in the preceding line?

cjmayo: I'm no asciidoc expert but the first appears to act as some kind of template. If you change the previous one as well:
```
@@ -1121,9 +1121,9 @@ on read_pattern_list().</p></div>
 <div class="sectionbody">
 <div class="paragraph"><p>The engine was developed at Hewlett Packard Laboratories Bristol and at
 Hewlett Packard Co, Greeley Colorado between 1985 and 1994, with some more
-changes made in 1996 to port to Windows, and some C++izing in 1998. A
-lot of the code was written in C, and then some more was written in C++.
-The C++ code makes heavy use of a list system using macros. This predates
+changes made in 1996 to port to Windows, and some C\+\+izing in 1998. A
+lot of the code was written in C, and then some more was written in C<code>.
+The C</code> code makes heavy use of a list system using macros. This predates
```

cjmayo:
I was just trying to create one for master. I think I got confused between 3.05, 4.00alpha and master as it was! This should now be consistent with master.

amitdo:
What's going with this PR? 
@stweil, does it need any changes?

api/tesseractmain.cpp
```diff
@@ -169,7 +169,7 @@ void PrintHelpMessage(const char* program) {
       "  --help-oem            Show OCR Engine modes.\n"
       "  -v, --version         Show version information.\n"
       "  --list-langs          List available languages for tesseract engine.\n"
-      "  --print-parameters    Print tesseract parameters to stdout.\n";
+      "  --print-parameters    Print tesseract parameters.\n";
```
stweil: ok

doc/tesseract.1.asc
stweil: `asciidoc` requires escaping all but the last C++ in a paragraph. Alternatively each `C++` could be written as `$$C++$$`.

PR #788, opened 2017-03-24
Update osdetect.cpp

ShahrukhSatti21:
//Keeping 
const int kMinCharactersToTry = 5
//Since working with text extraction from vehicle number plates needs less minimum number of characters. kMaxCharactersToTry may be //modified accordingly.

### DIFF
--- a/ccmain/osdetect.cpp
+++ b/ccmain/osdetect.cpp
@@ -33,7 +33,7 @@
 #include "tesseractclass.h"
 #include "textord.h"
 
-const int kMinCharactersToTry = 50;
+const int kMinCharactersToTry = 4;
 const int kMaxCharactersToTry = 5 * kMinCharactersToTry;
 
 const float kSizeRatioToReject = 2.0;


amitdo:
The algorithm is designed to work with the current minimum value. Changing that const value might break it. I suggest not to accept this PR without Ray's approval.

amitdo:
Another thing, Ray said he has a better OSD module (LSTM-based) that will replace the current one.

amitdo:
@theraysmith, I think this PR should be rejected, right?

theraysmith:
So far the new script id looks good.
I don't understand why you would ever want to run OSD on license plates? It makes no sense.

ShahrukhSatti21:
Hi,
I was a beginner and too immature to ask such a basic question that too as
an issue. Later, I dug deeper into tesseract and a ROI finding algorithm
and compiled an executable to extract characters on number plates. There's
no need to consider this PR as the algo might not stay as generic as now.
Thanks for considering anyway.

Issue #786, opened 2017-03-24
[OS X v10.11.6 'El Capitan'] Linkage Issues Under Homebrew


RandomDSdevel:
I'm forking this issue from #739 because the resolution of that issue has _not_ helped me fix Homebrew/homebrew-core#10380.  As logged [here](https://github.com/RandomDSdevel/Homebrew-Installation-Logs/blob/master/Tesseract/v3.05.00/Third%20Attempt/reinstall.out#L1179–L1311), my latest build attempt fails as follows:  

```
⋮
/bin/sh ../libtool  --tag=CXX   --mode=link clang++  -g -O2 -std=c++11 -version-info 3:5 -no-undefined  -L/usr/local/opt/icu4c/lib -o libtesseract.la -rpath /usr/local/Cellar/tesseract/3.05.00/lib  libtesseract_api.la ../ccmain/libtesseract_main.la ../textord/libtesseract_textord.la ../wordrec/libtesseract_wordrec.la ../classify/libtesseract_classify.la ../dict/libtesseract_dict.la ../ccstruct/libtesseract_ccstruct.la ../cutil/libtesseract_cutil.la ../viewer/libtesseract_viewer.la ../ccutil/libtesseract_ccutil.la ../opencl/libtesseract_opencl.la ../cube/libtesseract_cube.la ../neural_networks/runtime/libtesseract_neural.la -llept 
[...]

PR #808, opened 2017-04-09
opencl: Add missing TIFF library for macOS

stweil:
The OpenCL code of Tesseract uses TIFF functions, but the TIFF library
was not added to the linker flags for macOS.

This fixes builds with OpenCL on Mac.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -226,7 +226,7 @@ case "${host_os}" in
       fi
       AM_CPPFLAGS="-DUSE_OPENCL $AM_CPPFLAGS"
       OPENCL_CPPFLAGS=""
-      OPENCL_LDFLAGS="-framework OpenCL"
+      OPENCL_LDFLAGS="-framework OpenCL -ltiff"
     fi
     ;;
   *)


RandomDSdevel:
Hmm, maybe this might fix #786 after all, given what it changes, so, like I said, I'll give it a shot.  Could we please possibly also get this backported to branch `3.05`, though, @stweil?

RandomDSdevel:
Whoops, perhaps I should have addressed the second half of [my last comment](https://github.com/tesseract-ocr/tesseract/pull/808#issuecomment-292812467) to _you_, @zdenop, seeing as _you're_ in charge of backporting changesets.

stweil:
I'm not sure whether it is sufficient to backport this PR to 3.05. I just tried building 3.05 on a Mac, and at least some fixes from 4.0 for Leptonica and libicu were missing, too, to get the build started. In addition I know that OpenCL in 4.0 has been improved, and I have more patches for OpenCL in 4.0 in my working queue which will also be missing for 3.05. That's why I think that it won't be possible to maintain 3.05 in the long run given the fact that the resources of volunteers are very limited.

Shreeshrii:
@stweil 

We should at least make sure that the 3.05 release includes all the fixes/improvements made for 4.0 till now.

Please add a comment on the patches that you think should be added to 3.05 as you identify them. It will help @zdenop or others with access to backport them to 3.05.

Thank you for all your work on helping fix and streamline the code.

RandomDSdevel:
@stweil:  When 4.0 is _officially_ tagged released, Homebrew (and I, being one of its users) will then switch to it.  (Building from `HEAD` _is_ supported under the package manager, but it's _not_ a common use case.  In addition, only the latest stable releases of any given package are pre-built for distribution from Homebrew's Bintray server for those who don't wish to build from source.  Similarly, Homebrew _heavily_ prefers to _only_ use patches submitted upstream to package maintainers and contributors for review and possible/probable inclusion in the next release of said software if it can be helped at _all._)  For _now_, however, 3.05 is still the latest official minor release here upstream, so I agree with @Shreeshrii that it still makes sense to support it with additional patch releases containing fixes backported from the 4.0 release series while that's still in alpha/beta.  (That's only my opinion of the situation as a Tesseract _user_&thinsp;—&thinsp;yes, I've made a few commits, but still I consider myself relatively more of a newbie with the codebase than I suspect you are&thinsp;—; _you_, of course, are thus free to disagree with this sentiment as a more active contributor and discuss the state of things with other contributors and maintainer or maintainers&thinsp;—&thinsp;I've only really seen @zdenop active here in command, but I didn't want to leave anybody else out if they were out there…)

stweil:
I started collecting all needed commits for Tesseract 3.05 with OpenCL on macOS in pull request #809 and expect to finish that task until end of the week (sorry that I cannot spend more time on this).

RandomDSdevel:
@stweil:  Thanks, and no problem; take all the time you need if you expect it will take that long or will be delayed by other work.

amitdo:
>The OpenCL code of Tesseract uses TIFF functions

Not anymore.

So, this commit should be reverted.

Same for #809 (3.05).

stweil:
Fixed for master branch in PR #907.

Issue #557, opened 2016-12-09
Don't utilize Leptonica debug parameters or functions


jbreiden:
@theraysmith 

These two Leptonica calls have optional debugging parameters.

pixGenerateHalftoneMask
pixGenHalftoneMask

They are found in imagefind.cpp in code and comments, to support an optional
parameter called textord_tabfind_show_images

https://github.com/tesseract-ocr/tesseract/blob/master/textord/imagefind.cpp#L86

In general, the debug paths in Leptonica are not great from security perspective
and setting that debug parameter to null or false is recommended. If possible
I encourage removing the the textord_tabfind_show_images feature entirely.

In addition, there is significant use of pixDisplayWrite for debugging purposes.
This is deprecated and also a security headache, and I encourage removing
all calls to this function.

It is possible that this list is not exhaustive but it probably is. The rule of thumb 
is try to avoid debug-specific Leptonica code paths wherever possible in release builds.

DanBloomberg:
I am fairly sure that pixGenHalftoneMask() (and the wrapper pixGenerateHalftoneMask()) are the only leptonica functions in tesseract that have a debug argument.

It should be OK in the new tesseract (4.00) to use pixGenerateHalftoneMask() with NULL for the debug pixa.  It would also be nice at the same time for the debug calls to the deprecated functions pixDisplayWrite() and pixDisplayWriteFormat() were removed, because there is no reason that tesseract needs to generate these debug images.

DanBloomberg:
I believe that Ray is removing all use of temp files due to pixDisplayWrite() in tesseract.

theraysmith:
Fixed in commit a1c22fb.

PR #835, opened 2017-04-22
3.05: Backports from master branch

stweil:


stweil:
Tesseract 3.05 still includes `vs2010` (which was removed in Tesseract 4.00). The VS2010 support for Tesseract 3.05 was not updated to use Leptonica 1.74, so it won't work. Should it be removed, or is it still used? Then it has to be fixed.

amitdo:
Is there a real dependency in 3.05's code (not the build system) on Leptonica 1.74?

Shreeshrii:
See
https://github.com/tesseract-ocr/tesseract/pull/835/commits/37478e7d99c3406a29dc31d76a749592171e4703

>Since commit 245eebd
<https://github.com/tesseract-ocr/tesseract/commit/245eebdf293ac19f1fb85c36e51daaaa1b5e5a3e>,
Tesseract 3.05

requires Leptonica 1.74.



ShreeDevi

amitdo:
OK. Thanks Shree.

amitdo:
https://github.com/tesseract-ocr/tesseract/issues/739#issuecomment-283608563

stweil:
Tesseract 3.05 uses the function `pixReadFromMultipageTiff` which was added in Leptonica 1.74 unconditionally, so this is a hard dependency (not only for training or some special configurations).

amitdo:
I already got an answer for the the dependency issue.

Please give an answer to my question:
https://github.com/tesseract-ocr/tesseract/pull/835#issuecomment-296431715

Another one to consider backporting:
https://github.com/tesseract-ocr/tesseract/issues/557

stweil:
Sure, there are more commits which could be backported from the master branch. I did not add the libtiff related commits because 3.05 could be built without them.

amitdo:
#233 has been fixed in both 3.05 and 4.00, but in 3.05 the OpenCL code path is slower **for multi page tiff** than the regular code path. in 4.00 the OpenCL code path was removed.

PR #843, opened 2017-04-26
opencl: Remove unused code

stweil:


jbreiden:
I think this can be a little more aggressive. It should now be possible to remove the tiff.h and tiffio.h includes, and also the libtiff build dependency.

PR #857, opened 2017-04-30
Avoid unnecessary new / delete code

stweil:


stweil:
This is similar to pull request #352 and avoids additional new / delete operations. There remain some more cases, but those are more difficult, so they will be handled in a later pull request.

PR #860, opened 2017-05-01
Replace memalloc / memfree by C++ new / delete

stweil:


stweil:
memfree is still used – without matching memalloc, so these calls of memfree should be replaced. This must be addressed in a separate pull request.

PR #865, opened 2017-05-02
Replace Tesseract data types by POSIX data types

stweil:


stweil:
I tried to avoid directories with a high probability of merge conflicts. Ray, maybe this kind of cleanup could be done in combination with clang-tidy.

PR #868, opened 2017-05-03
Warnings are people too!

rfschtkt:
Important warnings warrant attention because they may indicate real problems.
Unimportant warnings warrant attention because they obscure the important warnings.

Unfortunately for me the nicest example of the former category, where I also immediately had a solution, was already fixed independently...

### DIFF
--- a/textord/textord.cpp
+++ b/textord/textord.cpp
@@ -285,6 +285,8 @@ void Textord::TextordPage(PageSegMode pageseg_mode, const FCOORD& reskew,
     // RAW_LINE, SINGLE_LINE, SINGLE_WORD and SINGLE_CHAR all need a single row.
     gradient = make_single_row(page_tr_, pageseg_mode != PSM_RAW_LINE,
                                to_block, to_blocks);
+  } else {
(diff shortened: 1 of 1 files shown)


rfschtkt:
In case you're wondering, this is more or less a rebase of an earlier pull request (part of it was no longer relevant), but unfortunately still on my master branch, so I may have to fix that also... unless the pull request is accepted soon. I also see that there is a new conflict.

api/pdfrenderer.cpp
```diff
@@ -952,7 +957,7 @@ bool TessPDFRenderer::EndDocumentHandler() {
   if (n >= sizeof(buf)) return false;
   AppendString(buf);
   size_t pages_objsize  = strlen(buf);
-  for (size_t i = 0; i < pages_.size(); i++) {
+  for (size_t i = 0; i < pages_.unsigned_size(); i++) {
```
rfschtkt: or "declspec(pages_.size())"?

rfschtkt:
Will try again with topic-related branch and a different approach.

PR #896, opened 2017-05-09
Too many warnings!

rfschtkt:
Important warnings warrant attention because they may indicate real problems.
Unimportant warnings warrant attention because they obscure the important warnings.

Unfortunately for me the nicest example of the former category, where I also immediately had a solution, was already fixed independently...

This replaces earlier attempts, now up to date, on its own branch, and with some revisions.

### DIFF
--- a/api/pdfrenderer.cpp
+++ b/api/pdfrenderer.cpp
@@ -670,7 +670,7 @@ bool TessPDFRenderer::BeginDocumentHandler() {
   long int size = ftell(fp);
   fseek(fp, 0, SEEK_SET);
   const std::unique_ptr</*non-const*/ char[]> buffer(new char[size]);
-  if (fread(buffer.get(), 1, size, fp) != size) {
+  if (fread(buffer.get(), 1, size, fp) != static_cast<unsigned long>(size)) {
     fclose(fp);
     return false;
   }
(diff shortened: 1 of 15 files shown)


amitdo:
>Warnings are people too!

I see that you changed the title this time... :-)

stweil:
I suggest to make smaller pull requests which focus on closely related problems. That would make reviewing and pulling them much easier. Example: Make a pull request for all changes related to `NumFeatureSets`.

For that example, the `decltype(CharDesc->NumFeatureSets) i` will work. Nevertheless I personally prefer the much simpler `uinT32 i`. Even better because we want to use POSIX instead of Tesseract data types: `uint32_t i`.

rfschtkt:
You mean like the diminutive "Use POSIX data types and macros" with only 21 commits and 243 files changed?

Yeah, at first I thought typedecl() would be less unpopular than a public GenericVector::unsigned_size(), but I went a bit overboard with it, and I let the cat out of the bag with STRING::unsigned_size() anyway, so... I've rewritten the pull request without any typedecl(), with public GenericVector::unsigned_size(), but with size_t rather than uintT32/uint32_t.

stweil:
I don't expect that pull request #878 will be applied like that, but yes, it consists of smaller commits which cover limited modifications and which can be applied individually.

rfschtkt:
Even individual #878 commits are _way_ bigger than this whole pull request (or at least the one I randomly picked). Feel free to split it up further as you like, but I don't see the point: it's already themed by kind of warning, and the larger of the two commits is only +32/−24.

rfschtkt:
Rebase...

zdenop:
@rfschtkt: there is merge a conflict in api/pdfrenderer.cpp. Can you fix it?

rfschtkt:
Rebased.

PR #897, opened 2017-05-09
RAII

rfschtkt:
Use the [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) idiom, e.g., to help prevent memory leaks.

### DIFF
--- a/api/baseapi.cpp
+++ b/api/baseapi.cpp
@@ -1980,9 +1980,9 @@ bool TessBaseAPI::AdaptToWordStr(PageSegMode mode, const char* wordstr) {
   PageSegMode current_psm = GetPageSegMode();
   SetPageSegMode(mode);
   SetVariable("classify_enable_learning", "0");
-  char* text = GetUTF8Text();
+  const std::unique_ptr<const char[]> text(GetUTF8Text());
   if (debug) {
-    tprintf("Trying to adapt \"%s\" to \"%s\"\n", text, wordstr);
+    tprintf("Trying to adapt \"%s\" to \"%s\"\n", text.get(), wordstr);
   }
   if (text != NULL) {
     PAGE_RES_IT it(page_res_);
@@ -2022,7 +2022,6 @@ bool TessBaseAPI::AdaptToWordStr(PageSegMode mode, const char* wordstr) {
       tesseract_->EnableLearning = true;
       tesseract_->LearnWord(NULL, word_res);
     }
-    delete [] text;
   } else {
     success = false;
   }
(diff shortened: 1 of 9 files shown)


stweil:
Build error: https://travis-ci.org/tesseract-ocr/tesseract/jobs/230329088#L290.

rfschtkt:
Yes, I think it's a bug in that g++ version. I had successfully built with 5.4.0. I'll try again with an explicit move constructor...

Shreeshrii:
Please use syntax which will work with all of the following.

Supported Compilers

   - GCC 4.8 and above
   - Clang 3.4 and above
   - MSVC 2015, 2017


ShreeDevi

rfschtkt:
Are they all tested? I think that it was valid C++11, though.

amitdo:
Since you change the API, I think we need Ray's approval for this PR.

rfschtkt:
If that's public API and cast in stone, maybe there should be two versions, with the backward-compatible version calling the safe version with release(). Or the leak could be fixed ad hoc, but that's not really progress.

amitdo:
I would start with fixing the leak with the current API.
If you want to do it, open a new PR.

Fixing a leak is still progress...

rfschtkt:
Maybe tomorrow... but I would still suggest to add a new version and deprecate the current one, because this is C++, and RAII is one of C++'s most useful idioms (to put it mildly). That's why I made a new thread called `RAII`, not `lookifoundaleak`...

egorpugin:
Oh, btw.
Isn't GetUTF8Text() should return C++-ish std::string? Instead of that stupid non-owning `char*` memory ptr.

rfschtkt:
Yes, but it can also return nullptr, and I hadn't yet figured out whether there's a difference with the empty string. Adding to a STRING is the same for nullptr and empty string, but maybe not other applications. And there's at least one use of release() that would then instead become a strcpy() operation (right?).

rfschtkt:
Amended without API change, awaiting checks...

amitdo:
LGTM. @stweil?

stweil:
I'll plan to run a test with Valgrind later (~ 1 day) and will add my review comment then.

amitdo:
>but I would still suggest to add a new version and deprecate the current one, because this is C++, and RAII is one of C++'s most useful idioms (to put it mildly).

The new version of this PR uses RAII technique, without an API change :)
http://stackoverflow.com/questions/29372976/is-stdunique-ptr-an-application-of-raii

rfschtkt:
There seems to be another leak in `void POLY_BLOCK::fill(ScrollView* window, ScrollView::Color colour)`, which doesn't do `delete segments;`. But I've had enough fun for now...

rfschtkt:
Well, it seemed to be leaked, from comparing with other uses, although I'm not 100% certain...

rfschtkt:
Rebase

PR #900, opened 2017-05-10
Reviewed uses of reinterpret_cast

rfschtkt:
In most cases static_cast was sufficient, and sometimes there was no need for casting at all.

### DIFF
--- a/ccmain/thresholder.cpp
+++ b/ccmain/thresholder.cpp
@@ -312,7 +312,7 @@ void ImageThresholder::ThresholdRectToPix(Pix* src_pix,
       bool white_result = true;
       for (int ch = 0; ch < num_channels; ++ch) {
         int pixel = GET_DATA_BYTE(const_cast<void*>(
-                                  reinterpret_cast<const void *>(linedata)),
+                                  static_cast<const void *>(linedata)),
                                   (x + rect_left_) * num_channels + ch);
         if (hi_values[ch] >= 0 &&
             (pixel > thresholds[ch]) == (hi_values[ch] == 0)) {
--- a/ccstruct/matrix.h
+++ b/ccstruct/matrix.h
@@ -38,7 +38,7 @@
 class BLOB_CHOICE;
 class BLOB_CHOICE_LIST;
 
-#define NOT_CLASSIFIED reinterpret_cast<BLOB_CHOICE_LIST*>(0)
+#define NOT_CLASSIFIED static_cast<BLOB_CHOICE_LIST*>(0)
 
 // A generic class to hold a 2-D matrix with entries of type T, but can also
 // act as a base class for other implementations, such as a triangular or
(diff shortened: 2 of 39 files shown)


ccmain/thresholder.cpp
```diff
@@ -312,7 +312,7 @@ void ImageThresholder::ThresholdRectToPix(Pix* src_pix,
       bool white_result = true;
       for (int ch = 0; ch < num_channels; ++ch) {
         int pixel = GET_DATA_BYTE(const_cast<void*>(
```
stweil: I think that both casts here can be removed. This also applies to all other instances of GET_DATA_BYTE. I suggest to fix that in a separate pull request (easier review).

ccstruct/matrix.h
```diff
@@ -38,7 +38,7 @@
 class BLOB_CHOICE;
 class BLOB_CHOICE_LIST;
 
-#define NOT_CLASSIFIED reinterpret_cast<BLOB_CHOICE_LIST*>(0)
+#define NOT_CLASSIFIED static_cast<BLOB_CHOICE_LIST*>(0)
```
stweil: Replace by `nullptr` (no type cast)?

classify/adaptmatch.cpp
```diff
@@ -2242,7 +2242,7 @@ void Classify::ShowBestMatchFor(int shape_id,
   tprintf("Static Shape ID: %d\n", shape_id);
   ShowMatchDisplay();
   im_.Match(ClassForClassId(PreTrainedTemplates, shape_id),
-            AllProtosOn, reinterpret_cast<BIT_VECTOR>(&config_mask),
+            AllProtosOn, &config_mask, // TODO: or reinterpret_cast<BIT_VECTOR>(&config_mask) anyway?
```
stweil: What about this TODO comment?

rfschtkt: I don't know what the issues are here, so it seemed presumptuous to just remove all traces of the cast as if I did. Currently there's no need for a cast, but then why was it there in the first place?

ccstruct/matrix.h
rfschtkt: At first sight it could probably be 0, NULL, or nullptr, with or without the cast, i.e., one of 6 possibilities. I would definitely keep the cast to let the compiler check for appropriate usage, but otherwise I'm neutral.

ccmain/thresholder.cpp
rfschtkt: I tried to keep this commit to just a(n admittedly long) list of reducing or removing reinterpret_cast, which should be used sparingly, but I'll leave any follow-up to whoever feels that urge.

amitdo:
Related:
https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-reinterpret-cast.html

ccmain/thresholder.cpp
stweil: Ok, type casts for `GET_DATA_BYTE` can be removed after this PR has been applied.

rfschtkt:
In the reference provided from there, there are also some supporting [guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md) for revising tesseract's style of downcasting, e.g., in the Network hierarchy.

PR #903, opened 2017-05-11
Fix wrong format string

stweil:
The local variable intval is of type int.

### DIFF
--- a/ccutil/params.cpp
+++ b/ccutil/params.cpp
@@ -102,7 +102,7 @@ bool ParamUtils::SetParam(const char *name, const char* value,
   IntParam *ip = FindParam<IntParam>(name, GlobalParams()->int_params,
                                      member_params->int_params);
   if (ip && ip->constraint_ok(constraint) &&
-      sscanf(value, INT32FORMAT, &intval) == 1) ip->set_value(intval);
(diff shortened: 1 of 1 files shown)


ccutil/params.cpp
```diff
@@ -102,7 +102,7 @@ bool ParamUtils::SetParam(const char *name, const char* value,
   IntParam *ip = FindParam<IntParam>(name, GlobalParams()->int_params,
                                      member_params->int_params);
   if (ip && ip->constraint_ok(constraint) &&
-      sscanf(value, INT32FORMAT, &intval) == 1) ip->set_value(intval);
+      sscanf(value, "%d", &intval) == 1) ip->set_value(intval);
```
amitdo: There are more places where INT32FORMAT is being used
https://github.com/tesseract-ocr/tesseract/search?q=INT32FORMAT

INT32FORMAT is not being used.

stweil: Yes. That other locations are technically correct (not like the one above), but my goal is to either eliminate or replace all of them.

amitdo: OK :-)

stweil: PR #904 removes 4 of those `INT32FORMAT` lines.

PR #904, opened 2017-05-11
Improve index range check

stweil:
A wrong array index must raise an assertion instead of printing an
error message and continuing program execution.

Replace also floor by floorf which can be faster.


rfschtkt:
Some random thoughts while waiting for my pull requests to be merged:

How about C++-style `static_cast<int>` instead of C-style `(int)`? Since casting to integral types rounds toward 0, you can also omit the `floorf()` if `bottom` is known to be nonnegative... unless any overhead is known to be negligible or eliminated by the compiler anyway? If you capture this in a macro or in an always-inlined function (has C++11 standardised that?), it doesn't even have to be explained every time.

It's also quite ad hoc: a separate issue, and only addressed locally.

stweil:
That's true. While looking closer on that code, I noticed that it can be more simplified. We don't need float operations here. I'll update the commit.

Shreeshrii:
re: >waiting for my pull requests to be merged:

@rfschtkt see comment by Ray at  https://github.com/tesseract-ocr/tesseract/pull/744#issuecomment-284917919

>>Yes, each commit gets reviewed by me and someone else at Google.

amitdo:
Shree, Ray does his reviews **after** Zdenko merges PR here.

rfschtkt:
@stweil Hmm, I suspect that g++ kept -Wconversion out of -Wall/-Wextra because otherwise too much existing code would have to be adapted, not because not using casts for floating-integral conversions is necessarily a good idea... just my 2 cents.

stweil:
MS VC produces 1453 warnings for Tesseract (see results from Appveyor CI), most of them caused by data type conversions. It looks like most developers don't like fixing this kind of warning, and gcc simply supports that. :-)

amitdo:
Should we make it less verbose there?

stweil:
No, I don't think so. We should make the release builds with configure / make more verbose (`gcc -Wall` is currently only used for debug builds).

egorpugin:
Those warnings must be fixed sometime. Even in leptonica from which many warnings come too.

PR #910, opened 2017-05-12
Fix GenericVector and optimize some code which used GenericVector::init_to_size

stweil:


stweil:
There might be more cases where `init_to_size` is not needed.

PR #911, opened 2017-05-12
RFC: Add initial support for traineddata files in compressed archive formats (don't merge)

stweil:
This requires libminizip-dev, so expect failures from CI.

Up to now, little endian tesseract works with the new zip format.

More work is needed for training tools and big endian support and also to maintain
compatibility with the current proprietary format.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,12 @@ addons:
     sources:
     #- ubuntu-toolchain-r-test
     packages:
+      # Try different packages for compressed archives.
+      - libarchive-dev
+      # Trusty does not support libminizip-dev (requires Zenial).
+      #libminizip-dev
+      - libzip-dev
+      - libzzip-dev
     #- g++-6
 
 #matrix:
--- a/configure.ac
+++ b/configure.ac
@@ -422,6 +422,30 @@ else
   AC_MSG_ERROR([Leptonica 1.74 or higher is required. Try to install libleptonica-dev package.])
 fi
 
+PKG_CHECK_MODULES([libarchive], [libarchive], [have_libarchive=true], [have_libarchive=false])
+AM_CONDITIONAL([HAVE_LIBARCHIVE], [$have_libarchive])
+if $have_libarchive; then
+  AC_DEFINE([HAVE_LIBARCHIVE], [], [Enable libarchive])
+fi
+
+PKG_CHECK_MODULES([libzip], [libzip], [have_libzip=true], [have_libzip=false])
+AM_CONDITIONAL([HAVE_LIBZIP], [$have_libzip])
+if $have_libzip; then
+  AC_DEFINE([HAVE_LIBZIP], [], [Enable libzip])
+fi
+
+PKG_CHECK_MODULES([minizip], [minizip], [have_minizip=true], [have_minizip=false])
+AM_CONDITIONAL([HAVE_MINIZIP], [$have_minizip])
+if $have_minizip; then
+  AC_DEFINE([HAVE_MINIZIP], [], [Enable minizip])
+fi
+
+PKG_CHECK_MODULES([zziplib], [zziplib], [have_zziplib=true], [have_zziplib=false])
+AM_CONDITIONAL([HAVE_ZZIPLIB], $have_zziplib)
+if $have_zziplib; then
+  AC_DEFINE([HAVE_ZZIPLIB], [], [Enable zziplib])
+fi
+
 AM_CONDITIONAL([ENABLE_TRAINING], true)
 
 # Check availability of ICU packages.
--- a/src/api/Makefile.am
+++ b/src/api/Makefile.am
@@ -88,6 +88,17 @@ tesseract_LDFLAGS = $(OPENCL_LDFLAGS)
 
 tesseract_LDADD += $(LEPTONICA_LIBS)
 tesseract_LDADD += $(OPENMP_CXXFLAGS)
+if HAVE_LIBARCHIVE
+tesseract_LDADD += $(libarchive_LIBS)
+else
+if HAVE_LIBZIP
+tesseract_LDADD += $(libzip_LIBS)
+else
+if HAVE_MINIZIP
+tesseract_LDADD += $(minizip_LIBS)
+endif
+endif
+endif
 
 if T_WIN
 tesseract_LDADD += -ltiff
--- a/src/api/Makefile.am
+++ b/src/api/Makefile.am
@@ -96,6 +96,10 @@ tesseract_LDADD += $(libzip_LIBS)
 else
 if HAVE_MINIZIP
 tesseract_LDADD += $(minizip_LIBS)
+else
+if HAVE_ZZIPLIB
+tesseract_LDADD += $(zziplib_LIBS)
+endif
 endif
 endif
 endif
--- a/src/api/Makefile.am
+++ b/src/api/Makefile.am
@@ -90,18 +90,15 @@ tesseract_LDADD += $(LEPTONICA_LIBS)
 tesseract_LDADD += $(OPENMP_CXXFLAGS)
 if HAVE_LIBARCHIVE
 tesseract_LDADD += $(libarchive_LIBS)
-else
+endif
 if HAVE_LIBZIP
 tesseract_LDADD += $(libzip_LIBS)
-else
+endif
 if HAVE_MINIZIP
 tesseract_LDADD += $(minizip_LIBS)
-else
+endif
 if HAVE_ZZIPLIB
 tesseract_LDADD += $(zziplib_LIBS)
-endif
-endif
-endif
 endif
 
 if T_WIN
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -477,8 +477,7 @@ int TessBaseAPI::InitLangMod(const char* datapath, const char* language) {
     tesseract_ = new Tesseract;
   else
     ParamUtils::ResetToDefaults(tesseract_->params());
-  TessdataManager mgr;
-  return tesseract_->init_tesseract_lm(datapath, nullptr, language, &mgr);
+  return tesseract_->init_tesseract_lm(datapath, nullptr, language);
 }
 #endif  // ndef DISABLED_LEGACY_ENGINE
 
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -460,12 +460,13 @@ void Tesseract::SetupUniversalFontIds() {
 
 // init the LM component
 int Tesseract::init_tesseract_lm(const char *arg0, const char *textbase,
-                                 const char *language, TessdataManager *mgr) {
+                                 const char *language) {
+  TessdataManager mgr;
   if (!init_tesseract_lang_data(arg0, textbase, language, OEM_TESSERACT_ONLY,
-                                nullptr, 0, nullptr, nullptr, false, mgr))
+                                nullptr, 0, nullptr, nullptr, false, &mgr))
     return -1;
   getDict().SetupForLoad(Dict::GlobalDawgCache());
-  getDict().Load(lang, mgr);
+  getDict().Load(lang, &mgr);
   getDict().FinishLoad();
   return 0;
 }
--- a/src/ccmain/tesseractclass.h
+++ b/src/ccmain/tesseractclass.h
@@ -556,7 +556,7 @@ class Tesseract : public Wordrec {
   void SetupUniversalFontIds();
 
   int init_tesseract_lm(const char* arg0, const char* textbase,
-                        const char* language, TessdataManager* mgr);
+                        const char* language);
 
   void recognize_page(STRING& image_name);
   void end_tesseract();
--- a/src/ccutil/Makefile.am
+++ b/src/ccutil/Makefile.am
@@ -48,6 +48,10 @@ AM_CPPFLAGS += $(libzip_CFLAGS)
 else
 if HAVE_MINIZIP
 AM_CPPFLAGS += $(minizip_CFLAGS)
+else
+if HAVE_ZZIPLIB
+AM_CPPFLAGS += $(zziplib_CFLAGS)
+endif
 endif
 endif
 endif
--- a/src/ccutil/Makefile.am
+++ b/src/ccutil/Makefile.am
@@ -42,18 +42,15 @@ libtesseract_ccutil_la_SOURCES = \
 
 if HAVE_LIBARCHIVE
 AM_CPPFLAGS += $(libarchive_CFLAGS)
-else
+endif
 if HAVE_LIBZIP
 AM_CPPFLAGS += $(libzip_CFLAGS)
-else
+endif
 if HAVE_MINIZIP
 AM_CPPFLAGS += $(minizip_CFLAGS)
-else
+endif
 if HAVE_ZZIPLIB
 AM_CPPFLAGS += $(zziplib_CFLAGS)
-endif
-endif
-endif
 endif
 
 if T_WIN
--- a/src/ccutil/Makefile.am
+++ b/src/ccutil/Makefile.am
@@ -40,6 +40,18 @@ libtesseract_ccutil_la_SOURCES = \
     unichar.cpp unicharcompress.cpp unicharmap.cpp unicharset.cpp unicodes.cpp \
     params.cpp universalambigs.cpp
 
+if HAVE_LIBARCHIVE
+AM_CPPFLAGS += $(libarchive_CFLAGS)
+else
+if HAVE_LIBZIP
+AM_CPPFLAGS += $(libzip_CFLAGS)
+else
+if HAVE_MINIZIP
+AM_CPPFLAGS += $(minizip_CFLAGS)
+endif
+endif
+endif
+
 if T_WIN
 AM_CPPFLAGS += -DWINDLLNAME=\"lib@GENERIC_LIBRARY_NAME@\"
 endif
--- a/src/ccutil/tessdatamanager.cpp
+++ b/src/ccutil/tessdatamanager.cpp
@@ -24,6 +24,16 @@
 #include "tessdatamanager.h"
 
 #include <cstdio>
+#include <string>
+
+#if defined(HAVE_LIBARCHIVE)
+#include <archive.h>
+#include <archive_entry.h>
+#elif defined(HAVE_LIBZIP)
+#include <zip.h>
+#elif defined(HAVE_MINIZIP)
+#include <unzip.h>
+#endif // ZIP supported
 
 #include "errcode.h"
 #include "helpers.h"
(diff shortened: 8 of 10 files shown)


stweil:
Open questions:

- Do we want a new format? As I'm not the first one who had the idea, I think the answer is yes.
- Do we need support for both old (current) and new format? I'd drop support for the old format
  and remove `combine_tessdata`.
- Should the `traineddata` files in the new format add `.zip` to the file names? I'd omit `.zip`.
- Should the code for minizip be added to the Tesseract sources, or should we add an external dependency to libminizip-dev?

Shreeshrii:
Yes, there have been requests for more compact/compressed traineddata files.

Another Qn. 

* Should the new format be limited to tesseract 4.0 or also applied to 3.05?

stweil:
`libminizip-dev` was added in Ubuntu Xenial (16.04), so the current Travis build environment which is based on Ubuntu Trusty does not provide it.

Shreeshrii:
>libminizip-dev was added in Ubuntu Xenial (16.04), so the current Travis build environment which is based on Ubuntu Trusty does not provide it.

Why not use a different compression library that is available on different o/s as well as older ubuntu versions?

stweil:
On my Debian system I find these libraries: [`minizip`](https://github.com/madler/zlib/tree/master/contrib/minizip), [`libzip`](https://nih.at/libzip/), [`zzlib`](http://zziplib.sourceforge.net/zzip-file.html). As far as I know all use licenses which are compatible with Tesseract. I assume any of those can be used and expect that none of them will be available as a binary for Windows (maybe also not for macOS), but I did not have a look.

stweil:
Zip format reduces `eng.traineddata` from about 31 MiB to 16 MiB (48 % compression).

zdenop:
Please move discussion to tesseract-dev forum. This is significant change.

egorpugin:
What about lz4?

stweil:
> Please move discussion to tesseract-dev forum. This is significant change.

See [this discussion](https://groups.google.com/forum/?hl=en#!searchin/tesseract-dev/zip|sort:date/tesseract-dev/U5HSugUeeeI) in the forum. I added a link to GitHub there.

stweil:
> libarchive handles all formats

It is also supported by current Linux distributions and would be interesting if compressed tar instead of zip is preferred. I added it to my previous post.

egorpugin:
libarchive supported since very early ubuntu versions.
http://packages.ubuntu.com/search?suite=precise&searchon=names&keywords=libarchive

Personally, I'm using libarchive in cppan. The code for any formats is very simple, see:
https://github.com/egorpugin/primitives/blob/master/pack/include/primitives/pack.h
https://github.com/egorpugin/primitives/blob/master/pack/src/pack.cpp

stweil:
> What about lz4?

It does not compress very good (see result added to the list above).

egorpugin:
Actually I wanted to say lzma, which is .xz extension.
Sorry. :)

stweil:
Rebased and added support for `libzip`.

egorpugin:
What libraries are currently in use in your PR?
libarchive? minizip? libzip? 
I see libarchive in build scripts, but not in code.
Maybe it worth it to use only one implementation (library)? I don't like multiple implementations for same thing.

stweil:
This is experimental code, as there is still no decision whether compressed archives should be supported with which format and which library.

The current code uses `libzip`, if that is not found `minizip`. If neither of those is found, it uses the normal code. I prepared the code for more experiments to support compressed tar archives with `libarchive` for example.

stweil:
As you can see [here](https://github.com/tesseract-ocr/tesseract/pull/911/files#diff-4ec9988bbd350d92539e7ee9a30a2f91R48), the implementations for the two currently supported libraries are very similar.

stweil:
The latest code also supports `libarchive` (highest priority). With that library, all kinds of compressed archives should work (up to now I tested with zip only).

stweil:
As `libarchive` indeed supports all formats, I could compare the time needed for each format. Tesseract was run 5 times on each format with English on a simple hello world text. Below is the result sorted by time in seconds for each test. Interpretation:

- The original Tesseract format, uncompressed tar and lz4 tar are similar and fastest.
- zip needs about 150 ms more time than the original Tesseract format.
- 7z and xz tar need about 850 ms more time than the original Tesseract format.
- bz2 tar is slowest and needs about 1450 ms more time than the original Tesseract format.

The file i/o from disk did not play a role in this test because of the Linux file cache and the SSD of my computer.

      0.13 eng.traineddata.tar
      0.14 eng.traineddata
      0.14 eng.traineddata.tar
      0.14 eng.traineddata.tar
      0.14 eng.traineddata.tar
      0.15 eng.traineddata
      0.15 eng.traineddata.tar
      0.15 eng.traineddata.tar.lz4
      0.16 eng.traineddata
      0.16 eng.traineddata
      0.17 eng.traineddata.tar.lz4
      0.17 eng.traineddata.tar.lz4
      0.18 eng.traineddata
      0.18 eng.traineddata.tar.lz4
      0.22 eng.traineddata.tar.lz4
      0.29 eng.traineddata.zip
      0.29 eng.traineddata.zip
      0.29 eng.traineddata.zip
      0.30 eng.traineddata.zip
      0.30 eng.traineddata.zip
      0.97 eng.traineddata.tar.7z
      0.98 eng.traineddata.tar.7z
      0.98 eng.traineddata.tar.7z
      0.99 eng.traineddata.tar.7z
      0.99 eng.traineddata.tar.xz
      0.99 eng.traineddata.tar.xz
      1.00 eng.traineddata.tar.xz
      1.00 eng.traineddata.tar.xz
      1.00 eng.traineddata.tar.xz
      1.04 eng.traineddata.tar.7z
      1.55 eng.traineddata.tar.bz2
      1.56 eng.traineddata.tar.bz2
      1.61 eng.traineddata.tar.bz2
      1.62 eng.traineddata.tar.bz2
      1.66 eng.traineddata.tar.bz2

Shreeshrii:
Please also try the test with a different language. Maybe one which has the
largest traineddata size, to see if filesize has any impact to the relative
speeds
Thanks.

On May 14, 2017 1:09 PM, "Stefan Weil" <notifications@github.com> wrote:

As libarchive indeed supports all formats, I could compare the time needed
for each format. Tesseract was run 5 times on each format with English on a
simple hello world text. Below is the result sorted by time in seconds for
each test. Interpretation:

   - The original Tesseract format, uncompressed tar and lz4 tar are
   similar and fastest.
   - zip needs about 150 ms more time than the original Tesseract format.
   - 7z and xz tar need about 850 ms more time than the original Tesseract
   format.
   - bz2 tar is slowest and needs about 1450 ms more time than the original
   Tesseract format.

The file i/o from disk did not play a role in this test because of the
Linux file cache and the SSD of my computer.

  0.13 eng.traineddata.tar
  0.14 eng.traineddata
  0.14 eng.traineddata.tar
  0.14 eng.traineddata.tar
  0.14 eng.traineddata.tar
  0.15 eng.traineddata
  0.15 eng.traineddata.tar
  0.15 eng.traineddata.tar.lz4
  0.16 eng.traineddata
  0.16 eng.traineddata
  0.17 eng.traineddata.tar.lz4
  0.17 eng.traineddata.tar.lz4
  0.18 eng.traineddata
  0.18 eng.traineddata.tar.lz4
  0.22 eng.traineddata.tar.lz4
  0.29 eng.traineddata.zip
  0.29 eng.traineddata.zip
  0.29 eng.traineddata.zip
  0.30 eng.traineddata.zip
  0.30 eng.traineddata.zip
  0.97 eng.traineddata.tar.7z
  0.98 eng.traineddata.tar.7z
  0.98 eng.traineddata.tar.7z
  0.99 eng.traineddata.tar.7z
  0.99 eng.traineddata.tar.xz
  0.99 eng.traineddata.tar.xz
  1.00 eng.traineddata.tar.xz
  1.00 eng.traineddata.tar.xz
  1.00 eng.traineddata.tar.xz
  1.04 eng.traineddata.tar.7z
  1.55 eng.traineddata.tar.bz2
  1.56 eng.traineddata.tar.bz2
  1.61 eng.traineddata.tar.bz2
  1.62 eng.traineddata.tar.bz2
  1.66 eng.traineddata.tar.bz2

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/tesseract-ocr/tesseract/pull/911#issuecomment-301296366>,
or mute the thread

stweil:
Test results with `libarchive` for `mya.traineddata`. I did not test lz4, but added a test with gz format.

    0.48 mya.traineddata.tar
    0.49 mya.traineddata
    0.49 mya.traineddata.tar
    0.49 mya.traineddata.tar
    0.49 mya.traineddata.tar
    0.50 mya.traineddata
    0.52 mya.traineddata
    0.52 mya.traineddata.tar
    0.54 mya.traineddata
    0.54 mya.traineddata
    0.79 mya.traineddata.tar.gz
    0.80 mya.traineddata.tar.gz
    0.80 mya.traineddata.tar.gz
    0.82 mya.traineddata.zip
    0.84 mya.traineddata.zip
    0.85 mya.traineddata.zip
    0.86 mya.traineddata.tar.gz
    0.86 mya.traineddata.zip
    0.88 mya.traineddata.tar.gz
    0.90 mya.traineddata.zip
    2.38 mya.traineddata.7z
    2.38 mya.traineddata.7z
    2.38 mya.traineddata.tar.xz
    2.40 mya.traineddata.7z
    2.41 mya.traineddata.tar.xz
    2.45 mya.traineddata.7z
    2.45 mya.traineddata.tar.xz
    2.46 mya.traineddata.7z
    2.46 mya.traineddata.tar.xz
    2.49 mya.traineddata.tar.xz
    3.69 mya.traineddata.tar.bz2
    3.74 mya.traineddata.tar.bz2
    3.75 mya.traineddata.tar.bz2
    3.79 mya.traineddata.tar.bz2
    3.84 mya.traineddata.tar.bz2

egorpugin:
lzma compresses slower but better? Or is it also decompress slower?

stweil:
lzma created the xz files. 7zip and lzma gave the best compression ratios, but both also need some time for the decompression (which is relevant for Tesseract): they need about 1.9 s more time (but still are faster than bz2).

Please note that the current code for all formats reads all parts of the `tessdata` file, no matter whether they are used or not, so the decompression overhead could be reduced.

Shreeshrii:
@theraysmith  wrote on 4/18/14

>I have no objection to switching to zip (with no tar) for the tessdata files. That should be usable by everybody more easily.

and on 4/20/14

>I spent some time looking at zlib. It doesn't seem to make it easy to randomly access named entities in >a gzip file, unless I am missing something. The memory compress/uncompress functions are quite nice >though.
>
> For the next version it would be nice to:
> Update tessdatamanager to cope with compressed components.
> Eliminate fread/fscanf from file input code and allow everything to read from a memory buffer.
> These can probably both be achieved with the TFile class that I added for 3.03.
> 
> This is a change in direction from my previous work with new classifier experiments, where I have been writing everything to use Serialize/DeSerialize and FILE streams, but this doesn't seem to be as portable as I had hoped, due to its reliance on fmemopen. It seems it would be better to make everything use memory buffers and push the file I/O responsibility out to TessDataManager/TFile, which could then just as easily deal with compressed files or in-memory data.

@stweil Do all the methods you tested support ` randomly accessing named entities`?

@theraysmith Is there a particular reason for ` zip (with no tar)`?

stweil:
The current Tesseract code reads the whole `tessdata` file into memory and gets all data from memory. My implementation for compressed archive files does that, too. Therefore random access is trivial: all component files are in a vector of byte arrays.

stweil:
Why zip not tar? Maybe because zip is supported on Windows by default, while tar and other formats need additional installations? I use 7-Zip on Windows to compress and decompress files. See also https://en.wikipedia.org/wiki/Comparison_of_file_archivers.

stweil:
> I have no objection to switching to zip (with no tar) for the tessdata files. That should be usable by everybody more easily.

On Windows, that would require a rename with `.zip`extension, or the installer must register `.traineddata` as a zipped data type. Otherwise the Windows explorer won't decompress such files easily. Personally I prefer to keep the `.traineddata` extension even for zip files (like Java does it with its `.jar` extension).

stweil:
Now I added implementation and timing results for zziplib. It looks like this is the fasted code for reading zipped `traineddata` files.

Shreeshrii:
>I'd drop support for the old format
and remove combine_tessdata.

combine_tessdata is used for extracting components, replacing some, etc. 

Probably first step can be a zippeddata file which after unzipping provides the traineddata.

Unless, there are zip/unzip equivalent features for combine_tessdata functionality.

Shreeshrii:
Would same extension cause confusion as to which version of traineddata file was being used?

stweil:
> combine_tessdata is used for extracting components, replacing some, etc.
Probably first step can be a zippeddata file which after unzipping provides the traineddata.

Zipping a tessdata file means extracting the components from that file and writing a zip archive file which includes those components. So combine_tessdata can be replaced by your favorite zip / unzip program.

zdenop:
I would suggest to use different file extensions (e.g. traineddata4). This would help us to have installation of 3.05 and 4.00 at the same time and to avoid mistakes like using combine_tessdata on compressed data.

I would suggest to use zip compression as default because it seems to be supported on all platforms easily... Maybe support for multiple compression formats would be interesting (e.g. if somebody will be focused on size and not on speed), but this is not must at the moment.

theraysmith:
On Windows, that would require a rename with .zipextension, or the
> installer must register .traineddata as a zipped data type. Otherwise the
> Windows explorer won't decompress such files easily. Personally I prefer to
> keep the .traineddata extension even for zip files (like Java does it
> with its .jar extension).
>
>
> This comment has me confused. Are we talking about
1. compressing the files just for installation. after installation files
are uncompressed
2. compressing the components of a file using zlib and changing
tessdatamanager to allow that. FIles are otherwise uncompressed.
3. Leaving compressed files on disk after installation and relying on some
auto decompression feature of the OS to decompress the files on reading
?

I thought we were talking about 2, but registering something sounds like 3.
If we allow a backwards-breaking change, we could change the format of
traineddata, add a version number, and allow compression of components. It
would be fairly easy.


-- 
Ray.

stweil:
We are talking about 2: files in zip archive format which include compressed components.

Such files can be written and read by all common operating systems. Linux file explorers are clever and know how to handle such files even if they are named `eng.traineddata`. The Windows explorer is not clever. It expects a know file extension. So if users want to see what is inside of `eng.traineddata`, I must set a Windows registry key which tells the file explorer that the `traineddata` extension is handled like `zip`.

stweil:
Currently `combine_tessdata` crashes when it is tried with a compressed archive. It can be made more robust to simply fail with such data, or it could report if it gets a zipped archive (if we restrict to zip support). Such code can also be back ported to 3.05.

Platform support is good on Linux (minizip in latest distributions, libzip, libzzip and libarchive also in older ones), but macOS with brew should also support all four libraries (untested), and so does Cygwin with Mingw-w64 for Windows.

stweil:
> This would help us to have installation of 3.05 and 4.00 at the same time and to avoid mistakes like using combine_tessdata on compressed data.

Currently `combine_tessdata` crashes when it is tried with a compressed archive. It can be made more robust to simply fail with such data, or it could report if it gets a zipped archive (if we restrict to zip support). Such code can also be back ported to 3.05. I don't think that using the same tessdata` directory for 3.05 and for 4.x is a good idea. Therefore I don't think that 4.x needs a different file extension for `tessdata` in zip format.

Platform support is good on Linux (minizip in latest distributions, libzip, libzzip and libarchive also in older ones), but macOS with brew should also support all four libraries (untested), and so does Cygwin with Mingw-w64 for Windows. I don't know the situation for Visual Studio, but expect that small libraries like `libzzip` for example are much easier to build than larger libraries with more functionality like `libarchive`.

egorpugin:
>  I don't know the situation for Visual Studio, but expect that small libraries like libzzip for example are much easier to build than larger libraries with more functionality like libarchive.

We are using cppan build for Visual Studio, libarchive is already there. It can be included adding 1-3 lines. Other small libs you listed are not a problem too if needed.

stweil:
Then I suggest to use `libzzip` (sometimes also called `zziplib`) if zip support is sufficient because it looks like the fastest one, and it is also small.

`libarchive` is a good choice (the only one I know) if any common compressed archive format should be supported.

stweil:
> Would same extension cause confusion as to which version of traineddata file was being used?

Software can detect the format automatically. That can be added to 3.05, too, so that version would simply refuse to read files in the wrong format.

With a standard archive format (like zip), it would be easy to write additional information into the archive files, for example a README or a VERSION file. Tesseract would ignore any component which it does not know.

amitdo:
zziplib seems to be inactive for years.

https://sourceforge.net/projects/zziplib/?source=navbar
https://sourceforge.net/p/zziplib/svn/HEAD/tree/
https://sourceforge.net/projects/zziplib/files/zziplib13/
https://sourceforge.net/p/zziplib/patches/?source=navbar

Choosing more than one library and one format for compression will make maintenance more complicated.

theraysmith:
The most important features for me are:
TessdataManager::LoadMemBuffer must still work, ie the compression library must still be able to work on a pre-loaded file
Speed. There is little point compressing if the decompression speed is poor.
Command-line and API must still work. If combine_tessdata goes away, then there must be an easy-to-use replacement, and Tessdatamanager::SaveFile must still work.
Minimal extra dependencies - that was the main motivation for using zlib, as it is already an indirect dependency.
Summary: implement inside TessdataManager, and keep its existing interface in tact.

stweil:
> zziplib seems to be inactive for years.

Then either it is a small and stable code, or it is unsupported for some reason.
Which library would you suggest?

amitdo:
Check what other well known libraries use.
For example, the Debian cmake package depends on zlib and libarchive.

egorpugin:
libarchive choice is obvious.

stweil:
> zziplib seems to be inactive for years.

I found it maintained on [GitHub](https://github.com/gdraheim/zziplib).

jbarlow83:
Speed is more important than disk space. That means LZ4 is probably the best candidate, as the English test indicated. It was designed for this sort of situation and is the tool of choice for e.g. Linux kernel and initramfs decompression on low memory devices.
https://lz4.github.io/lz4/#interoperable-lz4

stweil:
> Speed is more important than disk space.

We are talking about typically less than 300 ms needed for large zip archives, so a Tesseract run with several languages would maybe run 1 s longer with zip archives. Is that too much when typically Tesseract runs for much more than 10 s?

Tesseract could be optimized by processing many image files in a single run. Then loading of `tessdata` files is needed only once.

zdenop:
> With a standard archive format (like zip), it would be easy to write additional information into the archive files, for example a README or a VERSION file. Tesseract would ignore any such component which it does not know.

There is already possibility to include `config` file to traineddata. I would avoid suggest to add other file and made the `config `file as required with needed information (e.g. version number of data file, minimum (maximum?) version of tesseract, author, licence etc...)

stweil:
If a standard format is used for `traineddata`, we no longer can derive the data endianness from the file format. So either endianness information must be added somewhere in the archive (in the `config` component?), or all code must be changed to read and write a fixed endianness. I have a personal opinion on this issue, but in the end that's something @theraysmith must decide.

jbarlow83:
@stweil My typical Tesseract workload is tens of thousands of images in a cloud. Combining disparate images into a multipage TIFF is a pain. My files come as PDFs anyway; it's a good format to normalize files to. Then I use my program ocrmypdf to split them into one page images and recombine. In short, I map-reduce Tesseract processes over single page images, with all the usual scaling and robustness benefits. A long running process that does a lot of work is expensive to have crash.

Since AWS compute time is per unit time, adding 1 second per Tesseract process would be something like a 1-2% increase in my bills, while decreasing storage would have no measurable savings because the incremental gain is small compared to the size of the operating system. (And most charge for a pool of storage rather than storage actually used.)

I do support the change, I just think LZ4 hits the sweet spot for this sort of problem.

zdenop:
Just for the note: Google released during last years these compression algorithms:

- [zopfli](https://github.com/google/zopfli) focused on size
- [brotli](https://github.com/google/brotli) focused on [speed](https://opensource.googleblog.com/2015/09/introducing-brotli-new-compression.html)

I found out [result of test several compression algorithm](https://quixdb.github.io/squash-benchmark/#results-table), if somebody would like to make more test...

stweil:
@jbarlow83, I did not mean using multipage TIFF images, but enhancing the Tesseract command line API to allow passing more than one image file. That should reduce the compute time for your use case already with the current `traineddata` file format.

amitdo:
https://github.com/powturbo/TurboBench

amitdo:
>Tesseract could be optimized by processing many image files in a single run. Then loading of tessdata files is needed only once.

Actually, this feature is supported with 4.0/3.x.

stweil:
> Actually, this feature is supported with 4.0/3.x.

I did not know that. How would I start Tesseract to process page1.png and page2.png in a single run?

amitdo:
Prepare a text file that has the path to each image:

```
path/to/1.png
path/to/2.png
path/to/3.tiff
```

Save it, and then give its name as input file to Tesseract.

`tesseract savedlist output`

stweil:
Thank you, good to know that. It looks like the documentation and the program help text need an update.

Currently all pages are written to one output file (per format). Some formats include page information (hOCR, PDF). Others like TXT don't, but could use a page separator character (ASCII 0x0C = FF). Would it help to support an output base parameter with placeholders like page number or image base name to generate one output file per input image?

amitdo:
Maybe we should discuss it in another issue.

I don't like mailing lists. Also Ray replies more in Github issues/PRs. 
What about opening a new repo on Github for dev discussions via issues?

stweil:
The multi-page feature was added by commit 25a8c7b72006a11ba0b31dfd5210f7b9401eb27d.

Shreeshrii:
@amitdo I have added a separate issue re multipage feature - https://github.com/tesseract-ocr/tesseract/issues/928

Rather than a new repo, we can use RFC: for issues in this repo itself, unless @theraysmith or @zdenop have an objection.

atoron:
Guys, can you add checkboxes to the eng.traineddata? 
6-1-2017 3-51-27 pm

stweil:
I'd like to summarize the current status:

- A new format would be good, and there is a preference for the `zip` format.
- The new format would replace the old one for Tesseract 4 (no fallback necessary).
- We can keep the current names (*.traineddata). 3.05 won't load new format, 4.x won't load old format.
- `libarchive` seems to be the best choice to implement the `zip` format as it is widely available and supported.
- The `zip` format no longer includes an implicit indicator for the endianness of the trained data. A fixed endianness (little endian) for stored data does not look like a possible option for now, so an explicit endianness indicator must be added to the data in the config component.

Is there anything I missed or are there other opinions? If not, I'd implement the code and create a pull request.

Shreeshrii:
@stweil Please also make a PR for tessdata to convert all current 4.0 traineddata files in new zipped format.

@zdenop I would also suggest that a new tag/release be made 4.0 (alpha1/2 or beta) for tesseract, tessdata as well as langdata with this new PR.

Thanks!

Shreeshrii:
Please also add version info as part of traineddata. 
I think @zdenop had suggested adding it to the config files.

stweil:
Rebased PR and fixed merge conflicts.

amitdo:
@theraysmith commented:

 >Are we talking about
>...
>2. compressing the components of a file using zlib and changing
tessdatamanager to allow that. FIles are otherwise uncompressed.
>...
>?
>
>I thought we were talking about 2...

@stweil commented:

>We are talking about 2: files in zip archive format which include compressed components.

It seems to me that Stefan and Ray are thinking about two different ways to implement compression:

Stefan's way is to create one zip archive from the files in traineddata. 
combine_tessdata won't be needed anymore. The user will use an external program to compress and decompress the traineddata (zip) file. 
Stefan patch adds the ability read (uncompress) the zip archive by TessdataManager.

Ray's way is to use the zlib library directly to compress individual files inside traineddata without creating a zip archive.
combine_tessdata will still be needed by the user.

stweil:
Thank you, Amit, for the clarification. Yes, I'd replace the proprietary `tessdata` format by the standard `zip` format. Users could still use `combine_tessdata` (it would support both old and new format), but they could also use any other program (for example the Windows file explorer) to create or expand the new tessdata files in zip format.

As there was a preference for using `libarchive`, there would even be more options because that library supports a whole range of compression and archive formats. Some of those formats compress the whole archive (typically a `tar` format) instead of compressing the single components.

amitdo:
Ray seems to want just zlib without anything else.

We need an 'OK' from him to go in your way which depends on another library.

stweil:
PR rebased to fix merge conflicts.

christophered:
@stweil will the zip compression be implemented?

stweil:
@christophered, it is mostly implemented, and I think it can be integrated as soon as there is consensus what zip compression means (see discussion above) and that we want to have it.

stweil:
PR rebased to fix merge conflicts.

stweil:
Please remove that.

christophered:
@stweil @amitdo @egorpugin have you test zstd compression? I have tested it, and its very fast. Also if you add a dictionary to it, the compression ratio would be even better, I think it's a game changer.
https://github.com/facebook/zstd

ZSTD compressing by a dictionary:
- Create the dictionary
`zstd --train FullPathToTrainingSet/* -o dictionaryName`

- Compress with dictionary
`zstd -D dictionaryName FILE`

- Decompress with dictionary
`zstd -D dictionaryName --decompress FILE.zst`

- Increase dicitonary size
`zstd --train dirSamples/* -o dictionaryName --maxdict=1024KB`

stweil:
I have not tested it yet, but it looks like we get Zstandard support with libarchive. Pull request https://github.com/libarchive/libarchive/pull/905 added Zstandard there.

zdenop:
AFAIR there was intention to use already used libraries e.g. not to increase dependencies.
To build tesseract with VS without cppan on Windows is already pain...

egorpugin:
With next libarchive release I'll add zstd dependency into it, so tesseract will get it automatically.

stweil:
Tesseract only needs to add a dependency on `libarchive` to get support for compressed archives.

zdenop:
So If I understand it right if we compress datafiles with Zstandard users will need on all platform to compile libarchive + Zstandard...

stweil:
That's correct. Therefore I still would distribute the datafiles with zip format which hopefully has good support on all platforms. But users who need maximum performance then could repack their needed datafiles with a different compression standard.

Shreeshrii:
A fast compressor/decompressor 

https://github.com/google/snappy

zdenop:
Milestone is set to 4.1.0. Is it time to merge it? There was not a lof of changes here in last months...

stweil:
Pull request #2290 now includes the implementation with `libarchive`, so this proof of concept is now obsolete and can be closed.

Issue #529, opened 2016-12-04
~ObjectCache(): WARNING! LEAK! object 


Shreeshrii:
While trying to process a gif file, when leptonica was not built with giflib, get the following messages

```
/mnt/c/Users/User/shree$ tesseract san001.gif san001-gif --psm 6 --oem 4 -l san
Tesseract Open Source OCR Engine v4.00.00alpha with Leptonica
Error in pixReadStreamGif: function not present
Error in pixReadStream: gif: no pix returned
Error in pixRead: pix not read
Error during processing.
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x1fd72f0 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddatalstm-punc-dawg)
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x1fd8290 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddatalstm-word-dawg)
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x1fd7110 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddatalstm-number-dawg)
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x426afd0 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddatapunc-dawg)
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x1fd6f20 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddataword-dawg)
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x426ae60 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddatanumber-dawg)
ObjectCache(0x7f6a7eb9e5c0)::~ObjectCache(): WARNING! LEAK! object 0x472cef0 still has count 1 (id /mnt/c/Users/User/tesseract-ocr/tessdata/san.traineddatabigram-dawg)
[...]

PR #912, opened 2017-05-12
main: Fix two memory leaks and fix order of destructor calls

stweil:
When Tesseract terminates by calling the exit function,
the destructor of any local auto variable is not called.

Fix two cases by using static variables.

### DIFF
--- a/api/tesseractmain.cpp
+++ b/api/tesseractmain.cpp
@@ -424,7 +424,9 @@ int main(int argc, char** argv) {
   }
 
   PERF_COUNT_START("Tesseract:main")
-  tesseract::TessBaseAPI api;
+
+  // Avoid memory leak caused by auto variable when exit() is called.
+  static tesseract::TessBaseAPI api;
 
   api.SetOutputName(outputbase);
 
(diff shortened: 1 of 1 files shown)


amitdo:
Maybe we should use return instead of exit()

egorpugin:
Do we care about such leaks? 
api and renderer hold something important?

stweil:
Tesseract code includes more than 70 calls of `exit()`. Using `return` is a lot of work, even in `main()` (I already tried it). And static variables also reduce the stack frame size. I know someone who will be happy. :-)

egorpugin:
Probably exit()s can be replaced with exceptions?
And print errors in main.

stweil:
Yes, that sounds like the correct solution for C++ code.

amitdo:
but not in Google project...

egorpugin:
Ah, yes, remember their policy. :(

stweil:
The 2nd commit fixes issue #529. Now Valgrind is happy, too:

    LEAK SUMMARY:
       definitely lost: 0 bytes in 0 blocks
       indirectly lost: 0 bytes in 0 blocks
         possibly lost: 0 bytes in 0 blocks
       still reachable: 8 bytes in 1 blocks
            suppressed: 0 bytes in 0 blocks
    
    ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
    ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)

rfschtkt:
You must be kidding...

PR #913, opened 2017-05-12
test brew HEAD installs

stephengroat:
but allow failures

currently failing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,16 @@ addons:
     packages:
     #- g++-6
 
+matrix:
+  include:
+    - os: osx
+      script: brew install tesseract --HEAD
+      cache:
+        directories:
+          - $HOME/Library/Caches/Homebrew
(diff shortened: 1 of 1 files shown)


egorpugin:
What's this? We get very long build times with it.
See https://travis-ci.org/tesseract-ocr/tesseract/builds/232993174
@zdenop @stweil @amitdo

stephengroat:
all this does it show you that the patches that are used in `brew` to build tesseract from source aren't up to date.

the long build time looks like an anomaly, try killing the travis process and restart it

stweil:
That new test now [passes without errors](https://travis-ci.org/tesseract-ocr/tesseract/jobs/233010095) (the brew formula for tesseract was fixed). Nevertheless building Tesseract with brew takes some time, so this test adds about half an hour to the Travis test time for each pull request or update of a branch. I‌ wonder whether that's reasonable.

PR #914, opened 2017-05-13
Clean code

stweil:


amitdo:
if you rewrite some code with `NULL`, I suggest to change it to `nullptr`.

stweil:
True. Fixed with updated PR.

PR #915, opened 2017-05-13
Remove unused methods from Tessdatamanager

stweil:


stweil:
Maybe those methods are used by Google outside of Tesseract, so a review is needed.

PR #917, opened 2017-05-13
Fix crash if output file could not be opened

stweil:
This error case results in fout_ == nullptr.
Closing a nullptr file is not allowed.


stweil:
The problem was reported as part of issue #529.

Shreeshrii:
Thanks! No crash now.

```
tesseract p002-crop.bmp missing/bmp --oem 1 --psm 6 -l hin
Tesseract Open Source OCR Engine v4.00.00alpha-520-gffb1ec3 with Leptonica
Error during processing.
```

A more descriptive error message will be helpful.

PR #920, opened 2017-05-14
Improve robustness of TessdataManager

stweil:
Tesseract crashes with an unhandled exception (std::bad_alloc) if it gets
a bad tessdata file where the numEntries data field is very large (also
after swapping), for example 0x77777777.


ccutil/tessdatamanager.cpp
```diff
@@ -54,6 +54,7 @@ bool TessdataManager::LoadMemBuffer(const char *name, const char *data,
   swap_ = num_entries > kMaxNumTessdataEntries || num_entries < 0;
   fp.set_swap(swap_);
   if (swap_) ReverseN(&num_entries, sizeof(num_entries));
+  if (num_entries > kMaxNumTessdataEntries || num_entries < 0) return false;
```
amitdo: What about that part
https://github.com/tesseract-ocr/tesseract/blob/29f3de9be1fb/api/baseapi.cpp#L324
?

PR #921, opened 2017-05-15
RAII

rfschtkt:
Use the [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) idiom, e.g., to help prevent memory leaks.

RFC about ccutil/raiileptonica.h?

(2017-05-18 Heavily edited down.)

### DIFF
--- a/api/baseapi.cpp
+++ b/api/baseapi.cpp
@@ -48,6 +48,7 @@
 #include <memory> // std::unique_ptr
 
 #include "allheaders.h"
+#include "raiileptonica.h"
 
 #include "baseapi.h"
 #include "blobclass.h"
@@ -547,7 +548,7 @@ void TessBaseAPI::SetImage(const unsigned char* imagedata,
   if (InternalSetImage()) {
     thresholder_->SetImage(imagedata, width, height,
                            bytes_per_pixel, bytes_per_line);
-    SetInputImage(thresholder_->GetPixRect());
+    SetInputImage(thresholder_->GetPixRect()); // reference transfer
   }
 }
 
@@ -569,7 +570,7 @@ void TessBaseAPI::SetSourceResolution(int ppi) {
 void TessBaseAPI::SetImage(Pix* pix) {
   if (InternalSetImage()) {
     thresholder_->SetImage(pix);
-    SetInputImage(thresholder_->GetPixRect());
+    SetInputImage(thresholder_->GetPixRect()); // reference transfer
   }
 }
 
@@ -665,7 +666,7 @@ Boxa* TessBaseAPI::GetComponentImages(PageIteratorLevel level,
                                       const int raw_padding,
                                       Pixa** pixa, int** blockids,
                                       int** paraids) {
-  /*non-const*/ std::unique_ptr</*non-const*/ PageIterator> page_it(GetIterator());
+  /*used with reset()*/ std::unique_ptr</*non-const*/ PageIterator> page_it(GetIterator());
   if (page_it == NULL)
     page_it.reset(AnalyseLayout());
   if (page_it == NULL)
@@ -694,7 +695,7 @@ Boxa* TessBaseAPI::GetComponentImages(PageIteratorLevel level,
       ++component_count;
   } while (page_it->Next(level));
 
-  Boxa* boxa = boxaCreate(component_count);
+  const BoxaPtr boxa(boxaCreate(component_count));
   if (pixa != NULL)
     *pixa = pixaCreate(component_count);
   if (blockids != NULL)
@@ -709,18 +710,15 @@ Boxa* TessBaseAPI::GetComponentImages(PageIteratorLevel level,
   do {
     if (get_bbox->Run() &&
         (!text_only || PTIsTextType(page_it->BlockType()))) {
-      Box* lbox = boxCreate(left, top, right - left, bottom - top);
-      boxaAddBox(boxa, lbox, L_INSERT);
+      const BoxPtr lbox(boxCreate(left, top, right - left, bottom - top));
+      boxaAddBox(boxa.p(), lbox.p(), L_CLONE);
       if (pixa != NULL) {
-        Pix* pix = NULL;
-        if (raw_image) {
-          pix = page_it->GetImage(level, raw_padding, GetInputImage(), &left,
-                                  &top);
-        } else {
-          pix = page_it->GetBinaryImage(level);
-        }
-        pixaAddPix(*pixa, pix, L_INSERT);
-        pixaAddBox(*pixa, lbox, L_CLONE);
+        const PixPtr pix(
+          raw_image
+          ? page_it->GetImage(level, raw_padding, GetInputImage(), &left, &top)
+          : page_it->GetBinaryImage(level) );
+        pixaAddPix(*pixa, pix.p(), L_CLONE);
+        pixaAddBox(*pixa, lbox.p(), L_CLONE);
       }
       if (paraids != NULL) {
         (*paraids)[component_index] = paraid;
@@ -738,7 +736,7 @@ Boxa* TessBaseAPI::GetComponentImages(PageIteratorLevel level,
     }
   } while (page_it->Next(level));
   delete get_bbox;
-  return boxa;
+  return boxa.detach();
 }
 
 int TessBaseAPI::GetThresholdedImageScaleFactor() const {
@@ -753,7 +751,7 @@ void TessBaseAPI::DumpPGM(const char* filename) {
   if (tesseract_ == NULL)
     return;
   FILE *fp = fopen(filename, "wb");
-  Pix* pix = tesseract_->pix_binary();
+  Pix* pix = tesseract_->pix_binary(); // borrowed pointer
   int width = pixGetWidth(pix);
   int height = pixGetHeight(pix);
   l_uint32* data = pixGetData(pix);
@@ -917,6 +915,7 @@ int TessBaseAPI::RecognizeForChopTest(ETEXT_DESC* monitor) {
 // Takes ownership of the input pix.
 void TessBaseAPI::SetInputImage(Pix* pix) { tesseract_->set_pix_original(pix); }
 
+// Borrowed pointer.
 Pix* TessBaseAPI::GetInputImage() { return tesseract_->pix_original(); }
 
 const char * TessBaseAPI::GetInputName() {
@@ -974,15 +973,14 @@ bool TessBaseAPI::ProcessPagesFileList(FILE *flist,
       snprintf(pagename, sizeof(pagename), "%s", lines[page].c_str());
     }
     chomp_string(pagename);
-    Pix *pix = pixRead(pagename);
+    const PixPtr pix(pixRead(pagename));
     if (pix == NULL) {
       tprintf("Image file %s cannot be read!\n", pagename);
       return false;
     }
     tprintf("Page %d : %s\n", page, pagename);
-    bool r = ProcessPage(pix, page, pagename, retry_config,
+    bool r = ProcessPage(pix.p(), page, pagename, retry_config,
                          timeout_millisec, renderer);
-    pixDestroy(&pix);
     if (!r) return false;
     if (tessedit_page_number >= 0) break;
     ++page;
@@ -1003,22 +1001,22 @@ bool TessBaseAPI::ProcessPagesMultipageTiff(const l_uint8 *data,
                                             TessResultRenderer* renderer,
                                             int tessedit_page_number) {
 #ifndef ANDROID_BUILD
-  Pix *pix = NULL;
   int page = (tessedit_page_number >= 0) ? tessedit_page_number : 0;
   size_t offset = 0;
   for (; ; ++page) {
     if (tessedit_page_number >= 0)
       page = tessedit_page_number;
-    pix = (data) ? pixReadMemFromMultipageTiff(data, size, &offset)
-                 : pixReadFromMultipageTiff(filename, &offset);
+    const PixPtr pix(
+        data
+        ? pixReadMemFromMultipageTiff(data, size, &offset)
+        : pixReadFromMultipageTiff(filename, &offset) );
     if (pix == NULL) break;
     tprintf("Page %d\n", page + 1);
     char page_str[kMaxIntSize];
     snprintf(page_str, kMaxIntSize - 1, "%d", page);
     SetVariable("applybox_page", page_str);
-    bool r = ProcessPage(pix, page, filename, retry_config,
-                           timeout_millisec, renderer);
-    pixDestroy(&pix);
+    bool r = ProcessPage(pix.p(), page, filename, retry_config,
+                         timeout_millisec, renderer);
     if (!r) return false;
     if (tessedit_page_number >= 0) break;
     if (!offset) break;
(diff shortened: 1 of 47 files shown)


rfschtkt:
Important to note is that so far, even in this one file, I have only applied `unique_ptr_Pix` where `pixDestroy()` was already called. But, unless I am mistaken, there seem to be several places where it, or at least `pixDestroy()`, should also be used to prevent a leak, and these form the urgent rationale for yet another application of the RAII idiom.

rfschtkt:
Code freeze.

DanBloomberg:
Interesting proposal.

Here are a few things to consider with respect to the use of RAII with leptonica (I'm sure there are others, but I haven't thought much about this):

* does tesseract already use safe or shared ptrs, and if so, how does this proposed implementation fit with that; or if tesseract does not, would RAII be extended to all allocated data?

* leptonica is a C library with relatively little use of const (pretty much just for c strings).  An application that uses a lot of const leptonica variables will be very messy.

* leptonica uses ref counting clones to simplify resource ownership and make it easier for the developer to avoid memory leaks and double frees.  Clones are not thread safe; there's no locking in leptonica.  Would RAII change that?

* do the methods proposed here for exception handling clean up memory, or would they leak and therefore you have to crash the program?  (Does tesseract have exit calls on error?)

amitdo:
>does tesseract already use safe or shared ptrs, and if so

It uses C++11's `unique_ptr` in the new lstm code.
C++11's `shared_ptr` is not used.

Old code does not use RAII.

>how does this proposed implementation fit with that; 

It uses `unique_ptr` with custom deleleter that automatically calls pixDestroy() when the scope ends.

>would RAII be extended to all allocated data?

That is what we wish to do.

rfschtkt:
Regarding Dan's points, in the same order, and with `Pix` standing in for any leptonica type:
* See Amit's answer. I've only applied the types in local scope, so none of the function/procedure signatures have changed, let alone the public API. Interestingly, you could also apply RAII to `FILE`, although that raises the issue of error handling when writing. BTW, currently Tesseract very rarely evaluates the value returned by `fclose()`, even when writing!
* The `const` only applies to the `PixRef`, not to the referent. So it's like `Pix* const pix`, not `const Pix* pix`.
* These types merely call `pixDestroy()` at the appropriate time, nothing more. None of C++'s smart pointers make their _referents_ thread-safe. You can't even reliably change a `unique_ptr` or a `shared_ptr` _smart-pointer instance_ from another thread without locking said instance. The only thing you know is that `shared_ptr`'s underlying reference count handling is thread-safe. BTW, a `PixRef` would manage just a single reference count, because `boost::intrusive_ptr`, which would also call `pixClone()` at the appropriate time (not possible for, e.g., `Pixa`) would be... too intrusive (Tesseract does not yet use that library), and neither it nor `std::shared_ptr` would be very useful for local-only usage.
* So far it only lightens the mental burden of guarding the exits of the functions/procedures (unfortunately I only found 1 leak for leptonica types), and it makes the code a bit shorter (at the expense of having to use `get()` a lot). This is low-hanging fruit so far, exception safety would require a lot more concerted effort. There was an issue #529 recently with `exit()` and a static variable that was destroyed while a non-static variable was still holding references to it, causing the static variable's destructor to complain... so even `exit()` has problems, in C++ at least. :-)

DanBloomberg:
@Raf

Thank you very much for your clear and thorough explanation of my questions!

I was also glad you found that "unfortunately" there are essentially no memory corruption/leak issues from leptonica usage in tesseract, even though nearly everything happens on the heap  ;-)

rfschtkt:
Ah, but I only looked at local usage, and I did actually "wonder" whether using a `PixRef** pix` parameter without looking at the current value of `*pix`, as happens often in Tesseract, was always a responsible thing to do: I would require it to always be NULL as a precondition (with a check or at least a diagnostic), or always `pixDestroy()` a non-NULL value as part of the contract.

So the idea of my (off-line) query was to make you aware of a proposed set of C++ types to help manage those references, perhaps even to standardise them as part of leptonica++, so to say (an add-on, not a replacement). Any thoughts there, about the names perhaps (I thought PixRef would better fit the limited responsibility than PixPtr)?

rfschtkt:
It would also be easy enough to _derive_ `PixPtr` from `std::unique_ptr<Pix>`, and give it an extra function to allow `pix.p()` with the same meaning as `pix.get()`, saving two characters each time and not letting the raw-pointer access function tail visually wag the variable dog, which together might otherwise get annoying to the point that people reject the idea of using a smart pointer, and I actually think it was a bad decision to make it that _difficult_ to access the raw-pointer value.

I did a quick experiment (thanks to automated refactoring) and pushed the result. Somehow, some uses were not yet converted, I see, but I like the result so far...

rfschtkt:
Prefer const + clone() to non-const + release().

rfschtkt:
The rules may be "simple", but this is C++, so we should expect to be able to program at a higher level of abstraction than C allows, and leave matters of ownership and guarding all the exits (mostly) behind us, even if there are still too many problems left to achieve exception safety.

I do remember the example you're talking about, in baseapi.cpp, which looked particularly weird to me because it would first sort of _abandon_ `lbox` (`L_INSERT`)... and then still `L_CLONE` it further down, so that was sort of a surprise. This is probably straightforward enough if you are familiar with the function `boxaAddBox()` where the `L_INSERT` occurred (assumedly the instance gets added to a container as-is, and not, e.g., serialised and destroyed), but it "feels" wrong because it forces you to still think at that level. In my replacement, the `BoxRef` keeps the object alive throughout its scope (until the eventual `~BoxRef()`), whatever else happens, and in the meantime it can be `L_CLONE`d an arbitrary number of times. Strictly speaking, there are 2 more reference-count manipulations (one for `L_CLONE` instead of `L_INSERT`, then one for `boxDestroy()` inside `~BoxRef()`), but I think you have to strip the program of all functionality before that becomes noticeable. The story with `pix` is even simpler.

A `PixRef` just manages one reference count, whether it's from copying or cloning: the constructor takes over one reference, and the destructor destroys it. See for example `const PixRef word_in_xheight(pixCopy(NULL, pix));` in devanagari_processing.cpp... well, the other occurrences are with `reset()`.

amitdo:
>(Edited) "feels"->feels, are [->all] with

@rfschtkt, you don't have to add notes after fixing typos / grammar mistakes.
Just fix it :-)

DanBloomberg:
Don't get me wrong.  Lots of C++ users of leptonica prefer to use them with scoped ptrs, and I think that's fine.  Chaqu'un à son goût.

I did wonder about the CLONE vs COPY case for adding a pix to a pixa, for example.  If you use the CLONE and later change the pix, the version in the pixa is also changed.  So sometimes you'd prefer to use a COPY, which gives you two instances of the image.  The copy cannot be wrapped with a scoped ptr, because it's made internally in pixaAddPix(), but it probably doesn't matter because that pix will be destroyed when pixaDestroy() is called.

Does this also work when you INSERT a PixRef->p() into a pixa?  (i.e., how does the scoped ptr handle know that it has transferred ownership to the pixa?  Do you have to do an explicit reset?)

rfschtkt:
_Not_ paragraph by paragraph:

It's just a reference, i.e., a notch in the `refcount` (that's why I renamed it `PixRef`). Use `pixref.p()` like `pixpointer`, `pixClone(pixref.p())` or (equivalently) `pixref.clone()` like `pixClone(pixpointer)`, and `pixref.release()` like `pixpointer` later followed by `pixpointer = NULL;`.

You could equivalently do `pixaAddPix(pixa.p(), pix.p(), L_CLONE);` or `pixaAddPix(pixa.p(), pix.clone(), L_INSERT);`, or, if you aren't using `pix` afterwards, `pixaAddPix(pixa.p(), pix.release(), L_INSERT);`, but I would avoid the latter (no `const PixRef`) and prefer the first over the second (shorter, doesn't needlessly use extra feature `clone()`).

The pointer doesn't "know" what happened with the pointer you got from `p()`, so if you are explicitly transferring ownership you must use `release()` instead (again, I would prefer a `const PixRef` and cloning, just because we can, because it's reference-counted, no unique ownership implied); just `reset()` after `p()` would be incorrect because the `PixRef` would call `pixDestroy()` on the old value.

rfschtkt:
OK, let's get this done, because _RAII is awesome!_
* 1st Leptonica: `PixPtr` (you can't use -> with a "reference" in C++, dummy!) / `PixRef` (as in "one reference")?
* 2nd Leptonica: yes (far less visual dog wagging) / no (straight-up `unique_ptr`)?
* 3rd Leptonica (only if 2nd, I suppose): yes (more `const`) / no (straight-up `unique_ptr`)? if the former: `detach()` (like boost) / `clone()` (Leptonica-specific)?
* 4th Leptonica (only if 3rd, and only if `detach()`, I suppose): yes (more `const`) / no (lame!)?

(My current preferences are weakly or strongly on the left, so not exactly the current code.)

Once the design is fixed, I should probably collapse 1+2 and 3+4, or 1+2+3+4, and then maybe split them up again by directory, or something else.

(And to think I was actually just looking at the code to find out how to maybe get Tesseract to avoid recognising dotted lines as extra characters...)

DanBloomberg:
OK, I'm confused, but you should really be having this conversation with tesseract maintainers and developers, and in particular with Ray Smith.  Leptonica isn't going to change (e.g., into a C++ library).

To clone a boxa, use boxaCopy(..., L_CLONE)

As for the const issue, as mentioned before, leptonica doesn't use const as part of the function contract except for char* names (e.g., file paths).  Therefore, using const leptonica data in tesseract will be very ugly with all the const-removing reinterpret_cast<...> stuff

rfschtkt:
"OK, I'm confused"
Then I have failed miserably... :-(

"Leptonica isn't going to change (e.g., into a C++ library)."
Smart pointers can be used with unchanged C code, no active cooperation required.

"To clone a boxa, use boxaCopy(..., L_CLONE)"
Ah so... but it still seems easier to just directly manipulate `refcount` for all types, as an orthogonal implementation of `detach()`.

I think I already mentioned that the const is about the pointer (a reference is equivalent to a const pointer), not the referent (it's not a pointer to const). BTW, in C++, if you want to change a type that's declared `const`, use `const_cast`, not `reinterpret_cast`, which is not even allowed to remove `const`.

DanBloomberg:
I agree with everything, except perhaps blame for my confusion ...  :-)

Don't know how much benefit you get from enforcing constancy of the pointer, which typically only changes if it's "in-place" with a free and a heap allocation.  These are very infrequent; e.g., pixTransferAllData() and pixSwapAndDestroy().

rfschtkt:
It's a zen thing, it achieves stillness... (I don't know what I'm talking about.)

More prosaically, you can probably grasp code far quicker if you can instantly distinguish between values and variables. If you see `const`, you know you wouldn't have overlooked `pix.reset(pixConvertTo8(pix.p(), false));` somewhere between the definition of `pix` at the top of the function and its use half a page lower, because the compiler already checked that for you. So why not use it wherever you can? And that's even before potential concurrency, where invariability _really_ shines, and where `const` becomes even more valuable (pun not intended)!

Unfortunately, in C/C++, `const` was an afterthought, with a (5+1)-character handicap for values, unlike, e.g., Scala's `val`/`var` equal start. That's why I compensate or overcompensate by documenting variables, preferably providing a rationale.

amitdo:
Those `/*non-const*/` comments looks ugly to me and make the code less readable.

rfschtkt:
You mean `std::unique_ptr</*non-const*/ T>`? Yeah, that gets unwieldy.

Maybe I can do better:
```
$ grep 'UPC\|UPNC' */baseapi.cpp
template<typename T> using UPNC = std::unique_ptr</*non-const*/ T>;
template<typename T> using UPC  = std::unique_ptr<const T>;
  /*used with reset()*/ UPNC<PageIterator> page_it(GetIterator());
    if (! UPC<PageIterator>(AnalyseLayout())) {
  const UPNC<ResultIterator> it(GetIterator());
    const UPC<char[]> para_text(it->GetUTF8Text(RIL_PARA));
  const UPNC<ResultIterator> res_it(GetIterator());
      const UPC<char[]> grapheme(res_it->GetUTF8Text(RIL_SYMBOL));
  const UPNC<ResultIterator> res_it(GetIterator());
      tsv_str += UPC<char[]>(res_it->GetUTF8Text(RIL_SYMBOL)).get();
      const UPNC<char[]> text(it->GetUTF8Text(RIL_SYMBOL));
  const UPC<char[]> text(GetUTF8Text());
  const UPC<PageIterator> it(AnalyseLayout());
$ 
```

amitdo:
>\+  \/*used with reset()\*/ std::unique_ptr</*non-const*/ PageIterator> page_it(GetIterator());

Here you started a line of code with a comment. This makes my brain tread the whole line as a comment. You also implanted a comment in the middle of line, which makes it even worst.

IMO, comments should be put before code block/line or at the end of line.

rfschtkt:
Then you should re-educate your brain, because only `//` does that.

Also, syntax highlighting is not frivolous. Use an editor that supports it!

rfschtkt:
Eliminated even more *Destroy() calls, squashed commits for "RAII: Leptonica types".

If you like to have 159 fewer lines of code (after the addition of ccutil/raiileptonica.h), please give me a clear path to get this accepted, otherwise I'll probably just close this by the end of the month.

Shreeshrii:
@theraysmith  for review and response.

egorpugin:
Is this mostly for Pix raii?
If so I'd like to rename the PR because there are a lot of other places to be updated with raii.

amitdo:
The first two commits are not related to Leptonica.

```
RAII: TessBaseAPI::AnalyseLayout()
ee1a62a
RAII: TessBaseAPI::GetIterator()
25903b3
```

DanBloomberg:
@theraysmith 

Perhaps Ray could comment on this approach with const data.

As essentially a C programmer, this looks like a lot of potential trouble for working Tesseract code.
And as mentioned about 4 years ago, leptonica does not make const declarations except for C strings.
So you can call a leptonica function that takes a C string with either a const or non-const string and
it all works fine.

But if some tesseract function takes a const Boxa, for example, and it calls leptonica functions that
don't declare these as const (and none of them do), it will not compile.  For any leptonica function
that takes a Boxa, you will have to cast the const away with
      const_cast<Boxa *>(boxa)
Under what circumstances would you want Tesseract code to be littered with this?

amitdo:
@DanBloomberg,

Ray is not active in this project for several years. His absent might be correlated to the fact he moved from Google to DeepMind, according to his CV.

stweil:
> The first two commits are not related to Leptonica.

@rfschtkt, I think we can merge those two commits separately. Do you want to send a pull request for them?

Then the PR here can be updated to include only Leptonica related RAII changes.

rfschtkt:
@stweil, I have not looked at this in years, but feel free to just steal those two commits.

egorpugin:
Closing this as conflicts are too big now.
The PR is still here in archive for any references.

egorpugin:
Initial stubs for Pix wrapper are landed in https://github.com/tesseract-ocr/tesseract/commit/ce6e2f18217f1c3a71d6816916b6ff6e048a0cfd https://github.com/tesseract-ocr/tesseract/commit/a792b67983930bb92ff2a40d2a2f1f81d059557d
Raii will be added eventually.

stweil:
> I have not looked at this in years, but feel free to just steal those two commits.

Done, thank you (see commits d71413f4aa40b167624acbfdf71c65d886480444 and 6bbfef7c8515fea24dacde7e8b8239fcec4c4fa6).

PR #942, opened 2017-05-22
Optimize LSTM code for builds without OpenMP

stweil:
The constant value kNumThreads is not only used to configure the number
of threads but also to allocate vectors used in those threads.

There is only a single thread without OpenMP, so it is sufficient to
allocate vectors with only one element in that case.

Replace also the upper limit in the for loops by the known vector size.


stweil:
Updated PR (fixed syntax errors).

PR #953, opened 2017-05-25
Update documentation for installation

stweil:
Add two build dependencies which were missing and update the hints for
building ScrollView.jar.

### DIFF
--- a/INSTALL.GIT.md
+++ b/INSTALL.GIT.md
@@ -9,6 +9,8 @@ before new build.
 Known dependencies for training tools (excluding leptonica):
  * compiler with c++11 support
  * autoconf-archive
+ * automake
+ * pkg-config
  * pango-devel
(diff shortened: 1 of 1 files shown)


Shreeshrii:
There are multiple pages with installation instructions and they are sometimes in conflict.

https://github.com/tesseract-ocr/tesseract/blob/master/INSTALL
https://github.com/tesseract-ocr/tesseract/blob/master/INSTALL.GIT.md

https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation
https://github.com/tesseract-ocr/tesseract/wiki/Compiling

@stweil It will be great if we can merge all current and correct info in one page and then redirect the other pages to it. That way only one page needs maintaining on regular basis.

Also readme.md can be linked to that page.

Thaks!

Issue #644, opened 2017-01-09
LSTM: Training - Eval not run from trainer


Shreeshrii:
See comment from Ray at 
https://github.com/tesseract-ocr/tesseract/issues/542#issuecomment-265569774

> it doesn't run the eval from the trainer.
It doesn't harm the tutorial, but will be required before people start serious training.

xlight:
with --eval_listfiles , always coredump. 😭

lstmtraining: ../ccutil/genericvector.h:696: T& GenericVector<T>::operator[](int) const [with T = char]: Assertion `index >= 0 && index < size_used_' failed.

Shreeshrii:
@theraysmith Any update on this?

Shreeshrii:
@stweil Have you come across this problem while training? Any solution?

stweil:
No, I did not see that assertion up to now. How can it be reproduced?

Shreeshrii:
Please try the following making appropriate changes for your fonts directory and tessdata directory

```
training/tesstrain.sh \
  --fonts_dir /c/Windows/Fonts \
  --tessdata_dir ./tessdata \
  --training_text ../langdata/ara/ara.training_text \
  --langdata_dir ../langdata \
  --lang ara  \
  --linedata_only \
  --noextract_font_properties \
  --exposures "0"    \
  --fontlist "Arial" \
  --output_dir ~/tesstutorial/aratest
  
training/tesstrain.sh \
  --fonts_dir /c/Windows/Fonts \
  --tessdata_dir ./tessdata \
  --training_text ../langdata/ara/ara.training_text \
  --langdata_dir ../langdata \
  --lang ara  \
  --linedata_only \
  --noextract_font_properties \
  --exposures "0"    \
  --fontlist "Arial" \
  "Arial Unicode MS" \
  "Calibri" \
  "Courier New" \
  --output_dir ~/tesstutorial/araeval

mkdir -p ~/tesstutorial/aratuned_from_aratest 

combine_tessdata -e ../tessdata/ara.traineddata \
  ~/tesstutorial/aratuned_from_aratest/ara.lstm
  
lstmtraining \
  --continue_from ~/tesstutorial/aratuned_from_aratest/ara.lstm \
  --train_listfile ~/tesstutorial/aratest/ara.training_files.txt \
  --eval_listfile ~/tesstutorial/araeval/ara.training_files.txt \
  --model_output ~/tesstutorial/aratuned_from_aratest/aratuned \
[...]

PR #979, opened 2017-06-05
Update from Leptonica 1.74.1 to 1.74.2

stweil:
The newer version contains fixes for the pixUnsharpMaskingGray*
functions which are relevant for Tesseract (used in ImageData::PreScale
which calls pixScale).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,8 @@ before_install:
 install:
   #- if [[ $LINUX && "$CXX" = "g++" ]]; then export CXX="g++-6" CC="gcc-6"; fi
   - if test ! -d leptonica-$LEPT_VER/src; then curl -Ls https://github.com/DanBloomberg/leptonica/archive/$LEPT_VER.tar.gz | tar -xz; fi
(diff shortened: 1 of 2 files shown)


stweil:
See [Dan's comment](https://github.com/tesseract-ocr/tesseract/issues/644#issuecomment-306073620) for issue #644.

stweil:
Should we refuse builds with Leptonica < 1.74.2? Then additional changes for `CMakeLists.txt` and `configure.ac` would be needed. The newer version not only fixes problems with undefined memory for `lstmtraining`, but also for `tesseract --oem 1`.

stweil:
Tesseract cmake builds are broken with Leptonica 1.74.2 because Leptonica now requires an installation: the Leptonica_INSTALL_DIRS refer to /usr/local/include. Therefore Travis CI fails. @egorpugin, how should we fix that?

egorpugin:
Someone broke my leptonica cmake scripts? :)

egorpugin:
> Should we refuse builds with Leptonica < 1.74.2?

We can do anything, because it's dev branch.

PR #980, opened 2017-06-05
Remove portability code which is no longer needed

stweil:
MS Visual Studio does not provide that function, but can use strtok_s
which does exactly the same.


stweil:
Untested with MS C++, so please wait until AppVeyor CI‌ passed successful.

PR #1014, opened 2017-06-29
Download the leptonica source from github

elopio:
1.74.2 is no longer available from the leptonica website. But anyway, it seems safer going forward to download it from github. It's https, and it won't disappear as easily. Also, this is the same source used by travis, so there's less chance of shipping something untested.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -30,5 +30,5 @@ parts:
       - libcairo2-dev
     after: [leptonica]
   leptonica:
-    source: http://www.leptonica.org/source/leptonica-1.74.2.tar.gz
+    source: https://github.com/DanBloomberg/leptonica/archive/1.74.2.tar.gz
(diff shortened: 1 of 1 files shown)


amitdo:
CC: @DanBloomberg

DanBloomberg:
Just to be clear, the tarballs on leptonica.org are all configure/make ready.  On github, releases 1.73 - 1.74.2 are checkpoints of the github source, so one needs to use autotools to generate configure, etc.  Releases 1.74.3 and going forward are configure/make ready tarballs, generated directly from github source but with the necessary added files.

1.74.3 was flawed, in that I had included several files that should have been removed after testing.

elopio:
Thanks! Now you can install tesseract from the edge channel:

    sudo snap install tesseract --edge

And it will be autoupdated every time master changes. I will do some testing before moving it to stable.

Issue #1016, opened 2017-07-01
Two error messages do not print a new line


hotchkiss87:
### Environment
* **Tesseract Version**: 4.00.00alpha or 4.00.00dev
* **Platform**: Linux [hostname] 4.11.6 # 1 SMP Mon Jun 19 10:07:48 EDT 2017 x86_64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz GenuineIntel GNU/Linux

### Current Behavior:
Error messages 'Please call SetImage before attempting recognition.' and 'Please call Init before attempting to set an image.' do not print a new line.
### Expected Behavior:
They should appear the same as all of the other messages, with a new line after. It makes the error logs much easier to read too.
### Suggested Fix:
Add newline character to end of string.

PR #1017, opened 2017-07-01
Add new line to a few error messages.

hotchkiss87:
Fix for #1016


Shreeshrii:
\n also need to be added to messages from
https://github.com/tesseract-ocr/tesseract/blob/master/lstm/lstmtrainer.cpp
and
https://github.com/tesseract-ocr/tesseract/blob/master/training/lstmtester.cpp

example of runon messages:

```
At iteration 67503/153800/153948, Mean rms=0.091%, delta=0.592%, char train=1.988%, word train=7.408%, skip ratio=0%,  New worst char error = 1.988At iteration 67166, stage 1, Eval Char error rate=9.8022655, Word error rate=39.409614 wrote checkpoint.
```

Issue #1031, opened 2017-07-11
Not a bug: Cannot delete the tesseract::TessBaseApi instance


LeifUK:
Not a bug: This was due to using a different Visual Studio platform toolset for the app and the DLL. Apparently even when VS upgrades a project to VS 2013, it leaves the toolset at VS 2010. 

### Environment

\Tesseract-4.00.00alpha\
Windows 10 64 bit Visual Studio 2013 building the Win32 DLL. 

### Current Behavior:

Create an instance, call Init, then delete the instance, an exception is thrown. 

### Expected Behavior:

No exception! 

### Suggested Fix:

I have not yet found the cause of the exception.

LeifUK:
I have created a Windows console app which does no more than new tesseract::TessBaseAPI() and then delete the instance, and it throws an exception on the delete. This must be something obvious but I can't see the cause.

Issue #1030, opened 2017-07-11
Need a free call to match  GetUTF8Text();


LeifUK:
### Environment

Tessearct 4.00.00alpha
Windows 10 64 bit Visual Studio 2013 Win32 build. 

This issue is also present in earlier versions of Tesseract including 3.0.4 (which I fixed for our own use). 

### Current Behavior:

The GetUTF8Text() method (and other similar methods) allocates a buffer, but there is no call to free the memory, and it cannot be freed by the caller without heap corruption. 

### Suggested Fix:

Add a method called FreeXXText() (or whatever name is preferred) which does no more than free the memory along these lines: 

void TessBaseAPI::FreeXXText(char *text)
{
	delete[] text;
}

Deleting a null pointer is not a problem (on Windows anyway).

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/10779bd9e50d/api/baseapi.cpp#L1273

I assume it is a Windows only issue similar to https://github.com/tesseract-ocr/tesseract/issues/1029#issuecomment-314152843, right?

vidiecan:
This happens when you do **not** use CRT dynamic link and the heaps get confused. In projects using many dependencies, this can get tricky because all the dependencies should use the same (unless you really know what you are doing).

See https://msdn.microsoft.com/en-us/library/abx4dbyh(v=vs.80).aspx but also 
https://stackoverflow.com/questions/10820114/do-statically-linked-dlls-use-a-different-heap-than-the-main-program

In vsprops (or directly in GUI) you should use something like
```
    <ClCompile>
      <RuntimeLibrary>MultiThreadedDebugDLL</RuntimeLibrary>
    </ClCompile>
```

amitdo:
Another link:
https://msdn.microsoft.com/en-us/library/ms235460.aspx

LeifUK:
That's right, the app and DLL have their own heaps.

egorpugin:
+1
Never like this allocated `char *` value.
It's better to return `std::string`.

amitdo:
You mean `STRING`? :-)

That's what Tesseract uses in the API.

stweil:
I'm afraid that you will get an endless list of problems as long as your program uses more than one C runtime library on Windows. With only a single runtime library, all those problems should be solved.

It is possible to force the MS linker to always use the same msvcrt(d).dll.

[...]

Issue #1029, opened 2017-07-10
Memory management issues in tesseract 4 alpha


LeifUK:
### Environment
Tesseract4.00.00alpha
Windows 32-bit 

### Current Behavior:
Throws an exception in TessBaseAPI::ProcessPage()

### Expected Behavior:
No exception. 

### Suggested Fix:

The exception is thrown in imagedata.cpp in this method when calling free(data): 

void ImageData::SetPixInternal(Pix* pix, GenericVector<char>* image_data);

The data memory is allocated in the liblept library and should be freed using a call to the library. I assume the correct call is lept_free.

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/f5c18f78c09ab028791c28c638d5cc2f96c6d6fb/ccstruct/imagedata.cpp#L327

>The data memory is allocated in the liblept library and should be freed using a call to the library. I assume the correct call is lept_free.

You are right.

amitdo:
https://github.com/DanBloomberg/leptonica/blob/d20c5a68746d56a59a9b9e6f28a91364f5dd6e33/src/utils2.c#L128

Seems to only affect Windows.

LeifUK:
And another one in the same vein, not sure if this should be here or in a new thread. In this method: 

void ScrollView::Image(struct Pix* image, int x_pos, int y_pos)

The free(data) should be lept_free(data). 

Using Visual Studio 2013 on Windows 10 to create a 32 bit build.

stweil:
I now had a closer look on this issue and came to a different result. `ImageData::SetPixInternal` calls `pixWriteMem` which calls different functions depending on the image format. For JPEG, it calls `pixWriteMemJpeg` which calls `l_binaryReadStream` which allocates the memory using `LEPT_CALLOC` which is defined to be equal with `calloc`. So using `free` is not wrong. `lept_free` also basically calls `free`, so it should not make a difference whether `free` or `lept_free` is used.

@LeifUK, does the Visual Studio build use a different definition for `LEPT_CALLOC`, different implementations of `lept_free` or anything else which is different? Why does it throw an exception at all? And what kind of exception is that? Maybe it also depends on the image format. Which one did you process to get the exception?

LeifUK:
The exception is heap corruption. It occurs because in Windows the app and the DLL each have their own heap, therefore you cannot allocate in the DLL and free in the app. So yes this is a Windows issue. 

That is a good point that my app could call lept_free.

[...]

PR #1041, opened 2017-07-16
Use lept_free to free memory allocated by Leptonica

stweil:
This fixes problems on Windows when Tesseract and Leptonica use different
C runtime libraries.


amitdo:
Should we provide `tess_delete()`?

stweil:
And `tess_memcpy`, `tess_printf`, `tess_fopen`, ...? Definitely no. I like standards very much and prefer using them. That's why I try to add POSIX data types to Tesseract (a project that will hopefully be finished in the next five years).

amitdo:
See #1030, #1031, which like #1029 were opened by @LeifUK.

amitdo:
>That's why I try to add POSIX data types to Tesseract (**a project that will hopefully be finished in the next five years**).

😆

Issue #1047, opened 2017-07-20
error:'string' does not name a type


atuc17:
I guess the problem is that file unichar.h missed using namespace std;

In file included from /usr/local/include/tesseract/ltrresultiterator.h:26:0,
                 from /usr/local/include/tesseract/resultiterator.h:26,
                 from /usr/local/include/tesseract/baseapi.h:31,
                 from ../scr/detect_text.cpp:1:
/usr/local/include/tesseract/unichar.h:164:10: error: ‘string’ does not name a type
   static string UTF32ToUTF8(const std::vector<char32>& str32);
          ^~~~~~
`------------------------`

### Environment

* **Tesseract Version**:  4.00.00
* **Commit Number**: 
* **Platform**:  ubuntu 17.04

amitdo:
See #1045. Try again with the latest commit.

amitdo:
@zdenop, please close this issue.

PR #1049, opened 2017-07-23
Update unichar.h

atuc17:
add `using namespace std;` into unichar.h

### DIFF
--- a/ccutil/unichar.h
+++ b/ccutil/unichar.h
@@ -29,6 +29,7 @@
 // Maximum number of characters that can be stored in a UNICHAR. Must be
 // at least 4. Must not exceed 31 without changing the coding of length.
 #define UNICHAR_LEN 30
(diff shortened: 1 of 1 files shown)


stweil:
Could you please test latest Tesseract and see whether this pull request is still needed? If not, please close it (and also issue #1047).

fifothekid:
I think that you should replace string with std::string on line:
https://github.com/tesseract-ocr/tesseract/blob/master/ccutil/unichar.h#L164
It worked for me with this change

fifothekid:
Created pull request
https://github.com/tesseract-ocr/tesseract/pull/1082

amitdo:
See https://github.com/tesseract-ocr/tesseract/pull/1082#issuecomment-323009498

amitdo:
@zdenop, I suggest to close this PR.

PR #1064, opened 2017-08-04
Fix broken build for Windows

stweil:
Windows does not provide a mkdir function with two parameters.


stweil:
Currently the file `training/lang_model_helpers.cpp` is only compiled when building with autotools, but not when building with cmake (that's strange and needs more investigation). Continuous integration uses cmake, so it does not get that build error.

amitdo:
Ray never touches the CMake build files.

PR #1072, opened 2017-08-10
List available languages recursively

stweil:
Tesseract supports hierarchies of languages and uses them since
the new files best/*.traineddata were added.

Now `tesseract --list-langs` also shows any traineddata files in
subdirectories of the tessdata directory.


stweil:
This would work with Tesseract 3 as well. Should it be backported?

PR #1076, opened 2017-08-12
move code around so that list-langs will work without an English traineddata file

chrismamo1:
This causes errors caught by `TessBaseAPI::Init` in `tesseractmain.cpp:main` to not cause an exit immediately, though I doubt that this will cause serious problems.


amitdo:
Suggested solution:

1) Make addAvailableLanguages() a global function under tesseract namespace.
2) Call it from main().

stweil:
The problem is `tesseract_->datadir` which is needed by `addAvailableLanguages`.

amitdo:
Yeah, I though you can get it from main(), but didn't take into account the code in tesseract/ccutil/mainblk.cpp

amitdo:
There are some special init methods. Another one is needed for this case.

PR #1077, opened 2017-08-12
WIP: Chore/cleanup compiler warnings

chrismamo1:


arch/dotproductavx.cpp
```diff
@@ -90,6 +90,8 @@ double DotProductAVX(const double* u, const double* v, int n) {
   // instruction, as that introduces a 70 cycle delay. All this casting is to
   // fool the intrinsics into thinking we are extracting the bottom int64.
   auto cast_sum = _mm256_castpd_si256(sum);
+#pragma GCC diagnostic push
```
amitdo: I was wondering if clang supports those gcc pragmas.
Here's the answer:

https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas

>Controlling Diagnostics via Pragmas

>Clang can also control what diagnostics are enabled through the use of pragmas in the source code. This is useful for turning off specific warnings in a section of source code. Clang supports GCC’s pragma for compatibility with existing source code, as well as several extensions.

chrismamo1:
Sorry about the confusion, but I just wanted to see if the CI caught any breaking changes, I still have a few more warnings to fix. I suppose I can make another PR when those are ready to upload?

PR #1087, opened 2017-08-18
Update colfind.cpp

sturkmen72:
prevents "Empty page!!" messages


sturkmen72:
hi i am new on using tesseract
i am using OpenCV text module calling tesseract305.dll

in release mode i am getting "Empty page!!" messages so i did this fix in my local.
it is up to you deciding if it is useful for public.

thank you

zdenop:
Please provide user case where this is problem

PR #1098, opened 2017-08-27
Update colfind.cpp

sturkmen72:
hi i am new on using tesseract
i am using OpenCV text module calling tesseract400.dll

in release mode i am getting "Empty page!!" messages so i did this fix in my local.
it is up to you deciding if it is useful for public.

thank you


zdenop:
Duplicate to #1087

Issue #781, opened 2017-03-23
preserve_interword_spaces option not working on 4.00alpha


abieler:
I am running the same command line as for the 3.x versions with 

-c preserve_interword_spaces=1

as option. The resulting text does not preserve the white spaces, which it correctly did for the 3.x version.

Thanks
Andre

MoinulJoje:
I am also facing the same problem. After using "preserve_interword_spaces", there is actually no significant difference noticed between the normal OCR and "preserve_interword_spaces=1" parameterized OCR.
Is there any solution?

mm-manu:
Same problem here. "preserve_interword_spaces" has no effect.
Also many other parameters do not work in v4.0 :/

s-asish:
I have the same problem in version 4.0. I have tried with version 3.0.2 - It does not have this option.

mingodad:
I also noticed this problem and it seems related with the trained data, because if I use tesseract 4 with the trained data of 3.05 I do get the interword spaces.
```
#!/bin/sh
export PATH=$HOME/local/tesseract/bin:$PATH
export LD_LIBRARY_PATH=$HOME/local/tesseract/lib:$LD_LIBRARY_PATH
#ubuntu tesseract 3 trained data
export TESSDATA_PREFIX=/usr/share/tesseract-ocr
#export TESSDATA_PREFIX=$HOME/local/tesseract/share/tessdata
tesseract $*

#tesseract -l spa -psm 4 $1 scanned
#mytesseract -c preserve_interword_spaces=1 -l spa  $1 scanned
```

amitdo:
>I also noticed this problem and it seems related with the trained data, because if I use tesseract 4 with the trained data of 3.05 I do get the interword spaces.

With 4.00, If you don't use the --oem option, the default oem will be used:

>3    Default, based on what is available.

The traineddata for 3.05 does not have lstm data, so oem 3 in this case is equivalent to:

>0    Original Tesseract only.

mingodad:
Thanks for pointing out !
I dived on tesseract source code and found an almost solution to the preserve_interword_spaces problem, it seems that when transferring the words the spaces were not transfered see patch bellow.
With this patch the output is almost identical with the 3.05 version except for the missing spaces for the first column (need more research to see where the first/second word is transfered and why the spaces/blanks are not).
```
@@ -1329,11 +1329,11 @@ void PAGE_RES_IT::ReplaceCurrentWord(
[...]

PR #1109, opened 2017-09-06
Fix to preserve_interword_spaces work again

mingodad:
See this comment https://github.com/tesseract-ocr/tesseract/issues/781#issuecomment-323304593


zdenop:
Can you please provide user case that demonstrate problem&solution?

Shreeshrii:
@zdenop Please see https://github.com/tesseract-ocr/tesseract/issues/781

This is a problem only with the master branch/4.00.00alpha.

PR #1112, opened 2017-09-07
CMake: Add -mavx and -msse4.1 compile flags on unix

Sineaggi:
Otherwise we abort with the message `DotProductAVX can't be used on Android`.


Sineaggi:
Tested on macOS Sierra 10.12.6 and Alpine Linux 3.3.

egorpugin:
Thanks! Merged.

Issue #1074, opened 2017-08-11
Method WordFontAttributes does not work


czchao:
### Environment

* **Tesseract Version**: tesseract 4.00.00alpha
* **Commit Number**: 8e55e52be749b3faccca8ae41abdc0e3d3c7f887
* **Platform**: Ubuntu 16.04.1

### Current Behavior:
Method WordFontAttributes returns null if using tesseract 4.00.00alpha with 4.00 tessdata, but it returns font name if using tesseract 4.00.00alpha with 3.04.00 tessdata. The test image link is [eurotext.tif](https://github.com/sirfz/tesserocr/blob/master/tests/eurotext.tif)
I first met this problem when I use tesserocr [tesserocr#68] .(https://github.com/sirfz/tesserocr/issues/68)
### Expected Behavior:
With method WordFontAttributes we can get correct font attributes of recognized words.

amitdo:
The new LSTM engine does not support this feature and probably won't support it any time soon.

phildrip:
Is there an alternative way to get font sizing etc? Do you mean that just this method won't be supported, or the feature in general?

amitdo:
>Is there an alternative way to get font sizing etc?

You can still use --oem 0 with trainedata from here: https://github.com/tesseract-ocr/tessdata.
Note that the traineddata in the 'best' folder won't work with --oem 0.

amitdo:
>Do you mean that just this method won't be supported, or the feature in general?

I have reasons to believe that the new LSTM engine is unlikely to have a feature that includes font identification (name and properties like is_bold) in the near future.

Important note: I'm a contributer from the community, and the main developer not always shares all his plans for upcoming release(s) with the community.

phildrip:
Thanks for the reply! It looks like the old ocr engine is going to be removed, though (issue #707)... And does using `OcrEngineMode 0` mean the behaviour is the same as v3?

What I'm getting to is:
1. I need to be able to extract font size information (font names aren't so useful) - is there any way at all of doing so with LSTM/v4?
2. If I use `OcrEngineMode 0` to be able to get this info, will that be removed from v4 at a later date?
3. Is there any advantage to using v4 with `OcrEngineMode 0` vs v3.05?

Thanks again for the help!

amitdo:
>It looks like the old ocr engine is going to be removed, though (issue #707)... 

It not known when exactly it will be removed. Until then you can still use it.

[...]

PR #1125, opened 2017-09-13
Prevent uninitialized integer from being used for x_fsize output

ftzdomino:
If font_name is false, then res_it->WordFontAttributes and pointsize is uninitialized.  This results in seemingly random data being written to the hocr output for x_fsize.  This change moves the output of x_fsize into the check that verifies the function that initializes pointsize actually ran, preventing the output of the uninitialized value.

### DIFF
--- a/api/baseapi.cpp
+++ b/api/baseapi.cpp
@@ -1526,9 +1526,9 @@ char* TessBaseAPI::GetHOCRText(ETEXT_DESC* monitor, int page_number) {
     if (font_info) {
       if (font_name) {
         hocr_str += "; x_font ";
-        hocr_str += HOcrEscape(font_name);
+        hocr_str += HOcrEscape(font_name);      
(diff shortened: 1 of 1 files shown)


stweil:
Do you have a simple test case where there is no `font_name`?

amitdo:
I'm sorry but that's not the right solution.

amitdo:
Here is Ray's suggestion: 
https://github.com/tesseract-ocr/tesseract/issues/1074#issuecomment-327825021

And here is mine:
https://github.com/tesseract-ocr/tesseract/issues/1074#issuecomment-327831585

amitdo:
>so I suggested to move pointsize to the RowAttributes() method.

Actually, it's better to use function overloading for RowAttributes() to not break the C++ API.

amitdo:
The issue that this PR tries to solve was fixed in #1173.

PR #1126, opened 2017-09-13
Fix unittest/apiexample_test

stweil:
The test failed for out-of-tree builds with autotools because it could not find the required files.
It also failed when the "en_US.UTF-8" locale was missing.

Now it passes. This completes my previous PR #1123.


Shreeshrii:
Thanks, @stweil.

PR #1134, opened 2017-09-16
change type to UChar32 to fix IsValidCodepoint

Shreeshrii:
Fixes issue https://github.com/tesseract-ocr/tesseract/issues/1114


stweil:
This change does not look like the right solution for an existing problem. I'm afraid it should be reverted.

Shreeshrii:
@zdenop Stefan is right. 

I reverted this change and the training process is still working now. So, it must have been something else that changed in system for it to work.

Please revert. Thanks!

Issue #928, opened 2017-05-17
RFC: Multipage Feature


Shreeshrii:
Copied from: https://github.com/tesseract-ocr/tesseract/pull/911#issuecomment-302068376

> @stweil - How would I start Tesseract to process page1.png and page2.png in a single run?
> 
> @amitdo - Prepare a text file that has the path to each image:
> 
> path/to/1.png
> path/to/2.png
> path/to/3.tiff
> 
> Save it, and then give its name as input file to Tesseract.
> 
> tesseract savedlist output
> 
> @stweil - Thank you, good to know that. It looks like the ChangeLog, other documentation and the program help text need an update.
> 
> Currently all pages are written to one output file (per format). Some formats include page information (hOCR, PDF). Others like TXT don't, but could use a page separator character (ASCII 0x0C = FF). Would it help to support an output base parameter with placeholders like page number or image base name to generate one output file per input image?
> 
> The multi-page feature was added in 2014 by commit 25a8c7b.

Shreeshrii:
@stweil 

>Would it help to support an output base parameter with placeholders like page number or image base name to generate one output file per input image?

I can think of two cases for using this.

1. A document for which each page is available as a separate image.
2. Separate documents images which need to be run in a batch

For the first, current processing is sufficient, maybe with an additional  page separator character .

For the second, it would be useful to have the option you suggest.

stweil:
An implementation could also get pairs of image file and output base from the file list instead of image files only. That would allow maximum flexibility.

Shreeshrii:
savedlist

```
Sachitra_Saraswati_Prasad_004932_HR-g4_page0001_1L.tif
Sachitra_Saraswati_Prasad_004932_HR-g4_page0002_2R.tif
Sachitra_Saraswati_Prasad_004932_HR-g4_page0004_2R.tif
Sachitra_Saraswati_Prasad_004932_HR-g4_page0005_1L.tif
Sachitra_Saraswati_Prasad_004932_HR-g4_page0005_2R.tif
```

1. `tesseract savedlist output`

Current implementation creates output file `output.txt` with concatenated OCR text.

Console displays filenames being processed

```
Tesseract Open Source OCR Engine v4.00.00alpha with Leptonica
Page 0 : Sachitra_Saraswati_Prasad_004932_HR-g4_page0001_1L.tif
Page 1 : Sachitra_Saraswati_Prasad_004932_HR-g4_page0002_2R.tif
[...]

PR #1140, opened 2017-09-19
Remove Tesseract parameter "include_page_breaks" and use FF by default

stweil:
Now Tesseract adds a page break (normally form feed) by default.

It is still possible to suppress page breaks by setting an empty
page_separator.


stweil:
See discussion for issue #928.

Issue #1147, opened 2017-09-21
"Phase UP: Generating unicharset and unichar properties files" ERROR


ivanzz1001:
### Environment

* **Tesseract Version**: tesseract 4.00.00alpha
* **Commit Number**:  2cc531e
* **Platform**: Linux localhost.localdomain 3.10.0-514.el7.x86_64 #1 SMP Tue Nov 22 16:42:41 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

### Current Behavior:
I excute the following command:
<pre>
training/tesstrain.sh --fonts_dir /usr/share/fonts --lang chi_sim --linedata_only \
  --noextract_font_properties --langdata_dir ../langdata \
  --tessdata_dir ../tessdata \
  --fontlist "AR PL UKai CN" \
  "AR PL UKai HK" \
  "AR PL UKai TW" \
  "AR PL UKai TW MBE" \
  "AR PL UMing CN Semi-Light" \
  "AR PL UMing HK Semi-Light" \
  "AR PL UMing TW MBE Semi-Light" \
  "AR PL UMing TW Semi-Light" \
  "Arial Unicode MS" \
  "FangSong" \
  "KaiTi" \
  "LiSu" \
  "Microsoft YaHei" \
  "Microsoft YaHei Bold" \
  "NSimSun" \
  "Noto Sans SC" \
  "Noto Sans SC Bold" \
  "Noto Sans SC Heavy" \
  "Noto Sans SC Medium" \
  "Noto Sans SC Medium" \
  "Noto Sans SC Semi-Light" \
  "Noto Sans SC Semi-Light" \
  "STFangsong" \
  "STKaiti" \
  "STSong" \
  "STXihei" \
  "STXinwei" \
  "STZhongsong" \
  "SimHei" \
  "SimSun" \
  "WenQuanYi Micro Hei" \
  "WenQuanYi Micro Hei Mono" \
  "WenQuanYi Zen Hei Medium" \
  "WenQuanYi Zen Hei Mono Medium" \
  "WenQuanYi Zen Hei Sharp Medium" \
  "YouYuan" \
  --output_dir ../tesstutorial/chieval \
  --overwrite
</pre>

But when it goes to Phase UP, it generates the following error:
<pre>
=== Phase UP: Generating unicharset and unichar properties files ===
[...]

PR #1153, opened 2017-09-24
Update unicharset_extractor.cpp

ivanzz1001:
fix the "Phase UP: Generating unicharset and unichar properties files" ERROR

Please reference [#1147](https://github.com/tesseract-ocr/tesseract/issues/1147)

### DIFF
--- a/training/unicharset_extractor.cpp
+++ b/training/unicharset_extractor.cpp
@@ -50,7 +50,7 @@ static void AddStringsToUnicharset(const GenericVector<STRING>& strings,
                                      /*report_errors*/ true,
                                      strings[i].string(), &normalized)) {
       for (const string& normed : normalized) {
-        if (normed.empty() || IsWhitespace(normed.c_str())) continue;
+        if (normed.empty() || IsUTF8Whitespace(normed.c_str())) continue;
         unicharset->unichar_insert(normed.c_str());
       }
     } else {
--- a/training/unicharset_extractor.cpp
+++ b/training/unicharset_extractor.cpp
@@ -50,7 +50,7 @@ static void AddStringsToUnicharset(const GenericVector<STRING>& strings,
                                      /*report_errors*/ true,
                                      strings[i].string(), &normalized)) {
       for (const string& normed : normalized) {
-        if (normed.empty() || IsWhitespace(normed[0])) continue;
+        if (normed.empty() || IsWhitespace(normed.c_str())) continue;
         unicharset->unichar_insert(normed.c_str());
       }
     } else {


training/unicharset_extractor.cpp
```diff
@@ -50,7 +50,7 @@ static void AddStringsToUnicharset(const GenericVector<STRING>& strings,
                                      /*report_errors*/ true,
                                      strings[i].string(), &normalized)) {
       for (const string& normed : normalized) {
-        if (normed.empty() || IsWhitespace(normed[0])) continue;
+        if (normed.empty() || IsUTF8Whitespace(normed.c_str())) continue;
```
stweil: The old code tried to check whether the first character of `normed` is a whitespace character.
The new code checks whether `normed` consists only of whitespace characters. If there are two or more characters, that is not the same operation (and it will take more time because not only the first character is examined).

amitdo: `if (normed.empty() || IsUTF8Whitespace(normed[0].c_str())) continue;`

Is that OK?

amitdo: `if (normed.empty() || IsUTF8Whitespace(normed.substr(0,1).c_str())) continue; `

Is that OK?

Any other suggestion?

ivanzz1001: I have printed "normed"，and found it is just a single word read from chi_sim.training_text   every time

stweil: Did you mean "single character"? Then the new code is fine.

ivanzz1001: yes. I think it is ok. And from the context, I think it just want to strip the whitespace

training/unicharset_extractor.cpp
```diff
@@ -50,7 +50,9 @@ static void AddStringsToUnicharset(const GenericVector<STRING>& strings,
                                      /*report_errors*/ true,
                                      strings[i].string(), &normalized)) {
       for (const string& normed : normalized) {
-        if (normed.empty() || IsWhitespace(normed[0])) continue;
+
+       // normed[0] has converted to utf-8
+        if (normed.empty() || IsUTF8Whitespace(normed.c_str())) continue;
```
stweil: In theory that new code should work (although the comment is not quite correct).

The code fails nevertheless because IsUTF8Whitespace() does not work as expected: when a UTF-8 character like "«" is seen, the function calls SpanUTF8Whitespace() which calls IsWhitespace() which should get the 32 bit codepoint 0xc2ab, but it only gets 0x00ab. @theraysmith, this looks like a problem in the UNICHAR class.

stweil: I was wrong. The codepoint for "«" is 0xab. The new code works perfectly.

stweil: @ivanzz1001, could you fix the comment, please (something like "// normed is a UTF-8 encoded string") and maybe squash your commits? Then this PR is ready to get merged.

amitdo:
@zdenop 

You might want to try this:
https://github.com/blog/2141-squash-your-commits
https://help.github.com/articles/about-pull-request-merges/

Note that it will give you the option to change the commit message.


My suggestion:
>Fix encoding issue in unicharset_extractor.cpp (Close #1153)

stweil:
@theraysmith, @zdenop, is this bug fix which is needed for everybody who wants to do training ready to get merged (preferably with squashing the sever commits), or would you prefer a more direct solution (like [this commit](https://github.com/stweil/tesseract/commit/94ea31d4bad6174a8a2999a940bfe76b9ce31f26))?

694376965:
@ivanzz1001, After I changed unicharset_extractor.cpp file, should I build the unicharset_extractor.cpp file to generate a new unicharset_extractor.o file ?

g++ unicharset_extractor.cpp -o unicharset_extractor

use the command ok ?

PR #1082, opened 2017-08-14
Fixed unqualified class "string"

fifothekid:
The unqualified "string" class created class not found error when compiling with OpenCV

PR #1158, opened 2017-09-27
fix string ambiguity in ccutil/unichar.h

greenbrettmichael:
This PR removes a namespace ambiguity in ccutil/unichar.h


stweil:
Technically your pull request is good, but it violates Google conventions.
See pull request #1082 and several related issues.

PR #1163, opened 2017-10-03
Add Leptonica's pkg-config-found library directory to the search path

cysp:
I must confess to not being a CMake expert, but I found [this mailing list post](https://cmake.org/pipermail/cmake-developers/2016-May/028364.html) that seems to confirm my suspicion that the `FindPkgConfig` module doesn't fix up the library search path to allow linking against the libraries it has found.

N.B.: I also needed https://github.com/DanBloomberg/leptonica/pull/282 to get this compiling on macOS.


amitdo:
In general, unless you compile on Windows with MSVC, you should use autotools to build Leptonica and Tesseract.

CC: @egorpugin

egorpugin:
> In general, unless you compile on Windows with MSVC, you should use autotools to build Leptonica and Tesseract.

True.
But in general, PR is ok.

PR #1206, opened 2017-11-10
Fix help message for unicharset_extractor

stweil:
If unicharset_extractor was called without any argument,
a help message was printed by tesseract::ParseCommandLineFlags.

Replace that by the local help message which is better.


amitdo:
@stweil, my suggestion:

```
if (argc == 1) {
  ...
  return EXIT_FAILURE;
}  

tesseract::ParseCommandLineFlags(argv[0], &argc, &argv, true);
return tesseract::Main(argc, argv);
```

stweil:
Yes, that would simplify the code a little bit (without changing functionality).

PR #1209, opened 2017-11-14
fix vs2015(chi_sim) build error

crazypenguincode:
vs2015 enterprise version update3 
use cppan and cmake  find a build error


egorpugin:
This was merged in https://github.com/tesseract-ocr/tesseract/pull/1218

PR #1239, opened 2017-12-20
Update README.md

amitdo:


amitdo:
Here you can see how it will be displayed:
https://github.com/amitdo/tesseract/tree/patch-1

PR #1265, opened 2018-01-11
Added JPEG quality option parameter (-c jpg_quality=n)

tleegwater:
I needed to be able to specify the JPEG quality level in PDF output files, so I made this parameter optional. Default JPEG quality will still be 85.


amitdo:
https://github.com/tesseract-ocr/tesseract/commit/bce2cd5f331b66

amitdo:
If my memory serves me well, @jbreiden (author of the pdf renderer code) didn't like this feature.

tleegwater:
Ah, I see, d0cb107 says so. I guess issue 1300 + 1285 can't be read anymore?

amitdo:
https://web.archive.org/web/20150413012101/https://code.google.com/p/tesseract-ocr/issues/detail?id=1285

https://web.archive.org/web/20150413012101/https://code.google.com/p/tesseract-ocr/issues/detail?id=1300

amitdo:
https://web.archive.org/web/20150413012101/https://code.google.com/p/tesseract-ocr/issues/detail?id=1300
@jbreiden commented:
>I don't have a strong opinion about tessedit_pdf_jpg_quality. It might cause some confusion.

jbreiden:
Let's at least talk about this for a bit. Tesseract tries really hard to inline images instead of transcode them, which means that quality parameter should be rarely used. @tleegwater can you tell us what sort of image files you are feeding to Tesseract? Maybe attach one if possible?

tleegwater:
I'm feeding TIFF file to Tesseract. And I'm aware that as soon as I feed it JPEG, or something else that's supported, Tesseract will try to inline the inputfile.
But even if someone would feed JPEG and set the quality parameter to 50, Tesseract still inlines the image without transcoding to q=50, or q=85 for that matter.

jbreiden:
That's fine, and I'm fine with this changelist.

Can you please tell me what flavor of TIFF you are working with? Uncompressed? LZW? Pack? Deflate? JPEG? CCITT Group 4? You can find out with `tiffinfo` from libtiff, or `identify -verbose` from ImageMagick. I'm considering supporting a few more image format for inline, and this might influence prioritization.

tleegwater:
We use no compression at all for our TIFF's We are generating them ourselves so there's no chance we'll get anything other than Uncompressed.

jbreiden:
Got it. Please be aware that the built in JPEG encoder is standard libjpeg. If you ever want more precise control (for example, turning off chroma subsampling) a fancier encoder like Guetzli, then JPEG encode the images before feeding to Tesseract.

https://research.googleblog.com/2017/03/announcing-guetzli-new-open-source-jpeg.html

zdenop:
Thanks for pointing to this issue. What it better for maintaining compatibility of C-API:

1. `TESS_API TessResultRenderer* TESS_CALL TessPDFRendererCreate(const char* outputbase, const char* datadir, BOOL textonly, int jpg_quality=85);`
2. or to define `TessPDFRendererCreate ` (the old call) and `TessPDFRendererCreate2` (new call with jpg_quality )?

amitdo:
It's not just C API break, it's also a C++ break.
There are at least 2 bindings that use the C++ API directly (Python, R).

amitdo:
As said. I think the C++ code can be changed to not break previous API.
Then you can revert the C API change.

zdenop:
Ok. I got it. First I think about extending API, but it is not need because jpg quality is handled by tesseract parameter....

amitdo:
I didn't test it, but the change LGTM.

PR #1286, opened 2018-01-23
Fix #1222 Unable to detect text in few images when using multi language hints

### LINKED ISSUE #1222: Unable to detect text in few images when using multi-language hints
I am using tesseract-OCR to extract text from images which can contain text in many popular Indian languages (it can also be bi-lingual or multi-lingual in the same image). 
While trying multi-language hints in my command, I often encounter error
`contains_unichar_id(unichar_id):Error:Assert failed:in file ../ccutil/unicharset.h, line 513`

On manually checking that image and providing that language hint works correctly. 

Command resulting in error: 
`tesseract --tessdata-dir /usr/local/share/tessdata/ tamil.jpg -l hin+eng+tam+guj+pan+kan+mal+ben+tel+mar stdout`

Correctly working command: 
`tesseract --tessdata-dir /usr/local/share/tessdata/ tamil.jpg -l tam stdout`

Affected Sample Image: 
tamil

### Environment

* **Tesseract Version**: 
 tesseract 4.00.00alpha
 leptonica-1.74.1
 libjpeg 8d : libpng 1.6.32 : libtiff 4.0.8 : zlib 1.2.8
 Found AVX2
 Found AVX
 Found SSE
* **Commit Number**:  ebbfc3a
* **Platform**: Darwin   - Kernel Version 16.7.0 (MacOS)

### Current Behavior: 
Using multi-language hints in command results in error:
`contains_unichar_id(unichar_id):Error:Assert failed:in file ../ccutil/unicharset.h, line 513`

### Expected Behavior: 
It should extract text matching the language profiles provided in command without error.

### Suggested Fix:
Since directly using text's language profile works correctly, it may have to do something in providing multi-language hints. So, in providing multiple languages this behavior is unexpected and may need correction.

syzer:


amitdo:
I'm not sure if this is the 'right' fix to the issue.

The real source of the issue is somewhere higher in the call chain.

What you are doing here is just 'hiding' a bug. The bug will still be there.

syzer:
@amitdo you are right..
Its just a patch, dosen't fix actually issue
Real fix would be write failing test(dataset) and fix higher in the chain.
However: consider this... this is it actually breaking clients code.
And in my opinion it should just give a warning.
Because if you have a line/part of image that you cannot read... you might still have 10 other lines with valuable information.

amitdo:
I understand your view.

Anyway, I'm not the one who decides here...

TheCreativeMind:
I am struggling with the same error, is there a plan when this PR will be merged?

TheCreativeMind:
As this has been an open issue for quite some time, will this PR ever be merged or will there be an actual fix for the bug in the near future?

amitdo:
There will be bug fix when someone will find a better way to fix the issue...

syzer:
I just use this branch with `brew` and I'm quite happy that this assert is not crasching my code.

amitdo:
@zdenop, please close this. it's not the right fix for the issue.

PR #1328, opened 2018-02-19
Don't try alternate path for tessdata

stweil:
This simplifies the code and the user interface.

### DIFF
--- a/ccmain/tessedit.cpp
+++ b/ccmain/tessedit.cpp
@@ -106,20 +106,10 @@ bool Tesseract::init_tesseract_lang_data(
   // Initialize TessdataManager.
   STRING tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
   if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string())) {
-    // Try without tessdata.
-    m_data_sub_dir.set_value("");
-    main_setup(arg0, textbase);
-    language_data_path_prefix = datadir;
-    language_data_path_prefix += lang;
-    language_data_path_prefix += ".";
-    tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
-    if (!mgr->Init(tessdata_path.string())) {
-      tprintf("Error opening data file %s\n", tessdata_path.string());
-      tprintf(
-          "Please make sure the TESSDATA_PREFIX environment variable is set"
-          " to your \"tessdata\" directory.\n");
-      return false;
-    }
+    tprintf("Error opening data file %s\n", tessdata_path.string());
+    tprintf("Please make sure the TESSDATA_PREFIX environment variable is set"
+            " to your \"tessdata\" directory.\n");
+    return false;
   }
   if (oem == OEM_DEFAULT) {
     // Set the engine mode from availability, which can then be overidden by


stweil:
@egorpugin, do you have an idea why AppVeyor CI fails since 24 days? The build complains about missing `dirent.h`.

Shreeshrii:
@stweil

Are alternate paths required for handling how tessdata files are packaged under Linux.

I think it is 

/usr/share/tesseract-ocr/4.00/tessdata

The old version might have been without the extra subdirectory 4.00

stweil:
A path like `/usr/share/tesseract-ocr/4.00/tessdata` needs special handling during the build in any case. If the build had used `/usr/share/tesseract-ocr/4.00`, then this has to be changed to `/usr/share/tesseract-ocr/4.00/tessdata`.

By the way: @jbreiden @AlexanderP, do we really want to keep that path for the new Debian packages, or should be switch to `/usr/share/tesseract-ocr/4.0/tessdata` or even  `/usr/share/tesseract-ocr/4/tessdata` (and hope that Tesseract will use semantic versioning soon)?

AlexanderP:
The path `/usr/share/tesseract-ocr/4.00` was chosen to avoid conflict when installing libtesseract3, which I had.
If necessary it can be changed.

Shreeshrii:
@AlexanderP

Since the format of old traineddata and 4.0 traineddata is different, it makes sense to have a separate directory for the 4.0 files.

I think @stweil is suggesting using 4.0 or 4 instead of 4.00 as the subdirectory name.

stweil:
Yes, exactly.

AlexanderP:
I got it.
It is necessary to wait for the answer @jbreiden

jbreiden:
I have no preference between `4` `4.0` and `4.00`.  I am a little worried because the Ubuntu 18.04 freeze date is in 8 days. This would be a pretty major change and if we mess it up, there might not be enough time to recover.

   https://wiki.ubuntu.com/BionicBeaver/ReleaseSchedule

jbreiden:
Basically "now or never" if we want to change this for Ubuntu. With some risk if we attempt it.

Shreeshrii:
> Ubuntu 18.04 freeze date is in 8 days.

@jbreiden  traineddata for CJK and thai should also be updated before that - see

https://github.com/tesseract-ocr/tessdata_fast/pull/7
https://github.com/tesseract-ocr/tessdata_fast/pull/8

stweil:
As we have `libtesseract3` and `libtesseract4`, using `4` would be logical for me. That also fits both for semantic versioning and for the current versioning.

stweil:
An unrelated question:

The new Debian tesseract-ocr uses package names for traineddata in two different forms, for example `tesseract-ocr-deu` and `tesseract-ocr-script-frak`. While the first description "tesseract-ocr language files for German" seems fine, I do not understand the description "tesseract-ocr script files for Fraktur". What are "script files"? Who has a better description? When looking closer on the descriptions, both should use the singular, as there is only a single traineddata file.

stweil:
> traineddata for CJK and thai should also be updated before that

@zdenop, that requires merging the related pull requests sent by @Shreeshrii first.

jbreiden:
Script in the linguistic sense. It means not trained to a specific language. tesseract-ocr-script-frak is good for Fraktur writing whether Dutch or German.

As for `4` that is a good choice, but do you think it is important enough to be worth the risk?

stweil:
So the description should be "tesseract-ocr file for Fraktur script".

I think the risk of fixing the path is low. A valid test only requires building the packages locally, installing them on a machine without any other relicts from previous tesseract experiments and running `tesseract --list-langs` there.

Shreeshrii:
>As we have libtesseract3 and libtesseract4, using 4 would be logical for me. That also fits both for semantic versioning and for the current versioning.

@jbreiden What will be tesseract version as per Debian/Ubuntu release? Is it still 4.00alpha?

> description should be "tesseract-ocr file for Fraktur script" or maybe even better "tesseract-ocr data for Fraktur script".

I agree.  eg. for Devanagari script can be used for Hindi, Sanskrit, Marathi, Nepali  languages.

"tesseract-ocr traineddata file for Devanagari script" describes the package better than "tesseract-ocr script files for Devanagari"

Shreeshrii:
Should some of the supported languages also be listed as part of description? There was a question sometime back regarding `lat` language and `Latin` script traineddata files.

eg. for https://packages.debian.org/sid/graphics/tesseract-ocr-script-deva

```
This package contains the data needed for processing images in a particular script (e.g. Hindi, Sanskrit, Marathi, Nepali  languages written in Devanagari script).
```

Please see related PR https://github.com/tesseract-ocr/tessdata_fast/pull/6/files
It includes Ray's description of the script level files.

```
+Initial capitals indicate the one model for all languages in that script. 
+
+**Latin** is all latin-based languages, 
+except vie, which has its own **Vietnamese**.
+
+**Devanagari** is hin+san+mar+nep+eng
+
+**Fraktur** is basically a combination of all the latin-based languages that have an 'old' variant.
+
+Most of the script models include English training data as well as the script, but not for **Cyrillic**, as that would have a major ambiguity problem. 
+
```

AlexanderP:
I updated the description in the rules.
https://github.com/AlexanderP/tesseract-lang-debian/commit/61c6150f90c3fc7a7698f6f71a06b9fd347aa055
I'm waiting for suggestions.

AlexanderP:
For ubuntu lts packages are taken from Debian testing. Packages will switch
from sid to testing after 6 days, after loading into the sid. how long it
takes to synchronize from Debian testing to ubuntu, I do not know. We have
very little time, and there is a chance not to be in time.
At a minimum, need to update tesseract-lang.
it is necessary to make a decision.


22 февр. 2018 г. 10:22 пользователь "Stefan Weil" <notifications@github.com>
написал:

@AlexanderP <https://github.com/alexanderp>, do you think that changing the
tessdata path from /usr/share/tesseract-ocr/4.00 to
/usr/share/tesseract-ocr/4 is still possible before the Ubuntu deadline
(see discussion above)? I could help with testing or send pull requests.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/tesseract-ocr/tesseract/pull/1328#issuecomment-367589746>,
or mute the thread

stweil:
> Basically "now or never" if we want to change this for Ubuntu. With some risk if we attempt it.

I'm afraid "now" is passing by, and we will have to live longer with `4.00`.

AlexanderP:
I do not think that this is a big problem ... It will be possible to change the prefix before the release 4.0, and at any time.

jbreiden:
Attempting official packages updates tonight. Trying to read and understand all the discussion...

jbreiden:
@stweil @Shreeshrii Please look at Alexander's changelog and tell me if you see any trouble. For example, a commit that was reverted.

https://github.com/AlexanderP/dsc-file/blob/master/debian-tesseract/changelog

AlexanderP:
@jbreiden Thanks!
I updated tesseract-lang in PPA.
https://github.com/AlexanderP/tesseract-lang-debian/commit/ec69f6db6a6ddd138f1f21351654d0bd9347d8c3

repositories for rules
https://github.com/AlexanderP/tesseract-debian
https://github.com/AlexanderP/tesseract-lang-debian

amitdo:
>There is probably no time left to fix mistakes, but if someone notices a horrible problem, please tell me right away.

@jbreiden, I don't see why you can't fix mistakes after March 1st according to the Ubuntu release process.

https://wiki.ubuntu.com/UbuntuDevelopment/ReleaseProcess
https://wiki.ubuntu.com/BetaProcess

https://wiki.ubuntu.com/BionicBeaver/ReleaseSchedule

>March 1st [FeatureFreeze](https://wiki.ubuntu.com/FeatureFreeze), [Debian Import Freeze](https://wiki.ubuntu.com/DebianImportFreeze)
>
>March 8th [Beta 1 Freeze](https://wiki.ubuntu.com/Beta1Freeze) (for opt-in flavors; Monday), [Beta 1](https://wiki.ubuntu.com/Beta1Release) (for opt-in flavours)
>
>March 29th [DocumentationStringFreezep](https://wiki.ubuntu.com/DocumentationStringFreeze)
>
>April 5th Final Beta Freeze (Monday), Final Beta

Shreeshrii:
@AlexanderP    @jbreiden

The list of languages in https://github.com/tesseract-ocr/tesseract/blob/
master/doc/tesseract.1.asc is NOT complete.

I just noticed this. It says:

```
There are currently language packs available for the following languages (in
 https://github.com/tesseract-ocr/tessdata):
```

That should be changed to tessdata_fast.

The list should add all the new languages and scripts. I see the list in

https://github.com/AlexanderP/tesseract-lang-debian/blob/
master/debian/lang.sh


​It might be easiest for either you or Alex to update this based on

https://github.com/AlexanderP/tesseract-lang-debian/blob/master/debian/lang.txt
and
https://github.com/AlexanderP/tesseract-lang-debian/blob/master/debian/script.txt

​

Shreeshrii:
I have create a PR - for the language list. Please see
https://github.com/tesseract-ocr/tesseract/pull/1343

ShreeDevi

jbreiden:
We at the time limit, and I do not dare make any more changes.  As @AlexanderP explained, the packages flow from Debian Unstable -> Debian Testing -> Ubuntu. If everything works perfectly, we will just barely meet the deadline. You can watch the flow on the dashboards.

https://tracker.debian.org/pkg/tesseract-lang
https://tracker.debian.org/pkg/tesseract

stweil:
Thank you to both of you!

Shreeshrii:
Thank you all for your efforts to get it packaged to make it easily
accessible to a large number of users.

Some related and unrelated questions...

1. Will this be marked by a tag/release in the repo so that other packagers
eg. Mac, msys2 etc can also update their distributions?

2. Till now the recommendation to get latest version of tesseract was to
build from source. Now, thanks to Alex and Stefan, we have PPA for Ubuntu
and windows binaries.

Should the wiki be changed to give these priority over building from source?

@stweil, could you also build a Mac/osx version?

stweil:
> could you also build a Mac/osx version?

I plan to build one this week.

Shreeshrii:
@stweil Thanks! Please also look at the addition of tessdata in dirname issue.

stweil:
> Please also look at the addition of tessdata in dirname issue.

@Shreeshrii , could you please test whether #1356 fixes that?

Shreeshrii:
@stweil Yes, seems to be ok now after #1356.

Shreeshrii:
>Should the wiki be changed to give these priority over building from source?

I have made some changes to https://github.com/tesseract-ocr/tesseract/wiki

Shreeshrii:
@AlexanderP Please update the PPAs to include #1356. That completes the tessdata_prefix related changes from this PR.

AlexanderP:
@Shreeshrii done

PR #1343, opened 2018-02-23
Update language list based on tessdata_fast

Shreeshrii:
taken from https://github.com/AlexanderP/tesseract-lang-debian/blob/master/debian/lang.txt


AlexanderP:
@Shreeshrii It is necessary to replace - by _ in abbreviations

Shreeshrii:
>It is necessary to replace - by _ in abbreviations of languages

No. Got changed in my global replace in notepad.

Please modify as required.

PR #1353, opened 2018-03-01
Replace log2(n) by local functions

stweil:
This also adds support for environments without a log2 function (Android).

### DIFF
--- a/lstm/lstm.cpp
+++ b/lstm/lstm.cpp
@@ -67,6 +67,27 @@ const double kStateClip = 100.0;
 // Max absolute value of gate_errors (the gradients).
 const double kErrClip = 1.0f;
 
+// Calculate ceil(log2(n)).
+static inline uint32_t ceil_log2(uint32_t n)
+{
+  // l2 = (unsigned)log2(n).
+#if defined(__GNUC__)
+  // Use fast inline assembler code for gcc or clang.
+  uint32_t l2 = 31 - __builtin_clz(n);
+#else
+  if (n == 0) return UINT_MAX;
+  if (n == 1) return 0;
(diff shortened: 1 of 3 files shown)


stweil:
The implementation in `lstm/lstm.cpp` is much faster according to results of a small test program which calls it for all unsigned values: 37 s (old algorithm with log2), 39 s (non gnu code), 0.7 s (gnu code). I expect that the non gnu code is faster for typical integer values which require less loop iterations.

stweil:
What happened to the AppVeyor CI build? It does not look related to my modifications.

zdenop:
Windows/cppan build is broken (actually not tesseract directly but dependency: error C2065: 'FRIBIDI_TYPE_PDF_LRI': undeclared identifier ). 
@egorpugin : can you fix it?

amitdo:
For MSVC you can use:

_BitScanReverse
https://msdn.microsoft.com/en-us/library/fbxyd7zd.aspx

OR

__lzcnt
https://msdn.microsoft.com/en-us/library/bb384809.aspx

egorpugin:
> @egorpugin : can you fix it?

I'm debugging. Looks like bad codegen from MSVC.

egorpugin:
Build is fixed, but 64 has compiler issue.
https://developercommunity.visualstudio.com/content/problem/206985/clexec-bad-codegen-on-x64.html

amitdo:
@stweil, did you see my comment?

stweil:
Yes, I did, but needed to find time for coding and test. See PR #1369. It reduces the time for MS C code from 39 s to 0.7 s (same value as for gcc).

PR #1354, opened 2018-03-02
Remove old code for string class (no longer needed)

stweil:

### DIFF
--- a/ccstruct/ratngs.cpp
+++ b/ccstruct/ratngs.cpp
@@ -201,7 +201,7 @@ WERD_CHOICE::WERD_CHOICE(const char *src_string,
     : unicharset_(&unicharset){
   GenericVector<UNICHAR_ID> encoding;
   GenericVector<char> lengths;
-  string cleaned = unicharset.CleanupString(src_string);
+  std::string cleaned = unicharset.CleanupString(src_string);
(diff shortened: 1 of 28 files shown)


stweil:
Please wait with merging ... more changes are needed.

stweil:
@jbreiden, I am not sure whether [your comment](https://github.com/tesseract-ocr/tesseract/issues/1045#issuecomment-368153043) also covers the code changes which I‌ present here. And what about the definition for `DISABLE_HEAP_LEAK_CHECK` in [training/util.h](https://github.com/tesseract-ocr/tesseract/blob/master/training/util.h#L59)? If that is no longer needed at Google, we could remove `USE_STD_NAMESPACE` completely (and `DISABLE_HEAP_LEAK_CHECK` as well).

jbreiden:
@stweil I have not tested the heap check stuff. Suggest leaving it unless I can confirm. However, I think it is fair to switch the ifdef to make Google use an oddball build rather than the rest of the world.

amitdo:
https://github.com/tesseract-ocr/tesseract/search?utf8=%E2%9C%93&q=%22%23ifdef+GOOGLE%22

https://gperftools.github.io/gperftools/heap_checker.html

stweil:
Thank you. I now use `GOOGLE_TESSERACT` in pull request #1360.

PR #1355, opened 2018-03-02
Fix syntax error.

tdhintz:
Win64 receives syntax error.


amitdo:
> uint32_t int l2 = 0; 

I wonder why gcc and clang don't consider this an error.

stweil:
Both don't use that code. It's in the Windows MSVC part, and I did not notice it because of the broken AppVeyor CI. My own Windows builds use gcc.

PR #1368, opened 2018-03-10
doc: Add missing language to list

stweil:
tessdata_fast includes bre.traineddata.


Shreeshrii:
It is also there in tessdata_best.

https://github.com/tesseract-ocr/tessdata_best/blob/master/bre.traineddata

PR #1369, opened 2018-03-10
lstm: Use MS C intrinsic function for faster calculation of log2

stweil:

### DIFF
--- a/lstm/lstm.cpp
+++ b/lstm/lstm.cpp
@@ -24,6 +24,10 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#if !defined(__GNUC__) && defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


lstm/lstm.cpp
```diff
@@ -24,6 +24,10 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#if !defined(__GNUC__) && defined(_MSC_VER)
```
amitdo: Why not just

>#if defined(_MSC_VER)

stweil: That works typically, too. It's even possible to remove the include statement – obviously it is included somewhere else.

The current code is a little bit conservative. The code later also first checks `__GNUC__`, then `_MSC_VER`, and I wanted to make sure that the required include statement is there.

PR #1380, opened 2018-03-12
configure: Get version string from git or from VERSION file

stweil:
Use git to create the version string if possible.
Otherwise get the version from the VERSION file.


Shreeshrii:
@stweil Please also update ccutil/version.h to use VERSION.

stweil:
That file will be removed in one of the next steps.

PR #1394, opened 2018-03-17
autogen: Report missing requirements

stweil:
autoconf-archive and pkg-config are required, but users often missed that requirements.

The script now detects and reports those missing requirements and removes
the incomplete generated configure script.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -113,6 +113,13 @@ automake --add-missing --copy --warnings=all || bail_out
 echo "Running autoconf"
 autoconf || bail_out
 
+if grep -q AX_CHECK_COMPILE_FLAG configure; then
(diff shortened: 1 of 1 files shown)


zdenop:
thanks for this ;-)

autogen.sh
```diff
@@ -120,6 +120,13 @@ if grep -q AX_CHECK_COMPILE_FLAG configure; then
   bail_out
 fi
 
+if grep -q PKG_CHECK_MODULES configure; then
+  # The generated configure is invalid because pkg-confg is unavailable.
```
amitdo: >The generated configure is invalid because pkg-confg is unavailable. 

pkg-confg => pkg-config

stweil:
@amitdo, thank you for your very careful review. I fixed the typo in PR #1396.

PR #1399, opened 2018-03-18
viewer: Fix some compiler warnings (-Wstringop-truncation)

stweil:
gcc warnings:

viewer/scrollview.cpp:72:10: warning:
 ‘char* strncpy(char*, const char*, size_t)’ output truncated before
 terminating nul copying as many bytes from a string as its length
 [-Wstringop-truncation]
viewer/scrollview.cpp:118:14: warning:
 ‘char* strncpy(char*, const char*, size_t)’ specified bound depends on
 the length of the source argument [-Wstringop-overflow=]
viewer/scrollview.cpp:746:10: warning:
 ‘char* strncpy(char*, const char*, size_t)’ output truncated before
 terminating nul copying as many bytes from a string as its length
 [-Wstringop-truncation]
viewer/scrollview.cpp:830:10: warning:
 ‘char* strncpy(char*, const char*, size_t)’ output truncated before
 terminating nul copying as many bytes from a string as its length
 [-Wstringop-truncation]

### DIFF
--- a/viewer/scrollview.cpp
+++ b/viewer/scrollview.cpp
@@ -69,8 +69,7 @@ SVEvent* SVEvent::copy() {
   any->command_id = command_id;
   any->counter = counter;
   any->parameter = new char[strlen(parameter) + 1];
-  strncpy(any->parameter, parameter, strlen(parameter));
-  any->parameter[strlen(parameter)] = '\0';
+  strcpy(any->parameter, parameter);
   any->type = type;
   any->x = x;
   any->y = y;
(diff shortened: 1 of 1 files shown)


viewer/scrollview.cpp
```diff
@@ -115,7 +114,7 @@ void* ScrollView::MessageReceiver(void* a) {
 
     if (cur->window != NULL) {
       cur->parameter = new char[strlen(p) + 1];
-      strncpy(cur->parameter, p, strlen(p) + 1);
+      strcpy(cur->parameter, p);
       if (strlen(p) > 0) {  // remove the last \n
         cur->parameter[strlen(p)] = '\0';
```
stweil: That code looks strange as it does not do what the comment claims. It writes a 0 to a byte which is already 0.

PR #1408, opened 2018-03-22
Update version handling

stweil:
ccutil/version.h is now no longer needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,7 +340,6 @@ install(FILES
     ccutil/unicharcompress.h
     ccutil/unicharmap.h
     ccutil/unicharset.h
-    ccutil/version.h
 
     #from lstm/makefile.am
     lstm/convolve.h
--- a/api/apitypes.h
+++ b/api/apitypes.h
@@ -21,7 +21,6 @@
 #define TESSERACT_API_APITYPES_H_
 
 #include "publictypes.h"
-#include "version.h"
 
 // The types used by the API and Page/ResultIterator can be found in:
 //   ccstruct/publictypes.h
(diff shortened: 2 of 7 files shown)


zdenop:
Will this work also for Visual studio builds?

stweil:
[Appveyor](https://ci.appveyor.com/project/zdenop/tesseract/build/master.1389) says yes. I cannot test it myself as I don't have Visual Studio.

amitdo:
So git-rev info is dropped?

stweil:
Yes. `GIT_REV` (and also `vs2010/port/vcsversion.h`) will be removed in a follow-up pull request.

The current code for autoconf in configure.ac already uses the Git information if it is available (not only for the debug case as the old code did). For CMake, this still needs an implementation which first checks Git and then looks in file `VERSION` if Git is unavailable.

stweil:
@zdenop, making a new Tesseract release will require updating and committing the [VERSION file](https://github.com/tesseract-ocr/tesseract/blob/master/VERSION), then adding a Git tag with some descriptive text, then describing the new release in GitHub.

amitdo:
I want this feature:

Have a variable `is_stable`. Its value will be set to `false` during development. Just before tagging a final (non alpha/beta/rc) it should be set to `true`. If `false`,  then the the git revision will be appended to the version info in `PACKAGE_VERSION`.

What do you think about this proposal?

stweil:
I‌ think it is not needed. The git revision is only appended if last tagged revision and current revision differ. It is automatically omitted if you are working on a tagged revision.

amitdo:
Sounds good, thanks!

stweil:
@zdenop, do you have more questions, or can this PR be merged?

zdenop:
I just wantted to test it. It seems to work with Visual Studio 15 2017 Win64

PR #1431, opened 2018-03-26
autoconf: Get version components from PACKAGE_VERSION

stweil:
AX_SPLIT_VERSION only works after AM_INIT_AUTOMAKE, so that macro had
to be moved.

GENERIC_MAJOR_VERSION, GENERIC_MINOR_VERSION and GENERIC_MICRO_VERSION
are now set automatically and can be used in further processing.


stweil:
Running configure now shows the versions:

    [...]
    checking dependency style of g++-8... gcc3
    checking for a sed that does not truncate output... /bin/sed
    checking Major version... 4
    checking Minor version... 0
    checking Point version... 0-beta.1-54-g6b0b
    [...]

PR #1432, opened 2018-03-27
Create version.h from available version information

stweil:


amitdo:
What about installing this header file like the other public headers?

stweil:
Sure, that could be a next step. Tesseract 3.05 and Tesseract 4.0.0-beta.1 use different approaches: while 3.05 (and also 4.00.00alpha) has the version information directly in baseapi.h, 4.0.0-beta.1 has that information in version.h which is included by baseapi.h.

I think that the latest API is not easily manageable: the Debian package installs 58 (!) include files to /usr/include/tesseract and adds that directory to the compiler's search path for include files. How will users know which of those include files should be included directly in their code? How can name conflicts be avoided? `version.h`, but also `network.h` and other names used are very generic, so chances of name conflicts are very high.

PR #1438, opened 2018-03-28
cmake: Automatically build tesseract.rc with version from file

stweil:


stweil:
Is that file needed at all, or could we also remove it? It looks like cmake does not need it.

zdenop:
I will try to test if tesseract.rc is needed today evening or tomorrow...

zdenop:
So rc. file should add additional information to executable (and library?) e.g. about producer, full app name, description, version etc. [It worked for 3.02 but not for this PR](https://ibb.co/iGkbaS). Anyway I will accept it to generate rc file and next we need to find reason, why this details are not visible...

PR #1439, opened 2018-03-28
classify: Fix warnings from Doxygen

stweil:


stweil:
@zdenop, would it be possible to change the way how pull requests are merged? Currently commits get a new date when you merge them, and several commits (like the 5 commits here) are squashed into a single commit. That makes my work a little bit more difficult.

stweil:
Yes, please. I suggest to squash selected pull requests only as a result of reviews and discussions for those pull requests.

Issue #1453, opened 2018-04-08
4.0 bugs on MAC OS X and a step by step for reference


FernandoGOT:
This is  step by step that I used to install tesseract 4.0 on my MAC OS X and the fixes/workaround I needed to do so I could make it work.
I'm sharing this "guide" with the intention of helping other people who may have the same problems I had.

Special thanks for Shree that helped me at the google groups

Project and more details: https://github.com/tesseract-ocr/tesseract

where to get help?

google group: https://groups.google.com/forum/#!forum/tesseract-ocr
git: https://github.com/tesseract-ocr/tesseract/issues

Platform: MAC OS X 10.13.3
Tesseract: 4.0.0-beta.1-69-g10f4
 leptonica-1.75.3
  libjpeg 9c : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11

 Found AVX2
 Found AVX
 Found SSE

# Compiling Tesseract - tesseract 4.0

Reference: https://github.com/tesseract-ocr/tesseract/wiki/Compiling#macos

Warning: Don't install tesseract using brew, since you can't generate the ```ScrollView.jar``` from it! (At least I wasn't able to generate it)

### Steps

1 - Install these libs

```
brew install automake autoconf autoconf-archive libtool
brew install pkgconfig
brew install icu4c
brew install leptonica
brew install gcc
```

2 - Run the code
```
ln -hfs /usr/local/Cellar/icu4c/60.2 /usr/local/opt/icu4c
```

Obs.: ```text2image``` is set to use icu4c/60.2 but the actual version is icu4c/61.1

3 - Clone tesseract repo

```
git clone https://github.com/tesseract-ocr/tesseract/
```

4 - Enter in the folder

```
cd tesseract
```

5 - Run the script

```
./autogen.sh
```

6 - Run the code, and copy the ``` CPPFLAGS ``` and ``` LDFLAGS ```

```
brew info icu4c
```

7 - Update the ``` CPPFLAGS ``` and ``` LDFLAGS ``` and execute the code

```
./configure \
  CPPFLAGS=-I/usr/local/opt/icu4c/include \
  LDFLAGS=-L/usr/local/opt/icu4c/lib
```

8 - Run the code

```
make -j
```

9 - Run the code

```
sudo make install
```

10 - Run the code

```
sudo update_dyld_shared_cache
```

Obs.: this is the ```sudo ldconfig``` version for MAC OS X

11 - Run the code

```
make training
```

[...]

PR #1467, opened 2018-04-10
Fix for mktemp bug on MAC OS X

FernandoGOT:
as reported on #1453 `mktemp --tmpdir` doesn't work for MAC, so you need to use `mktemp -t`

### DIFF
--- a/training/tesstrain_utils.sh
+++ b/training/tesstrain_utils.sh
@@ -192,7 +192,11 @@ parse_flags() {
 
 # Function initializes font config with a unique font cache dir.
 initialize_fontconfig() {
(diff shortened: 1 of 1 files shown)


FernandoGOT:
Please do some testing on linux and windows

amitdo:
BSD distros will have the same issue.

stweil:
So that patch is not sufficient? Could we simply use `mktemp -d` for all platforms? That form is already used for `WORKSPACE_DIR`. Maybe we could also put the `FONT_CONFIG_CACHE` in a subdirectory of `WORKSPACE_DIR`.

amitdo:
https://en.wikipedia.org/wiki/Uname#Examples

PR #1471, opened 2018-04-11
add missing headers to install target

dschoerk:
add missing headers to install target

ccutil.h required for example in the following api call
`GenericVector<STRING> languages;
api->GetAvailableLanguagesAsVector(&languages);`


amitdo:
https://github.com/tesseract-ocr/tesseract/blob/2cc46fa6d456f9b49bc8a70036fa989a9c3baf50/ccutil/Makefile.am

Only headers that are part of `include_HEADERS` should be installed by CMake.

zdenop:
How .gitmodules affect cmake install?

stweil:
It doesn't. At least that part of the pull request is wrong. It also unnecessarily changes the indentation (by using tabs instead of spaces?).

dschoerk:
@amitdo thanks for the quick reply. I first assumed these headers just got lost in translation, but from the makefile I can see they were left out on purpose. 

Offtopic:
I didn't intend to include the change to .gitmodules into the PR. It was a change made for my personal fork to ensure cmake's ExternalProject_Add does not pull the googletest submodule. ExternalProject_Add does not allow to specify no submodules being cloned via GIT_SUBMODULES (Or at least I can't see a way to do it). A separate flag to build or not build the tests in tesseract might be useful, as that is currently only decided by the existance of the "googletest" folder.

PR #1475, opened 2018-04-14
cmake switch to turn off test building

dschoerk:
adds a switch to suppress building googletest and tests

reason: when adding tesseract to another project via [ExternalProject_Add](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html) it is impossible to not clone submodules, which forces tesseracts cmake to build the tests. the added switch retains current default behaviour.


egorpugin:
Thanks!

PR #1479, opened 2018-04-15
training: Add initial support for --version argument and check library version

stweil:


stweil:
`(*argc)` tests whether the number of arguments was not 0. That does not make much sense, because there is at least one argument, namely the name of the executable.

Lines 90 and 91 added help text for the new `--version` argument. Why do you think that it is a mistake?

amitdo:
I missed the ` | `. **My** Mistake :)

I think the way `tesseract` cl present the usage is clearer.

```
Usage:
  tesseract --help | --help-extra | --version
  tesseract --list-langs
  tesseract imagename outputbase [options...] [configfile...]
```


I think we can agree on dropping 
`if (*argc) {`

stweil:
Yes and yes. The help messages for the training tools can be improved (for example they don't document `--help`), and the `if` can be dropped.

There is also a strange feature which I'd like to drop: some training commands support any number of leading `-`, so for example `------help` or `-help` or valid arguments. See the code [here](https://github.com/tesseract-ocr/tesseract/blob/master/training/commandlineflags.cpp#L174).

amitdo:
Sure, you can drop those strange pieces of code. 

The command line parsing that the training tools use looks like incomplete clone of 'gflags'.
https://gflags.github.io/gflags/#special

stweil:
See PR #1488 which addresses the strange pieces of code.

PR #1481, opened 2018-04-16
training: Support new argument --version for remaining executables

stweil:


amitdo:
Maybe you want to use this rule:
https://clang.llvm.org/extra/clang-tidy/checks/readability-else-after-return.html

PR #1484, opened 2018-04-17
Fixed Resource Leak detected by Coverity

noahmetzger:
CID: 1375399 Variable bb1 going out of scope leaks the storage it points to

### DIFF
--- a/ccmain/superscript.cpp
+++ b/ccmain/superscript.cpp
@@ -475,6 +475,7 @@ WERD_RES *Tesseract::TrySuperscriptSplits(
     delete core;
     delete prefix;
     delete suffix;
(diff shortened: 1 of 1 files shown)


zdenop:
and bb0 (after line 490)?

noahmetzger:
i thought of that too but coverity showed for some reason only bb1 as memory leak. So i decided to first fix this one and see how coverity reacts afterwards.

stweil:
I'll run a new scan with Coverity by the end of the week. Therefore I suggest to apply this PR even if it is incomplete. It will be interesting to test whether Coverity scan will complain about `bb0` in the new run.

Issue #1179, opened 2017-10-22
build training tools failed


xinqinglxl:
I meet a error while building training tools from source(4.0 master)
First, I build the tesseract-ocr which is successful.
1 yum install autoconf automake libtool libjpeg-devel libpng-devel libtiff-devel zlib-devel
2 build leptonica from source(leptonica-1.74.4)
3 build tesseract-ocr
   ./autogen.sh
  ./configure --prefix=/usr/local/tesseract --with-extra-libraries=/usr/local/leptonica/lib   
  make
  make install
--------------------------
Then I try to build training tools
1 yum install libicu-devel
2 yum install pango-devel
3 yum install cairo-devel
make training
But it comes the following error

`depbase=`echo boxchar.lo | sed 's|[^/]*$|.deps/&|;s|\.lo$||'`;\
/bin/sh ../libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I..  -O2 -DNDEBUG -DUSE_STD_NAMESPACE -DPANGO_ENABLE_ENGINE -I../ccmain -I../api -I../ccutil -I../ccstruct -I../lstm -I../arch -I../viewer -I../textord -I../dict -I../classify -I../display -I../wordrec -I../cutil  -I/usr/include/pango-1.0 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include   -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng15 -I/usr/include/libdrm    -g -O2 -std=c++11 -MT boxchar.lo -MD -MP -MF $depbase.Tpo -c -o boxchar.lo boxchar.cpp &&\
mv -f $depbase.Tpo $depbase.Plo
[...]

PR #1485, opened 2018-04-17
configure.ac - check the presence of icu 52.1 or higher

amitdo:


amitdo:
https://github.com/tesseract-ocr/tesseract/issues/1179#issuecomment-363057346

PR #1493, opened 2018-04-19
lstmtraining: Fix handling of --max_iterations

stweil:
The iteration counter should be checked for each iteration,
not only at the end of a batch.


amitdo:
I think the current behavior is intentional.

PR #1499, opened 2018-04-20
[WIP] configure.ac: Fix debug mode

amitdo:
Debug mode should be "-g -O0".

See: https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/C_002b_002b-Compiler.html


amitdo:
DON'T MERGE YET.

Needs testing.

amitdo:
It does not work well  :(

Closing for now.

amitdo:
I figured out what was the problem.

I'll submit a fix in a few days.

stweil:
A simple solution which works without a code change is running `./configure [...] CXXFLAGS="-g -O0"`. My usual variant for testing is `mkdir bin/debug && cd bin/debug && ../../configure --enable-debug --disable-shared --disable-static CXXFLAGS="-Wall -Wextra -g -O0"`.

Issue #455, opened 2016-10-28
Replace all NULLs with nullptr


egorpugin:
Since tess is using c++11 we need to `s/NULL/nullptr/`.

amitdo:
http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nullptr.html

stweil:
What about finishing this job? Several thousand `NULL` macros are still in the code. Replacing them could be done per directory. About 293 files would need changes. Here are the file counts per directory:

      7 api
     36 ccmain
     48 ccstruct
     40 ccutil
     40 classify
      6 cutil
      9 dict
      4 doc
     20 lstm
     52 textord
      2 training
      4 viewer
      4 vs2010
     21 wordrec

egorpugin:
It's all about Ray. Such changes can break his merges.

amitdo:
@theraysmith, can you fix it with clang-tidy?
http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nullptr.html

stweil:
@egorpugin, I think this issue can be closed now.

Issue #1500, opened 2018-04-20
[macos] unsupported argument '-q' to option 'Wa,'


jeroen:
Installing master from source on MacOS fails with: __configure: error: Your compiler does not have the necessary c++11 support! Cannot proceed.__ Looking at `config.log` the actual problem is 

```
configure:17144: result: no
configure:17149: checking whether compiler supports C++11
configure:17163: clang++ -c -g -O2 -Wa,-q -std=c++11  conftest.cpp >&5
clang: error: unsupported argument '-q' to option 'Wa,'
```

amitdo:
https://github.com/tesseract-ocr/tesseract/pull/1474
The cl passed.

stweil:
See my comments for PR #1474 which introduced this regression.

stweil:
@ic: The issue is still open after #1516.

ic:
@amitdo @stweil #1522 reverts the commits---sorry as I cannot handle the case more this week.

amitdo:
@zdenop, please close this issue.

jmercouris:
This is still an issue on the latest release of master, with Macports

stweil:
Don't use gcc, then latest code from git should work. And update the Wiki if it gives wrong advice.

ic:
@jmercouris Can you give more details? My comments on the wiki apply to building with `g++`, MacPorts packaging, and the `Makefile`. As far as I know, the main stream is to use `CMake`.

The recent changes I have tried for building with `g++` with MacPorts and the `Makefile` worked at the time of my last changes (commit `4ded0d066e640e60e1640722860fc73d5838befb`).

As this issue is closed and the main stream is on a different toolchain, feel free to get in touch directly ([email?](https://github.com/ic)). Perhaps I can share more details on how I have it working here, and we can strengthen the wiki on this scenario.

jmercouris:
Hi, sorry for the delay in replying! I used @stweil's advice and was able to get it to compile without complaint on MacOS. Apparently issues in my compiled version (4.0) were due to incompatible lang training data (from 3.5) rather than compilation issues.

As per why I posted in this thread, if you attempt to use cc that ships with Xcocde, and you do follow the wiki advice about the flags for g++ from macports, you'll see this error come up. My fault for not reading clearly.

Lastly, I did also attempt to compile with gcc7 from Macports, but it didn't work, and I don't have the log anymore. I hope that all made sense!

ic:
[...]

PR #1507, opened 2018-04-22
Replace NULL by nullptr

stweil:


amitdo:
Fixes #455

stweil:
It also fixes several thousands of warnings from clang (`zero as null pointer constant [-Wzero-as-null-pointer-constant]`). I was just examining issue #1500, saw those warnings and decided to send this pull request.

Issue #1149, opened 2017-09-21
text2image can't produce a complete tif file


ivanzz1001:
### Environment

* **Tesseract Version**: Tesseract4.0.0
* **Commit Number**: 2cc531e
* **Platform**: Linux localhost.localdomain 3.10.0-514.el7.x86_64 #1 SMP Tue Nov 22 16:42:41 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux (Centos7.3)

### Current Behavior:
I use the following command to produce the tif files:
<pre>
# /opt/tesseract4.0/bin/text2image --find_fonts \
--fonts_dir /usr/share/fonts \
--text ./langdata/chi_sim/chi_sim.training_text \
--min_coverage .9  \
--outputbase ./results/chi_sim/chi_sim\
|& grep raw | sed -e 's/ :.*/" \\/g'  | sed -e 's/^/  "/' >./results/chi_sim/fontslist.txt

# ls ./results/chi_sim/
chi_sim.Arial_Unicode_MS.tif               chi_sim.FangSong.tif                 chi_sim.Noto_Sans_SC.tif  chi_sim.STZhongsong.tif
chi_sim.AR_PL_UKai_CN.tif                  chi_sim.KaiTi.tif                    chi_sim.NSimSun.tif       chi_sim.WenQuanYi_Micro_Hei_Mono.tif
chi_sim.AR_PL_UKai_HK.tif                  chi_sim.LiSu.tif                     chi_sim.SimHei.tif        chi_sim.WenQuanYi_Micro_Hei.tif
chi_sim.AR_PL_UKai_TW_MBE.tif              chi_sim.Microsoft_YaHei_Bold.tif     chi_sim.SimSun.tif        chi_sim.WenQuanYi_Zen_Hei_Medium.tif
chi_sim.AR_PL_UKai_TW.tif                  chi_sim.Microsoft_YaHei.tif          chi_sim.STFangsong.tif    chi_sim.WenQuanYi_Zen_Hei_Mono_Medium.tif
chi_sim.AR_PL_UMing_CN_Semi-Light.tif      chi_sim.Noto_Sans_SC_Bold.tif        chi_sim.STKaiti.tif       chi_sim.WenQuanYi_Zen_Hei_Sharp_Medium.tif
[...]

Issue #1508, opened 2018-04-22
only 104 pages used for lstmf instead of the original amount


christophered:
I am using a .txt file containing ~9 million lines, with ~+42 million **unique and un-duplicated** words.
In the "Generating lstmf file" step, it shows `Loaded 104/104 pages (1-104)`
My original txt file size is ~570MB containing +100000 pages, the generated lstmf is 445kb, the **initial** generated traineddata is 1.3MB
**Is Tesseract stopping loading pages after facing `No block overlapping textline`? or is there a size/thresshold limit?**

tesstrain log:
```

=== Starting training for language 'ara'
[Sun Apr 22 08:54:59 PDT 2018] /usr/bin/text2image --fonts_dir=../fonts --font=Arial, --outputbase=/tmp/font_tmp.TfAHElS9Km/sample_text.txt --text=/tmp/font_tmp.TfAHElS9Km/sample_text.txt --fontconfig_tmpdir=/tmp/font_tmp.TfAHElS9Km
Rendered page 0 to file /tmp/font_tmp.TfAHElS9Km/sample_text.txt.tif

=== Phase I: Generating training images ===
Rendering using Arial,
[Sun Apr 22 08:55:01 PDT 2018] /usr/bin/text2image --fontconfig_tmpdir=/tmp/font_tmp.TfAHElS9Km --fonts_dir=../fonts --strip_unrenderable_words --leading=32 --char_spacing=0.0 --exposure=0 --outputbase=/tmp/tmp.FTa4hC8XeN/ara/ara.Arial.exp0 --max_pages=3 --font=Arial, --text=../text/arabic.txt
Rendered page 0 to file /tmp/tmp.FTa4hC8XeN/ara/ara.Arial.exp0.tif
Rendered page 1 to file /tmp/tmp.FTa4hC8XeN/ara/ara.Arial.exp0.tif
Rendered page 2 to file /tmp/tmp.FTa4hC8XeN/ara/ara.Arial.exp0.tif

=== Phase UP: Generating unicharset and unichar properties files ===
[Sun Apr 22 08:56:51 PDT 2018] /usr/bin/unicharset_extractor --output_unicharset /tmp/tmp.FTa4hC8XeN/ara/ara.unicharset --norm_mode 2 /tmp/tmp.FTa4hC8XeN/ara/ara.Arial.exp0.box
Extracting unicharset from box file /tmp/tmp.FTa4hC8XeN/ara/ara.Arial.exp0.box
[...]

PR #1509, opened 2018-04-22
Change max_pages to zero

Shreeshrii:
Fixes https://github.com/tesseract-ocr/tesseract/issues/1149 and https://github.com/tesseract-ocr/tesseract/issues/1508


Shreeshrii:
If need be, this should be made a variable that can be assigned by users when starting tesstrain.sh.

PR #1516, opened 2018-04-23
Fix wrong environment test that breaks clang++ builds.

ic:
g++ builds require extra flags rejected by clang++. The bug is that the
flags are actually added unconditionally. This commit fixes the
condition.

See https://github.com/tesseract-ocr/tesseract/pull/1474 and related.

Tests conducted and report: https://github.com/tesseract-ocr/tesseract/pull/1474#issuecomment-383437228

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -135,8 +135,8 @@ fi
 if $avx -o $avx2 -o $sse41; then
   case "${host_os}" in
     *darwin* | *-macos10*)
-       if test -z "$CLANG"; then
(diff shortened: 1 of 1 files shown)


stweil:
This commit does not fix the problem: on macOS without g++ from MacPorts there exists a g++ frontend which calls Apple's clang. And of course that g++ fails with the new compiler flags.

ic:
I see, sorry about that. For now, rolled back, and I put a workaround note in the wiki.

PR #1518, opened 2018-04-23
Remove 'AUTOMAKE_OPTIONS' from all Makefile.am files

amitdo:
'AM_INIT_AUTOMAKE()' in configure.ac sets options globally.


stweil:
Appveyor CI (MS compiler) currently fails because of one of my commits. I am just preparing a fix.

PR #1522, opened 2018-04-23
Revert failed attempt to support MacPort's g++

ic:
The support will require more work, and postpone for now.

This contributes to #1500, and removes #1474 and #1516.


ic:
If you are looking how to build with MacPort's `g++`, please refer to [this comment](https://github.com/tesseract-ocr/tesseract/pull/1474#issuecomment-383437744).

PR #1526, opened 2018-04-25
Fix line endings

stweil:
Replace DOS line endings (CRLF) by standard (LF only).


egorpugin:
I tried to setup git to always commit LF about 2-3 weeks ago.
Seems it's not working. :(
Sorry.
Maybe we could revert and force push with proper endings, so git blame shows original authors?

stweil:
`git blame -w` ignores that kind of modifications, so that's not a big problem. Reverting an official repository would not be good, because it could cause trouble for anybody who has fetched the intermediate version.

egorpugin:
Ok.
Thank you for fixes.
How do you autocorrect those files? Using git? Or with custom script?

stweil:
Replace CRLF by LF:

    perl -pi -e 's/[[:space:]]+$/\n/' $(find * -type f|xargs file|grep CR|sed s/:.*//)

Fix file endings: custom script

    #!/bin/sh

    for file in $@; do
      if test -s $file && (file $file | grep -qw text); then
        # File is not empty and a text file.
        lc=$(tail -1c $file | wc -l)
        if test "$lc" = "0"; then
          # File does not end with newline.
          echo "Add missing newline to $file ..."
          echo >>$file
        fi
        lc=$(tail -1 $file | wc -c)
        if test "$lc" = "1"; then
          echo "Remove empty newline from $file ..."
          head --lines=-1 $file >/tmp/nonewline && mv /tmp/nonewline $file
        fi
      fi
    done

Both could be combined of course.

PR #1534, opened 2018-04-27
Add missing override specifiers

UniQP:


amitdo:
>virtual StaticShape OutputShape(const StaticShape& input_shape) const override;
 
virtual is redundant here.

UniQP:
@amitdo I adapted the commit accordingly.

PR #1539, opened 2018-04-30
Remove old hack for Visual Studio

stweil:
It should not be needed with newer versions of Visual Studio.


amitdo:
Reference:
https://msdn.microsoft.com/en-us/library/hh567368.aspx

Issue #594, opened 2016-12-19
RFC: Reorganize source tree


egorpugin:
I'd like to propose changes to tesseract source tree structure.
Today the common way is to have `src` folder with all program stuff and `include` folder with public headers. Now we have a lot of dirs in the root - that's very annoying.
On the first stage I propose:
1. move all sources into `src`
2. move training tools from `training` to `tools/training`

Later we can try to move public headers to `include` directory.

The new look will be like:
![pic](http://i.imgur.com/K8NkI8T.png)

If there are no objections, I'll commit changes.

jbreiden:
Make sure to get positive confirmation from Ray before doing something like this. He's considering a different, very disruptive change involving the non-LSTM recognizer. It would collide very badly with source reorganization.

egorpugin:
Yea, I'll act only in case of positive reply from Ray.

stweil:
Related: I suggest to remove some large binaries (mainly in `testdata` and `testing`) which are used for tests and add them to a new repository `[test](https://github.com/tesseract-ocr/test)`. Like `googletest`, this new repository can then be added as a submodule.

stweil:
cc: @Shreeshrii

Shreeshrii:
>I suggest to remove some large binaries (mainly in testdata and testing) which are used for tests and add them to a new repository test. Like googletest, this new repository can then be added as a submodule.

Yes. I think it is a good idea. It would also be useful to streamline the various other test related folders at the same time.

Shreeshrii:
However this may require @theraysmith / @jbreiden to create the new repo.

zdenop:
done. enjoy ;-)

amitdo:
https://github.com/tesseract-ocr/test

theraysmith:
How about the following first stage:
Move all the source to src.
Create pointer headers in include that #include the real header for the public headers.
Try to avoid creating any APIs that make use of the old classifier.

stweil:
@egorpugin, will you send a pull request for the first step, moving all source to src?

egorpugin:
Sure.

egorpugin:
Changes are in master. Please, check build scripts. I did small updates, cmake & cppan seems ok, but autotools is untested.

Shreeshrii:
The training folder should also go under src.

Thanks.

egorpugin:
Done.

stweil:
[...]

PR #1549, opened 2018-05-02
RFC: Add examples, Move testing files to test

Shreeshrii:
I have added a folder for examples and added a couple of examples there.

As a precursor to moving files to test repo, I have moved unittest related folders under a new folder called test.

I have left the `testing` folder as is, since it has phototest.tif and eurotext.tif which have been around for many versions. The newer unittest files have been moved.

@zdenop Please review/test before merging. Thanks.


Shreeshrii:
>test should be removed from tesseract-ocr/tesseract to tesseract-ocr/test/

Yes, that is the plan. 

I wanted to get feedback regarding 'testing' folder, if that also needs to be moved to the separate test repo.

Also, my understanding is that it will be added back here as a submodule, which you may have to do.

Shreeshrii:
Looking again at comment by @stweil at https://github.com/tesseract-ocr/tesseract/issues/594#issuecomment-383833647 
I think he meant only the image and other test related files, not the test sources.

I will make the appropriate changes.

Shreeshrii:
Leaving this open for the time being for  further feedback

egorpugin:
My vote is for having test dir inside tess, not outside.
It's a common approach.
Sometimes you cannot find examples of API usage in `examples` dir and you go straight into `test` dir to check tests or unittests.

Shreeshrii:
@egorpugin I think the original  intention was to only move the large binaries (image files) to a external repo. 

So, it should be just the testdata folder that is be moved outside.

@zdenop may need to rename the new repo as testdata instead of test.

zdenop:
@Shreeshrii : you are not moving files to other repo! you are moving it to other directory.

Shreeshrii:
@zdenop 

This is still marked as RFC.

We need to rename the new test repo as testdata. 
I can then move all the testdata files there. 
We will need to then include testdata repo as submodule in test/testdata.

This way test source files will be under

tesseract/test/tests
tesseract/test/unittest

testdata files will be included as submodule in

tesseract/test/testdata


testing files (phototest.tif and eurotext.tif) + unlv related files will remain in 

tesseract/testing


If everyone is in agreement with that, then we can go ahead.

stweil:
I had suggested to name the new repository `test` because `testdata` is very similar to `tessdata` which might be confusing. In addition to binary files which are used for tests, we could also move the UNLV tests there. All other test code should remain in the `tesseract` repository.

Shreeshrii:
>I had suggested to name the new repository test because testdata is very similar to tessdata which might be confusing.

I agree, the names tessdata and testdata are confusing.

However 'test' seems to be used by other projects (from the little that I have seen)  for test sources also.

How about calling the new repo as `testing_data` ?

egorpugin:
> I think the original intention was to only move the large binaries (image files) to a external repo.

Also true. Having heavy files inside main is bad.

PR #1551, opened 2018-05-03
Fix Tesseract for big endian machines

stweil:


stweil:
@AlexanderP, @jbreiden, the 2nd commit fixes Tesseract for big endian ppc64 (and I expect also for other big endian machines).

stweil:
Thank you to the GCC Compile Farm Project who helps by giving me access to a big endian ppc64 test machine.

PR #1555, opened 2018-05-03
Don't disable compiler warnings for Visual Studio

stweil:
It's still possible to set the warning level in the project settings,
but single source files should normally not disable compiler warnings.


amitdo:
>single source files should normally not disable compiler warnings.

I agree.

PR #1559, opened 2018-05-04
Remove autoconf-archive dependency

stweil:
It creates much confusion and causes many issue reports,
so let us drop this dependency.

The two new files in the m4/ directory are current copies from GitHub
(https://github.com/autoconf-archive/autoconf-archive/).

### DIFF
--- a/INSTALL.GIT.md
+++ b/INSTALL.GIT.md
@@ -10,7 +10,6 @@ You need Leptonica 1.74.2 (minimum) for Tesseract 4.0x.
 
 Known dependencies for training tools (excluding leptonica):
  * compiler with c++11 support
- * autoconf-archive
  * automake
  * pkg-config
  * pango-devel
--- a/autogen.sh
+++ b/autogen.sh
@@ -113,13 +113,6 @@ automake --add-missing --copy --warnings=all || bail_out
 echo "Running autoconf"
 autoconf || bail_out
 
-if grep -q AX_CHECK_COMPILE_FLAG configure; then
-  # The generated configure is invalid because autoconf-archive is unavailable.
-  rm configure
-  echo "Missing autoconf-archive. Check the build requirements."
-  bail_out
-fi
-
 if grep -q PKG_CHECK_MODULES configure; then
   # The generated configure is invalid because pkg-config is unavailable.
   rm configure
--- a/m4/ax_check_compile_flag.m4
+++ b/m4/ax_check_compile_flag.m4
@@ -1,0 +1,53 @@
+# ===========================================================================
+#  https://www.gnu.org/software/autoconf-archive/ax_check_compile_flag.html
+# ===========================================================================
+#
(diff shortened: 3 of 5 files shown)


amitdo:
There is one problem here: **GPL**.

stweil:
That's needed for `AX_CHECK_COMPILE_FLAG`.

I'm not sure whether that is a problem as long as it is not part of the Tesseract releases.

The commits 2444671ea9eec1fa5dd80f5bf75931552bee3aba and b8e3eeeb0199df6236dbef6c2cb069bcbe48ba55 already added GPL'ed files to `m4` until that directory was removed in commit 4085f1b5e92d80b1ea84180eda9582eb53a59e4c.

@zdenop, what do you think?

stweil:
Maybe we can ask @gdraheim and @mkbosmans (the copyright holders) directly about the license issue. May we add your autoconf code (see this pull request) to Tesseract?

zdenop:
1. Downloading extra files from internet as replacement of common "autoconf-archive" is bad idea.
2. Relicence tesseract code is no goal.
3. Building sw from source requires knowledge. We should stop to provide free lessons and focus on real tesseract issues.

So real options is to:
- ask original author for permision us their code under Apache licence
- create own test for that compiler features...

amitdo:
https://web.archive.org/web/20120101192501/https://code.google.com/p/tesseract-ocr
>## Important License Note
>
>The code is all licensed with the Apache 2.0 License **EXCEPT** the tesseractTrainer.py, which is licensed with GPL. 

>Maybe we can ask @gdraheim and @mkbosmans (the copyright holders) directly about the license issue. May we add your autoconf code (see this pull request) to Tesseract?

You don't need their permission to this.

>Or can you relicense the code under the all-permissive license as suggested by the FSF?

Apache 2.0 would be the best option for us. second one is https://opensource.org/licenses/MIT.

gdraheim:
all my ax-macros should have been under a more liberal license, but the autoconf-archive default did hit it at some point. All-permissive seems fine.

amitdo:
https://github.com/autoconf-archive/autoconf-archive/blame/master/m4/ax_check_compile_flag.m4

mkbosmans:
Sure, I don't need strong copyleft on these macros. Any of the BSD/MIT/Apache/MPL is fine.

gdraheim:
I had never intended to put it under a restrictive license - that had been the decision of the maintainer of the macro archive when picking up older stuff. Having the license like that of autoconf did seem fair enough. After all, all macros could be "relicensed" but that's even more of cummunication.

zdenop:
So how to proceed? 
Should we change licence info in this commit with reference to this conversation and merge it? Or???

stweil:
Let us wait for [this pull request](https://github.com/autoconf-archive/autoconf-archive/pull/160) which I have just written to be accepted.

PR #1585, opened 2018-05-19
Small enhancements (adding const, etc.)

ZaMaZaN4iK:
Hello.

I started work on different refactorings inside tessaract (fixing warnings, adding missing const keyword, etc.).

It's just initial commit. Please give me any feedback about this. Is it good for Tessaract? Can I continue my work on refactoring?

### DIFF
--- a/src/classify/adaptive.cpp
+++ b/src/classify/adaptive.cpp
@@ -25,9 +25,9 @@
 #include "classify.h"
 
 #ifdef __UNIX__
-#include <assert.h>
+#include <cassert>
 #endif
-#include <stdio.h>
+#include <cstdio>
 
 using tesseract::TFile;
 
@@ -110,7 +110,6 @@ void FreePermConfig(PERM_CONFIG Config) {
  */
 ADAPT_CLASS NewAdaptedClass() {
   ADAPT_CLASS Class;
-  int i;
 
   Class = (ADAPT_CLASS) Emalloc (sizeof (ADAPT_CLASS_STRUCT));
   Class->NumPermConfigs = 0;
@@ -122,7 +121,7 @@ ADAPT_CLASS NewAdaptedClass() {
   zero_all_bits (Class->PermProtos, WordsInVectorOfSize (MAX_NUM_PROTOS));
   zero_all_bits (Class->PermConfigs, WordsInVectorOfSize (MAX_NUM_CONFIGS));
 
-  for (i = 0; i < MAX_NUM_CONFIGS; i++)
+  for (int i = 0; i < MAX_NUM_CONFIGS; i++)
     TempConfigFor (Class, i) = nullptr;
 
   return (Class);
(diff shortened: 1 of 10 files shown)


ZaMaZaN4iK:
E.g. some warnings about conditional simplifications.

ZaMaZaN4iK:
@stweil Is it important to fix just compiler warnings? Or I can refactore code, use somewhere modern C++11 features, etc.?

stweil:
A good general rule for commits and pull requests is to separate things which are not related. That makes it easier for reviewers. Your first commit was a mix, while your follow up commits cleanly explained what was changed and addressed single issues.

When fixing compiler warnings, I'd expect to know which warning for which compiler (gcc or MS C?) was fixed.

ZaMaZaN4iK:
>When fixing compiler warnings, I'd expect to know which warning for which compiler (gcc or MS C?) was fixed.

Agree. But warnings can be not only from compilers. I got different warnings from clang-tidy, cppcheck. But ok - next time I'll describe what were fixed by my changes with better description.

>For include statements, I personally prefer to have a well defined order of these statements. Usually I start with the system header files (#include <...>) which I sort alphabetically before the local header files (#include "...").

I also prefer this rule. I think I can fix it by another PR.

egorpugin:
> For include statements ... Usually I start with the system header files

Wrong. Correct order - http://llvm.org/docs/CodingStandards.html#include-style
1. this header
2. this project headers
3. your deps headers
4. system headers

Split groups with empty line and sort inside them.

ZaMaZaN4iK:
@egorpugin Does Tesseract has any guideline for source code? Can't find any information about it.

stweil:
As often with coding conventions there is no strict wrong or right. See [here](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) another style guide.

ZaMaZaN4iK:
@stweil @egorpugin What else should be done with this PR before merging?

amitdo:
You can use this for llvm include order style:
https://clang.llvm.org/extra/clang-tidy/checks/llvm-include-order.html

stweil:
@ZaMaZaN4iK, if you use tools to refactor the code, those tools and their parameters should also be documented in your commit.

As long as you continue to add more commits to pull request #1585, I suggest to rename it to "RFC: Refactor code" or something like that to make it clear that you want to discuss those changes.

For simple isolated changes like removing `register` I suggest to make new individual pull requests which can be applied without much discussion.

stweil:
Smaller pull requests would also make it easier to see which changes have broken the Windows build.

ZaMaZaN4iK:
@stweil Ok, understood. Already fixed an issue with MIN/MAX on Windows platform.

>For simple isolated changes like removing register I suggest to make new individual pull requests which can be applied without much discussion.

Ok. As for now I suggest left my changes about 'register' in this PR, but possible next small and independent improvements I will make as new PR.

amitdo:
Since most of the code comes from Google devs, we should use the Googlw style guide.

amitdo:
https://github.com/tesseract-ocr/tesseract/commit/a80a8f17bb32be8bdd5124057219620b711491a7#commitcomment-29137025

PR #1588, opened 2018-05-20
Use standard bool instead of BOOL8.

ZaMaZaN4iK:
Hello.

Here is refactoring about using bool instead of BOOL8.

### DIFF
--- a/src/ccmain/adaptions.cpp
+++ b/src/ccmain/adaptions.cpp
@@ -95,7 +95,7 @@ BOOL8 Tesseract::word_adaptable(  //should we adapt?
     return FALSE;
   }
 
-  if (flags.bit (CHECK_ONE_ELL_CONFLICT) && one_ell_conflict (word, FALSE)) {
+  if (flags.bit (CHECK_ONE_ELL_CONFLICT) && one_ell_conflict (word, false)) {
     if (tessedit_adaption_debug) tprintf("word has ell conflict\n");
     return FALSE;
   }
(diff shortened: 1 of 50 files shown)


stweil:
`BOOL8` is always a single byte. `bool` is typically also a single byte, but that depends on the compiler. So in addition to replacing one by the other, we must also take care when such data is read from or written to a file, for example.

ZaMaZaN4iK:
I didn't fix places, where BOOL8 is used for writing to file or something similar. I fixed a lot of places, where BOOL8 were used as typical bool. Also in modern sources bool is used. So I don't see a reason to use BOOL8 instead of standard bool (except cases, which you described above. In C++17 for this was added std::byte)

egorpugin:
@stweil @zdenop Any objections, more comments?

stweil:
I appreciate replacing proprietary data types like BOOL8 by standard data types like bool, but could not review all changes because of the size of this pull request.

I noticed lots of unrelated changes, most of them changing code formatting. Code formatting should be addressed separately. @egorpugin or @zenop, do you know how Ray did the code formatting clean-ups? Which tool / parameters did he use? Should Tesseract use `pointer *val` or `pointer* val`, for example?

egorpugin:
Formatting can be easily handled with provided `.clang-format`. We could write one for tess.
It solves all questions.

ZaMaZaN4iK:
@stweil Sorry about code formatting changes - my refactoring tool automatically reformat refactored code. We can fix it easily by providing script for fixing code-style (clang-format, uncrustify, etc. I suggest to use clang-format).

ZaMaZaN4iK:
@Shreeshrii Seems reasonable. but FMPOV TRUE/FALSE -> true/false and BOOL8 -> bool changes are related. But ok - it can be divided :-)

amitdo:
https://www.google.com/search?q=%22.clang-format%22+%22github.com%2Fgoogle%22

PR #1591, opened 2018-05-21
Use default keyword instead of empty ctors/dtors.

ZaMaZaN4iK:
In C++11 was introduced default keyword. It's useful for declaring empty ctors/dtors. In this PR I have replaced empty ctors/dtors by default keyword.

### DIFF
--- a/src/ccmain/ltrresultiterator.cpp
+++ b/src/ccmain/ltrresultiterator.cpp
@@ -37,9 +37,6 @@ LTRResultIterator::LTRResultIterator(PAGE_RES* page_res, Tesseract* tesseract,
     paragraph_separator_("\n") {
 }
 
-LTRResultIterator::~LTRResultIterator() {
-}
-
 // Returns the null terminated UTF-8 encoded text string for the current
 // object at the given level. Use delete [] to free after use.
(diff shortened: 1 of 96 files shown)


egorpugin:
@stweil @zdenop Any objections, comments?

stweil:
Commit message: 
s/intoruced/introduced/
s/dtros/dtors/

@ZaMaZaN4iK, is it still possible to set a breakpoint on the default constructor in the debugger?

ZaMaZaN4iK:
@stweil No, it is not. Because there is no place, where you can set breakpoint. I think it's not so useful with empty ctors/dtors, but if you want, for debugging you can replace default ctor/dtor with empty version, finish debug and revert changes.

egorpugin:
> is it still possible to set a breakpoint on the default constructor in the debugger?

It's possible - 
image

ZaMaZaN4iK:
@egorpugin Hm, didn't know that Qt Creator can do it. Thank you!

stweil:
s/emty/empty/ in the commit message, but that's not a real objection. As breakpoints work, the code modifications are fine for me.

PR #1595, opened 2018-05-22
Use std::unique_ptr instead of manual memory management.

ZaMaZaN4iK:
Here I started work on refactoring old C-style memory management. In simple cases new/delete can be replaced by std::unique_ptr.


ZaMaZaN4iK:
Fixed conflict.

Issue #854, opened 2017-04-30
Tesseract coding style


stweil:
From time to time Ray cleans the code by using `clang-tidy`. It would be nice to document that process to allow other developers to check their code contributions and avoid style problems from the beginning.

The current tidy process does not cover some significant code style issues. Most noticeable: it does not change from `( xxx )` (about 480 lines of code) to `(xxx)` (which is the preferred [Google style](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions), more than 50000 lines of code).

egorpugin:
For code style clang-format could be applied.

zdenop:
@stweil : anything else should be done here?

stweil:
Pull request #1935 fixes style issues for some more code. I think for 4.0.0 this is sufficient.

zdenop:
thanks.

PR #1598, opened 2018-05-23
Use clang-format for code formatting

ZaMaZaN4iK:
For fixing I used clang-format -i -style=Google **/*.h && clang-format -i -style=Google **/*.cpp

clang-format 6.0.0

In this PR I suggest to discuss about code-style for Tesseract. Because FMPOV Google style isn't ideal for Tesseract and we can add some Tesseract-specific options.


amitdo:
https://github.com/tesseract-ocr/tesseract/pull/848#issuecomment-298099240
#854

stweil:
Why is https://github.com/stweil/tesseract/tree/format different?

Shreeshrii:
https://github.com/tesseract-ocr/tesseract/pull/1598 is using clang-format 6.0.0

stweil:
I now created https://github.com/stweil/tesseract/tree/clang-format-6 with clang-format 6.0.0-3. The result is identical to that from clang-format 7.0.0 (that means not the same as https://github.com/ZaMaZaN4iK/tesseract/tree/fix_format). It needs three iterations to stabilize. The formatting was done on Debian Buster.

stweil:
https://github.com/stweil/tesseract/tree/clang-format-4 and https://github.com/stweil/tesseract/tree/clang-format-5 give different results.

amitdo:
Run with clang 5 and 6/7:

`clang-format -style=google -dump-config > .clang-format`

Do you get the same result?

amitdo:
appveyor failed.
It doesn't like the openmp formatting in lstm.cpp.

stweil:
Appveyor fails in my four tests, too. https://github.com/stweil/tesseract/commit/4ee0fdccbd36bef919051b4c23f4c068c11b7df5 shows the dumped configurations.

amitdo:
>I suggest to use the same variant everywhere.

:+1: 

`type *variable` vs. `type* variable`

>Personally I  prefer the first one, but using only the second one would also be fine.  What do you think?

It seems that Ray prefers `type* variable` (it's also my favorite style).
https://github.com/tesseract-ocr/tesseract/commit/c1c1e426b32794d5e84134ee81bf895ff0228fe5#diff-f5c88d7e2ff4fc9424ae517ce7b4e1a6

ZaMaZaN4iK:
@amitdo I also prefer `type* variable`

ZaMaZaN4iK:
I think we should dump existing one format profile (suggest use Google code style), correct it a little bit and push to Tesseract repo

stweil:
https://github.com/stweil/tesseract/tree/format now uses uniform left aligned pointers and also includes a `.clang-format` file. Stil missing: build fix for Appveyor.

zdenop:
If yes, than we should do it before 4.0 release.

ekene90:
it is good

kingmoduk:
were are you

zdenop:
@ZaMaZaN4iK , @stweil : Can be this PR fixed / replaced / closed this week?

stweil:
Yes, it can be closed. I think we won't reformat the whole code for 4.0.0.

amitdo:
I suggest to re-format the whole codebase for 5.0.0.

amitdo:
First step - reformat all headers in 'include/tesseract'.

PR #1605, opened 2018-05-25
Fix different warns from clang-tidy.

ZaMaZaN4iK:
Most of warns are about virtual ctors/dtors/functions, deprecated headers, default ctors/dtors, range-based for, nullptr use, using empty() instead of size() == 0 in GenericVector.


stweil:
@ZaMaZaN4iK, it would be good to have at least some of the issues which you had addressed in your pull request here fixed in the Tesseract code base. Are you planning new updated pull requests?

zdenop:
IMO replacing C headers with C++ headers  (e.g time.h with ctime) could be done easily with search and replace ;-)

PR #1607, opened 2018-05-26
Add configuration for clang-format-6.0 and clang-format 7

stweil:
It can be used to format the code according to the Google C++ Style Guide (https://google.github.io/styleguide/cppguide.html).


stweil:
I updated the configuration. It is now much shorter and gives identical results with clang-format-6.0 and clang-format-7.

PR #1614, opened 2018-05-29
Recognition progress in C API

j-kubik:
I need to use progress reporting while using C API. I have written bindings for the functionality.

### DIFF
--- a/unittest/progress_test.cc
+++ b/unittest/progress_test.cc
@@ -1,0 +1,161 @@
+///////////////////////////////////////////////////////////////////////
+// File:        apiexample_test.cc
+// Description: Progress reporting API Test for Tesseract.
+// Author:      Jaroslaw Kubik
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////
+
+// expects clone of tessdata_fast repo in ../../tessdata_fast
+
+#include "include_gunit.h"
+#include "gmock/gmock.h"
+#include "baseapi.h"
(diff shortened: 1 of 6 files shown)


zdenop:
AppVeyor faild with error:
[00:05:45]   c:\projects\tesseract\src\ccutil\ocrclass.h(179): error C4716: 'ETEXT_DESC::default_progress_func': must return a value

zdenop:
Thanks. Do you have a test case for it?

j-kubik:
@zdenop I didn't find any test for the progress reporting at all so I didn't bother. I could add some test cases if required.

zdenop:
1. You are modifying API, so it would be great to see why do you propose change.
2. There is not much contributors for C-API 
3. Generally: it would be great to have test suit for testing tesseract commit quality (e.g. if future commits break some tesseract usage, it should be identified ASAP)

j-kubik:
@stweil The typo has been corrected.

j-kubik:
@zdenop 
Ad 1. The change to the API was proposed (as described in commit message) because the progress reporting callback has no 'this' parameter, and is therefore forced to use global variables if some additional data/objects need to be accessed. This is visible even in the test case I created. I had to create a static `instance` pointer in order to work-around that limitation.
Ad 2. I need the C-API so I wrote functions I needed. Just thought it could be checked into the main repo in case anybody else wanted to use these as well.
Ad 3. I added tests for the progress reporting API.

zdenop:
thanks

unittest/progress_test.cc
```diff
@@ -0,0 +1,161 @@
...
+    ETEXT_DESC monitor;
+
+    ClassicMockProgressSink()
+    {
+        monitor.progress_callback = []( int progress, int, int, int, int ) ->bool {
+          return instance->classicProgress( progress );
+        };
+        monitor.cancel = []( void* ths, int words ) -> bool {
+          ((ClassicMockProgressSink*)ths)->cancel( words );
+        };
```
stweil: Here the compiler issues a warning:

    ../../../../unittest/progress_test.cc: In Lambda-Funktion:
    ../../../../unittest/progress_test.cc:59:9: Warnung: keine return-Anweisung in nicht void zurückgebender Funktion [-Wreturn-type]
             };
             ^

It is also reported by Coverity Scan ("CID 1393240 Missing return statement").

zdenop:
@j-kubik : Can you have a look at #1622 ?

wpapp:
it's ok now,thank u.
@stweil

Issue #1636, opened 2018-06-05
LoadDataFromFile should reserve an extra byte for '\0'


paulk124:
### Environment

* **Tesseract Version**: 
3.05
* **Commit Number**: 
65e5b7a4a65cdf6755503f07ce26d871aa2e2f01
* **Platform**: 
Windows 64-bit, Visual Studio 2015 Update 3

### Current Behavior:

Here is a sample of the problem it causes. I run the following to train the attached image and box file:

`tesseract gdt.symbols.exp0.tif gdt.symbols.exp0 box.train
`

And here is the output:

```
Tesseract Open Source OCR Engine v3.05.01 with Leptonica

Page 1
Bad box coordinates in boxfile string! ²²²²▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌╦ÇÆ§≡¿←
APPLY_BOXES:
   Boxes read from boxfile:       7
   Found 7 good blobs.
Generated training data for 3 words

```
The message about the bad box coordinates is caused because function ReadMemBoxes() reads memory past the end of the const char* box_data parameter.

### Expected Behavior:

Here is the expected output from running the tesseract command:

```
Tesseract Open Source OCR Engine v3.05.01 with Leptonica
Page 1
APPLY_BOXES:
   Boxes read from boxfile:       7
   Found 7 good blobs.
Generated training data for 3 words

```
### Suggested Fix:

To fix this, change function LoadDataFromFile() to always reserve an extra byte so the caller can add a ‘\0’ if they want. Then in ReadAllBoxes(), append ‘\0’ to the vector after calling LoadDataFromFile(). Here are the fixed functions:

```
inline bool LoadDataFromFile(const STRING& filename,
                             GenericVector<char>* data) {
  bool result = false;
  FILE* fp = fopen(filename.string(), "rb");
  if (fp != NULL) {
    fseek(fp, 0, SEEK_END);
    size_t size = ftell(fp);
    fseek(fp, 0, SEEK_SET);
    if (size > 0) {
      // reserve an extra byte in case caller wants to append a '\0' character
      data->reserve(size + 1);
      data->resize_no_init(size);
      result = fread(&(*data)[0], 1, size, fp) == size;
    }
    fclose(fp);
  }
  return result;
}
```

```
bool ReadAllBoxes(int target_page, bool skip_blanks, const STRING& filename,
                  GenericVector<TBOX>* boxes,
                  GenericVector<STRING>* texts,
[...]

PR #1637, opened 2018-06-05
Reserve extra byte in LoadDataFromFile() in case caller wants to appe…

paulk124:
Here is the related issue:
https://github.com/tesseract-ocr/tesseract/issues/1636


zdenop:
Fixes #1636

stweil:
I don't think this change was needed. It allocates an additional byte also when it is not needed. The problem was already fixed with my earlier pull request.

paulk124:
@stweil, I agree the bug was already fixed with your earlier pull request. The only issue was the costly reallocation needed when the caller wants to append a '\0' character. Reserving an extra byte in LoadDataFromFile() doesn't affect any other caller because the size of the GenericVector will still be the same as it was before. The only slight downside is that all callers will have an extra byte allocated for the vector. This shouldn't impact performance and uses a very small amount of extra memory. I think the trade-off is well worth avoiding the memory reallocation.

PR #1642, opened 2018-06-07
Remove some files which are generated by the UNLV test

stweil:
They don't contain useful information.


stweil:
@Shreeshrii, please review. Which report files should we add for current and future releases of Tesseract? Would  <RELEASE>.summary be sufficient?

Shreeshrii:
I had added a gitignore in the unlvtests directory.

The reports directory should be under it, if they are under unittest, then
it must be accidental.

The makefile had a list of reports included in original runs for
comparison. We can save the same level of detail, summary for each type of
document and total.

(Replying on phone without checking the file structure).

stweil:
Thank you. I modified the pull request. It now only removes the unneeded report files.

Shreeshrii:
The unlvtests are useful from historical point of view for comparing to old results. I wanted to make sure that the testsuite was functional. From that perspective keeping the old format of results is ok.

It's current setup is limited to English only. The Spanish dataset was available but not tested.

Going forward, we can choose the summary format and give results based on Tess version, traineddata file used and language.

Nick White's version of ocr-evaluation-tools also has a script for testing accuracy of a set of files for a given traineddata. We can give it a try.

Shreeshrii:
https://github.com/Shreeshrii/ocr-evaluation-tools/blob/master/tessaccsummary

Shreeshrii:
@stweil 
tess4.0.0-beta.1.bus.3B.sum	move testing and testdata to test, add unlvtests	2 days ago
tess4.0.0-beta.1.summary	move testing and testdata to test, add unlvtests	2 days ago
tess4.0.0-beta.1.total.sum	move testing and testdata to test, add unlvtests	2 days ago

These files are also from the test run and should be deleted.

Shreeshrii:
>Would RELEASE.summary be sufficient?

@stweil

Looked at the existing files just now  - 2.03 and 2.04 summary. Yes, RELEASE.summary would be sufficient, since they already include the detail regarding the different document types.

The other files were committed accidentally from the aborted test runs. Thanks for removing them.

Shreeshrii:
>I had added a gitignore in the unlvtests directory.

I see now that the gitignore in tesseract root dir has the lists for the various subdirs also. We should do the same for unlvtests.

PR #1644, opened 2018-06-08
add /src/ to api, training and vs2010 in gitignore

Shreeshrii:


zdenop:
vs2010 part should be removed from .gitignore - it is obsolete and current master can not be build with VS2010.

Shreeshrii:
OK, Thanks.

There are still some files in src/vs2010

Are they still required?

ShreeDevi

stweil:
> There are still some files in src/vs2010. Are they still required?

Yes, they are also used by newer versions of Visual Studio. Maybe we should rename the directory from `vs2010` to just `vs` (and update related files).

PR #1649, opened 2018-06-08
Test for correct locale settings

stweil:
Normal C++ programs like those which are built for tesseract automatically
set the locale "C".

There can be different locale settings if the tesseract library is used
in other software.

A wrong locale can cause wrong results from sscanf which is used at
different places in the tesseract code, so make sure that we have the
right locale settings and fail if that is not the case.

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -43,6 +43,7 @@
 #endif  // _WIN32
 
 #include <algorithm>
+#include <clocale>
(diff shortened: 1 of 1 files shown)


amitdo:
@stweil,
 Can you add a short text about this PR to Release Notes?

stweil:
Well, it's now a little bit more than a short text. I hope that's fine.

PR #1678, opened 2018-06-18
Require tesseract/ for API header files (fixes potential name conflicts)

stweil:
The tesseract/ subdirectory is no longer automatically added to the
include path of the compiler. Therefore old code which used code like

    #include "capi.h"

must now change that to

    #include "tesseract/capi.h"

This avoids name conflicts with header files from other projects.


egorpugin:
They'll also change for 4.0.0 release.

PR #1681, opened 2018-06-18
minor changes to unittests

Shreeshrii:
* Also added additional tests for tessdata_best and tessdata in apiexample.
* Split testcase in apiexample for better readability of code and results log.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,6 +103,8 @@ kernel*.bin
 /unittest/matrix_test
 /unittest/osd_test
 /unittest/tesseracttests
+/unittest/progress_test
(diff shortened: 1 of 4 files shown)


stweil:
`#include "tesseract/baseapi.h"` requires that the header files were installed.

Shreeshrii:
@stweil is tesseract/ not needed here?

stweil:
It is only needed for external code which uses the Tesseract library and the Tesseract API header files, and only if that code uses the compiler flags which are suggested by pkg-config.

Shreeshrii:
@stweil Thanks. Have removed them now.

PR #1682, opened 2018-06-19
Do not merge - langtests draft for review

Shreeshrii:
@stweil Here is a language test setup for Fraktur with results for the /manual transcription. I will rerun later with corrected transcriptions.

The bash scripts are modified versions of the same as in unlvtests. I have left the names to be same, is it ok?


Shreeshrii:
Is https://raw.githubusercontent.com/stopwords-iso/stopwords-de/master/stopwords-de.txt ok as a stopwords file for use with frk?

The sed command to change line endings to space did not work. I did it in notepad++.

Shreeshrii:
Closing. Will redo after using new transcriptions.

PR #1694, opened 2018-06-21
Clean usage of assert.h

stweil:
Remove unneeded include statements, remove conditional statements and
replace the remaining assert.h by their standard C++ variant cassert.


stweil:
This partially supersedes PR #1605.

PR #1698, opened 2018-06-22
Remove unneeded include statements for scanutils.h

stweil:

### DIFF
--- a/src/ccmain/reject.cpp
+++ b/src/ccmain/reject.cpp
@@ -18,7 +18,6 @@
  **********************************************************************/
 
 #include "tessvars.h"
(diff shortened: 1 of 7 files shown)


egorpugin:
Also see
https://include-what-you-use.org/
https://github.com/include-what-you-use/include-what-you-use

It is used in LLVM.

stweil:
Thank you for the hint. `include-what-you-use` is really a nice tool for cleaning the Tesseract code base.

egorpugin:
Tried it on tesseract on windows -
https://pastebin.com/bPkT2iwk

PR #1703, opened 2018-06-24
Clean more include files and include statements

stweil:


egorpugin:
Nice work!

stweil:
... but still some more work to be done.

PR #1733, opened 2018-07-03
Clean up cutil.h

stweil:
* Remove unrelated include statements from cutil.h.
* Remove macros FALSE, TRUE.
* Move macro CHARS_PER_LINE from cutil.h to dict.h.
* Remove unneeded macro _ARGS.
* Remove unused typedef statements.
* Remove macro new_line (only used once).
* Remove unused macro print_string.
* Update include statements for other source files.


amitdo:
Thanks for your work! (in general, not just this one)

PR #1744, opened 2018-07-04
 Fix build with legacy engine disabled

stweil:


amitdo:
>Instead of defining the DISABLED_LEGACY_ENGINE macro in config_auto.h
(which is not included by all source files), define it as a preprocessor
option for those parts of the code which require it.

I initially did just that, but then change my mind, thinking the second option is better.

amitdo:
I approve this PR. Thank you!

amitdo:
BTW, CMake also needs to be updated.

PR #1746, opened 2018-07-04
Fix some warnings from clang compiler

stweil:


amitdo:
>#if 0 // TODO: unused, remove if still unused after 2020.
 static bool NullIsBest(const NetworkIO& output, float null_thr,

?

stweil:
I was not sure whether that code (which is currently unused) might be needed for some future improvements.

stweil:
Maybe I should have used `#ifdef GOOGLE_TESSERACT` instead of `#if 0`. But with the current code we will know whether Google is missing that special function and can switch to `GOOGLE_TESSERACT` in that case.

zdenop:
There is no reason why code used only internally in google should be in community code.
I really appreciate the clean up work sweilt do. IMO all not used code should be removed.

Issue #1748, opened 2018-07-05
Missing ] at end of [Series]! : Failed to create network from spec ?


srdg:
### Environment

* **Tesseract Version**: 4.0.0-beta.1-376-gb1f79
* **Platform**: Ubuntu 16.04 LTS

### Current Behavior:
I'm trying to specify the network architecture entirely by myself. I used the architecture in the following command to train from scratch : 

`lstmtraining --debug_interval 0   --traineddata ./digits/digits.traineddata   --net_spec '[1,60,0,1 Ct3,3,32 Ct3,3,64 Mp2,2 Fr128 01c111]'   --model_output ~/tesstutorial/o/base --learning_rate 20e-4   --train_listfile ./digits.training_files.txt --max_iterations 5000 &>~/tesstutorial/o/basetrain.log`

What I get is this:
```
Invalid network spec:01c111]
Missing ] at end of [Series]!
Failed to create network from spec: [1,60,0,1 Ct3,3,32 Ct3,3,64 Mp2,2 Fr128 01c111]
```
The [documentation](https://github.com/tesseract-ocr/tesseract/wiki/VGSLSpecs#plumbing-ops) says that `[....]` indicate the network to be built as a series of layers, and I want to do just that. What am I missing here?

srdg:
Okay, I saw that I had used a 0 instead of an O in the last word. I replaced that, and the error is still persisting. If I'm not wrong, the error stems from the specification of the Fully Connected layer. My syntax is correct according to the [documentation](https://github.com/tesseract-ocr/tesseract/wiki/VGSLSpecs#functional-ops) - `input height and width *must* be constant` - do I keep them constant in the very first word itself?

amitdo:
Where is the lstm layers?

srdg:
@amitdo I went through the code. Something about it seems off. I'll try to explain below.
`int depth = strtol(*str + 1, str, 10);`
If I'm not wrong, `**str` holds the words in the VGSL specification, and `*str` holds a single word, lets say, `Fr64`. Now, the `strtol` function modifies `str` to point to the first character which a non-digit number, and assumes that ` *str+1 ` points to a number (of valid integer format) as a string (automatically skipping all the white spaces, and no other characters), where in reality, it seems to point to the memory address of `r` in `Fr164`. Isn't that bad argument? I don't see any other way in which `Fr164 ...` passed as argument to `strtol` could return 0 as a result.

amitdo:
[...]

PR #1761, opened 2018-07-06
Fixed "Invalid F spec" error while building FullyConnected layer.

srdg:
`int depth = strtol(*str + 1, str, 10);`
`**str` holds the words in the VGSL specification, and `*str` holds a single word, lets say, `Fr64`. Now, the `strtol` function modifies `str` to point to the first character which a non-digit number, and assumes that ` *str+1 ` points to a number (of valid integer format) as a string (automatically skipping all the white spaces, and no other characters), where in reality, it seems to point to `r` in `Fr164`.This is a bad argument, which results in `strtol` returning 0.
` strtol (*str + 2, str, 10)` should be passed instead.

Kindly check issue #1748 for the full thread of discussion.


amitdo:
:+1:  :smiley:

srdg:
@stweil @amitdo Happy to help!

PR #1784, opened 2018-07-18
Simplify API for serialization and add first users

stweil:
The current code uses fread (+ byte swapping) / fwrite (without byte swapping) and FRead or FReadEndian/ FWrite for de-serialization and serialization. The resulting file is always written in host endianness, so reading on a host with different endianness requires byte swapping. All functions return the number of elements read or written which must be compared with the intended value.

The new code always uses Deserialize / Serialize methods which return true for success or false for failure, like it is already done with the existing Deserialize / Serialize methods for a lot of classes.

With the new API it will be very easy to switch to a fixed little endian file format as soon as it is used for all serialization code.

### DIFF
--- a/src/api/pdfrenderer.cpp
+++ b/src/api/pdfrenderer.cpp
@@ -669,7 +669,7 @@ bool TessPDFRenderer::BeginDocumentHandler() {
   }
   fseek(fp, 0, SEEK_SET);
   const std::unique_ptr<char[]> buffer(new char[size]);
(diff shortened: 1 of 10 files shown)


amitdo:
Maybe we should move to Protocol Buffers or FlatBuffers.

TensorFlow uses the first, while TF Lite uses the second.

egorpugin:
Aren't templates are better?

amitdo:
About compatibility. For some time, the code should still keep the ability to **read** the old format. The write should always be done using the new format.

Yes, after 4.0.0 ...

Issue #1782, opened 2018-07-17
Cannot continue from _checkpoint


Shreeshrii:
Ubuntu 16.04.4 LTS (GNU/Linux 4.4.0-122-generic ppc64le)

 tesseract -v
tesseract 4.0.0-beta.3-180-gab1f
 leptonica-1.76.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.4.2) : libpng 1.2.54 : libtiff 4.0.6 : zlib 1.2.8 : libwebp 0.4.4 : libopenjp2 2.3.0

and

Ubuntu 14.04.5 LTS (GNU/Linux 4.4.0-130-generic x86_64)

 tesseract -v
tesseract 4.0.0-beta.3-199-gba757
 leptonica-1.76.0
  libjpeg 8d (libjpeg-turbo 1.3.0) : libpng 1.2.50 : libtiff 4.0.3 : zlib 1.2.8 : libwebp 0.4.0 : libopenjp2 2.3.0

----------------------
```
At iteration 4988/5000/5000, Mean rms=3.855%, delta=17.627%, char train=75.487%, word train=98.179%, skip ratio=0%,  wrote checkpoint.

Finished! Error rate = 54.122
shree@sanskrit:~/tesseract-ocr/tesseract$ lstmeval \
>   --model ../tesstutorial/engoutput/base_checkpoint \
>   --traineddata ../tesstutorial/engtrain/eng/eng.traineddata \
>   --eval_listfile ../tesstutorial/engeval/eng.training_files.txt
../tesstutorial/engoutput/base_checkpoint is not a recognition model, trying training checkpoint...
Loaded 72/72 pages (1-72) of document ../tesstutorial/engeval/eng.Impact_Condensed.exp0.lstmf
Deserialize failed
```

```
~/tesseract-ocr/tesseract$ lstmtraining \
>   --model_output ../tesstutorial/impact_from_small/impact \
>   --continue_from ../tesstutorial/engoutput/base_checkpoint \
>   --traineddata ../tesstutorial/engtrain/eng/eng.traineddata \
>   --train_listfile ../tesstutorial/engeval/eng.training_files.txt \
>   --max_iterations 1200
Loaded file ../tesstutorial/engoutput/base_checkpoint, unpacking...
Failed to continue from: ../tesstutorial/engoutput/base_checkpoint
```

Continue_from is NOT working currently from _checkpoint, however it works from the numbered checkpoints.

[...]

PR #1794, opened 2018-07-20
Increase limit and add assertions for deserialization of large arrays

stweil:
One of the checks was too restrictive, as lstmeval deserializes
char arrays with 14000000 elements, so raise the limit to 30000000.
That check was added in commit 992031e8248c3b95cda23315bf390303c7c7955c.

Add also assertions which help finding such problems in debug mode.

Signed-off-by: Stefan Weil <stweil@ub-backup.bib.uni-mannheim.de>


stweil:
This fixes issue #1782.

PR #1812, opened 2018-07-31
Fix issue reported by Coverity Scan

noahmetzger:
CID: 1375395 (Dereference after null check)

### DIFF
--- a/src/lstm/recodebeam.cpp
+++ b/src/lstm/recodebeam.cpp
@@ -512,8 +512,10 @@ void RecodeBeamSearch::ContinueContext(const RecodeNode* prev, int index,
            (previous->duplicate || previous->code == null_char_)) {
       previous = previous->prev;
     }
-    prefix.Set(p, previous->code);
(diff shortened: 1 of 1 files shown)


src/lstm/recodebeam.cpp
```diff
@@ -512,8 +512,10 @@ void RecodeBeamSearch::ContinueContext(const RecodeNode* prev, int index,
            (previous->duplicate || previous->code == null_char_)) {
       previous = previous->prev;
     }
-    prefix.Set(p, previous->code);
-    full_code.Set(p, previous->code);
+    if (previous != nullptr) {
```
stweil: Do you think that `previous == nullptr` would be possible and valid here? If not, adding `ASSERT_HOST(previous != nullptr);` might be better.

noahmetzger: i think so, the while loop just checks if the current node in the chain is not a nullpointer. If the next linked member is a nullpointer, it is assigned to previous and the loop ends.

PR #1818, opened 2018-08-01
Fix potential crash with --psm 0 and use osd.traineddata automatically

stweil:
Page segmentation mode "OSD only" requires osd.traineddata,
so use it automatically.

Report a warning if the user specified a different language.


amitdo:
https://github.com/tesseract-ocr/tesseract/blob/91c7504a35310ed576d9522211373f81f49bd854/src/api/baseapi.cpp#L366

Issue #837, opened 2017-04-23
Unusable OCR results with OpenCL


stweil:
Both 3.05 and latest git master produce unusable results with OpenCL (tested on Linux and macOS). It looks like that problem only occurs with [large images](https://digi.bib.uni-mannheim.de/~stweil/tesseract/0604.jp2).

Disabling OpenCL in `ccmain/thresholder.cpp` fixes the problem.

amitdo:
#635
Maybe it the right time to remove the OpenCL code entirely.

I have a feeling that you (Stefan) will disagree...  :)

stweil:
Indeed. :-)

Other projects using LSTM strive to use the computing power of graphic cards.

Should Tesseract be an exception because people are satisfied with the CPU based computation, should we try to fix the known problems of the current OpenCL based implementation and improve the code, or is there another solution how Tesseract can use advanced computing power?

amitdo:
The problems I see with the OpenCL code:
It's mostly duplicates parts of Leptonica, which is an **external** dependency.
It was contributed by someone from AMD. He does not keep maintaining it.
Google does not seems to want to invest time in maintaining it.
The only one that cares and tries to fix bugs is you.

 I respect your work, but I still think the benefit here is quite small and we should drop it.  

A compromise could be that you will maintain that code as a separate project and we will have an optional dependency on it.

amitdo:
I tried to sell it to Dan but he didn't buy it :laughing: 
https://github.com/tesseract-ocr/tesseract/issues/635#issuecomment-270327486

stweil:
I removed a duplicated part of Leptonica in PR #843. Are there more of them? At least I no longer find a comment naming Leptonica in the OpenCL code.

amitdo:
At least some with 'pix' in their names. pixSubtract for example.

jbreiden:
There's parallel set of functions for reading TIFF files.

https://github.com/tesseract-ocr/tesseract/blob/master/opencl/openclwrapper.cpp
https://github.com/DanBloomberg/leptonica/blob/master/src/tiffio.c

OpenclDevice::pixReadTiffCl() and friends
pixReadTiff() and friends

That whole section is a little weird for a few reasons.

1) TIFF decode is not very expensive, why bother with hardware assist?
2)  The OpenCL TIFF code has fallen behind Leptonica proper, enough to be
currently disabled.

but most interestingly to me

[...]

PR #1819, opened 2018-08-01
Fix ImageThresholder::OtsuThresholdRectToPix for OpenCL

stweil:
The ThresholdRectToPix OpenCL kernel only supports 4 channels.


stweil:
This seems to fix issue #837.

PR #1876, opened 2018-08-28
unittest: Fix and enable applybox_test

stweil:
It needs an update of the test submodule.

The tests only pass with a small modification of the ground truth texts
(kTruthTextWords, kTruthTextLine).


Shreeshrii:
Submitted PR https://github.com/tesseract-ocr/tesseract/pull/1884

stweil:
Yes, that's correct. I missed that dependency because my build with `configure --disable-shared` worked fine. Without `--disable-shared`, I get the linker error.

PR #1892, opened 2018-09-03
Improve snap packaging

Lin-Buo-Ren:
This PR fixes some issues regarding the snap packaging.


Lin-Buo-Ren:
The c-i build result doesn't seemed to be the result of this PR, just saying.

PR #1897, opened 2018-09-04
Fix some compiler warnings from clang

stweil:


src/ccstruct/pdblock.cpp
```diff
@@ -198,11 +198,11 @@ void PDBLK::plot(                //draw outline
     //              tprintf("Block %d bottom left is (%d,%d)\n",
     //                      serial,startpt.x(),startpt.y());
     char temp_buff[34];
-    #if defined(__UNIX__) || defined(MINGW)
+#if !defined(_WIN32) || defined(MINGW)
```
amitdo: I think `__MINGW32__` should be used here instead of `MINGW`.
https://sourceforge.net/p/predef/wiki/Compilers/

stweil: You are right. I addressed this in an additional commit.

PR #1917, opened 2018-09-19
Fix typo in cppan.yml affecting /openmp

cyanfish:
As far as I can tell the "official" cppan builds are fine, but when building from source, openmp wasn't working.


egorpugin:
Thanks!

PR #1921, opened 2018-09-20
Update edgblob.h

zhangxz189:


zdenop:
Can your provide details/úder čase fór this change?

zhangxz189:
if (buckets[[1]) is not empty,  start_scan, scan_next always return  &(buckets[[1]), because  index value has not  changed.

zdenop:
Can you provide image/user case to demonstrate this problem?

amitdo:
The loop already has `index++`

zhangxz189:
in loop index not changed. test case:

static int array[12] = { 0, 0, 1, 2, 3, 4, 5, 6, 0, 7, 8, 0 };
static int index;
static int start_scan() {
	for (index = 0; array[index]==0 && index < 12; index++);
	return array[index];
}
static int scan_next() {
	for (; array[index]==0 && index < 12; index++);
	return array[index];
}

int main(int argc, char** argv) {
	int scan = start_scan();
	printf("start scan: index=%d, out=%d\n", index, scan);
	for (int i = 0; i < 20; i++) {
		scan = scan_next();
		printf("scan next: index=%d, out=%d\n", index, scan);
	}
	return 0;
}

output result:
start scan: index=2,  out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1
scan next: index=2, out=1

stweil:
@zhangxz189, can you trigger the problem with the `tesseract` executable or the Tesseract API?

src/textord/edgblob.h
```diff
@@ -46,12 +46,12 @@ class OL_BUCKETS
     C_OUTLINE_LIST *start_scan() {
       for (index = 0; buckets[index].empty () && index < bxdim * bydim - 1;
         index++);
-      return &buckets[index];
+      return &buckets[index++];  //index value has not changed
```
stweil: The old code works because the for loop is left with `buckets[index]` being not empty, but being emptied later before calling the `scan_next` method. That also applies to the results of `scan_next`. See calling code in function `empty_buckets`.

Increasing `index` like it is done here would be needed in other contexts (like the test case in the descriptions) and could also improve the performance a little bit, but needs special handling for the corner case where `index` reaches the upper limit.

zdenop:
@zhangxz189 : can you please reply?

stweil:
@zhangxz189, do you plan to improve your pull request (see my comments above)?

src/textord/edgblob.h
```diff
@@ -46,12 +46,12 @@ class OL_BUCKETS
...
     }
                                  //next non-empty bucket
     C_OUTLINE_LIST *scan_next() {
       for (; buckets[index].empty () && index < bxdim * bydim - 1; index++);
-      return &buckets[index];
+      return &buckets[index++];  //index value has not changed
```
aleksey-nikolaev: it is a very bad idea making post-increment here, because for next `scan_next` you may catch segfault in buckets[index] ( index == bxdim * bydim)
But in old code `scan_next(); scan_next(); scan_next();` never find the next nonempty  element (a test for this: https://godbolt.org/z/vRbwUl)
Rather than making "beautiful" one-row code - make it readable:
```
C_OUTLINE_LIST *scan_next() {
  for (;  index + 1 < bxdim * bydim ; index++)
    if(!buckets[index].empty()) return &buckets[index++]; // next position in array
  return &buckets[index]; // last position. Don't increment!
}
```
Also, avoid copy-paste:
```
C_OUTLINE_LIST *start_scan() {
  index = 0;
  return scan_next();
}
```
or if you really want one-row code
```
C_OUTLINE_LIST *start_scan() {
  index = -1; // in this case we should start from -1
  return scan_next();
}
C_OUTLINE_LIST *scan_next() {
  // first of all check index!
  for (;  index + 1 < bxdim * bydim && buckets[++index].empty() ;);
  return &buckets[index];
}
```

amitdo:
@stweil, can you fix the issue?

egorpugin:
FYI:
If we `++it` here before scan to start from the next element, results are incorrect.
https://github.com/tesseract-ocr/tesseract/blob/f83f00496ed0d0291505018f1e7ac15c4b64388b/src/textord/edgblob.cpp#L90

PR #1923, opened 2018-09-20
Don't trigger a deliberate SIGSEGV for fatal errors in release code

stweil:
The error message "segmentation fault" confuses most users,
so enforce a segmentation fault only in debug code.

Release code simply calls the abort function.


amitdo:
Related PR: #341

PR #1942, opened 2018-10-04
issue 1586 lstmtraining --model_output doesn't generate new folders

### LINKED ISSUE #1586: lstmtraining --model_output doesn't generate new folders
When running lstmtraining:
```
lstmtraining --model_output ./output/
```
If the destination folder does not exist it wont create it, the program will keep training but the models are not saved, until you create the folder manually.

Also can't extract a traineddata model if the destination folder does not exist when running combine_tessdata:
```
combine_tessdata -e eng.traineddata ./output/eng.lstm
```

Basically what I am saying is that if the destination folder does not exist, create it automatically.

zdenop:
Fixes #1586 : 
- extract directory from filename path (used in combine_tessdata)
- create output directory if not exists (it does not create directory recursively!)


zdenop:
Tested on linux and windows  (VS2017 build). Would be nice if somebody with Mac could test it:
```
$ combine_tessdata -e eng.traineddata middle_earth/elf.version

$ lstmtraining --traineddata tessdata/eng.traineddata --model_output models --train_listfile tesstutorial/engtrain/eng.training_files.txt
```

stweil:
I think that for creating a directory which the user could easily create himself/herself this is a lot of code. Linux and macOS users only need the very short shell command `mkdir -p PATH`, and for Windows it is not much more difficult.

I'm also afraid that the real problem from issue #1586 is not fixed. Writing the output file can not only fail because of missing directories, but also because of an illegal file name, missing rights in the output directory, read only file system, disk full etc.

Creating directories is also not a job which is expected from an OCR software. Even the POSIX `cp` command won't create directories unless it is explicitly told to do so. Normally it will fail with an error message.

Therefore I'd prefer fixing the error handling, report any errors and fail if an expected output file cannot be written.

zdenop:
You are right, but we my intention is to fix at least what was it reported... 
As you can see code it not only about creating output, but also checking if the directory is accessible or output is file instead of directory.

The current state is that lstmtraining will run (I did not testing how long) and produce no output... We can leave it as it and wait ages for perfect solution or we can take baby steps and evolve.

src/training/lstmtraining.cpp
```diff
@@ -78,6 +78,8 @@ int main(int argc, char **argv) {
     tprintf("Must provide a --model_output!\n");
     return 1;
   }
+  if (CreateDirIfNotExists(FLAGS_model_output.c_str(), "model_path"))
+    return 1;
   if (FLAGS_traineddata.empty()) {
     tprintf("Must provide a --traineddata see training wiki\n");
     return 1;
```
stweil: The following code declares `STRING model_output` and fills it with a purified model name, but that new name is never used. That looks wrong. Maybe the intention was to use `model_output` instead of `FLAGS_model_output`, but that was never finished.

src/training/lstmtraining.cpp
```diff
@@ -78,6 +78,8 @@ int main(int argc, char **argv) {
     tprintf("Must provide a --model_output!\n");
     return 1;
   }
+  if (CreateDirIfNotExists(FLAGS_model_output.c_str(), "model_path"))
```
stweil: `FLAGS_model_output` is the basename for output models, not a directory name.

src/training/lstmtraining.cpp
stweil: Should we remove the dead code, or should we use it?

zdenop: If we use code it not dead anymore ;-)

src/training/lstmtraining.cpp
zdenop: so GetDirectory should be used first...

PR #1944, opened 2018-10-04
Allow orientation detection with any traineddata

stweil:
While orientation and script detection (OSD) normally requires
osd.traineddata to detect both, it must also be possible to do
only orientation detection with eng.traineddata or any other
traineddata.

Enforce osd.traineddata only if there was no `-l` command line option.

Commit 27ce472666a6332f38888cd6855e1162fb82c729 was too restrictive.

### DIFF
--- a/src/api/tesseractmain.cpp
+++ b/src/api/tesseractmain.cpp
@@ -367,11 +367,15 @@ static void ParseArgs(const int argc, char** argv, const char** lang,
   *arg_i = i;
 
   if (*pagesegmode == tesseract::PSM_OSD_ONLY) {
-    // That mode requires osd.traineddata, no other language or script files.
+    // OSD = orientation and script detection.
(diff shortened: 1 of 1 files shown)


amitdo:
https://github.com/tesseract-ocr/tesseract/wiki/ReleaseNotes

>Made 'osd' the default traineddata when psm 0 is requested (currently this feature is only implemented in the command line interface, but not in the API).

stweil:
@amitdo, thank you for updating the release notes. That's very important for 4.0.0.

I don't think that the API needs to be changed here, too. Users of the API will hopefully know what they do or have a look at the code for the `tesseract` executable and see the comments.

amitdo:
>(currently this feature is only implemented in the command line interface, but not in the API).

Should this remark be removed?

PR #1946, opened 2018-10-04
Don't set page segmentation mode for unlv config

stweil:
Setting the page segmentation mode to 6 ("Assume a single uniform block
of text") typically improves the layout detection for such texts, but
should not be done in the config file.

unlvtests/runtestset.sh adds `--psm 6` explicitly, so test results
won't change when using that script.

This is similar to commit ecfee53bac59e5467f8a8fb5026928a1d39b5c6d.

### DIFF
--- a/tessdata/configs/unlv
+++ b/tessdata/configs/unlv
@@ -1,2 +1,1 @@
 tessedit_write_unlv 1
-tessedit_pageseg_mode 6


amitdo:
I wonder if someone uses this output format...

stweil:
... me too. But it is used in the UNLV test script, so there is at least one user.

amitdo:
Also related is a special file format used with the unlv images.
https://github.com/tesseract-ocr/tesseract/search?q=uzn&unscoped_q=uzn

It gives tesseract blocks coordinations.

PR #1951, opened 2018-10-05
combine_tessdata, lstmtraining: Check for write failures

stweil:
This is done by creating a temporary file.
Report an error and terminate if that fails.

Use also EXIT_SUCCESS and EXIT_FAILURE for the return values of main().


stweil:
Now a second commit also handles `combine_tessdata`.

zdenop:
thanks

PR #1953, opened 2018-10-06
lstmtraining: Remove dead code for purified model name

stweil:
The purified model name `model_output` was unused,
so remove the comment and the unused code.


amitdo:
Why not change the code to use this string, instead of dropping the code?

amitdo:
Okay :-)

Issue #1205, opened 2017-11-09
some images translated to text using Tesseract 4 throw an error regarding "contains_unichar_id"


sallyhill:
### Environment

* **Tesseract Version**: 4.00.00alpha
* **Commit Number**: I used brew install tesseract --HEAD to install
* **Platform**: 15.6.0 Darwin Kernel Version 15.6.0: Mon Oct  2 22:20:08 PDT 2017; root:xnu-3248.71.4~1/RELEASE_X86_64 x86_64 (osx)
Files affected:
![bptbo](https://user-images.githubusercontent.com/12820071/32625770-8c5f9792-c563-11e7-8542-543a2aa0e9a3.png)
![jithy](https://user-images.githubusercontent.com/12820071/32625771-8c733d9c-c563-11e7-9538-9053c9f7c542.png)
![lrggj](https://user-images.githubusercontent.com/12820071/32625772-8c86419e-c563-11e7-8996-4acd3389f5f5.png)
![ouifv](https://user-images.githubusercontent.com/12820071/32625773-8c9adc3a-c563-11e7-9b5b-d71283c8c0d5.png)

### Current Behavior:
text to string of these images throws a TesseractError that prints: (-6, 'contains_unichar_id(unichar_id):Error:Assert failed:in file ../ccutil/unicharset.h, line 513') on the attached files

### Expected Behavior:
No error.

### Suggested Fix:
I am not sure. Right now I'm just running pytesseract.image_to_string in a try block

sallyhill:
Many other couple thousand images going in worked well, but these images were the error images.

psinger:
Any solution to that? Having similar issues.

amitdo:
Does this happen with:
 * `--oem 1`
 * `--oem 0`

?


@sallyhill , @psinger

psinger:
@amitdo 

I just tried it, and it works with both options. I thought `--oem 1` is the default anyways.

Any idea what's going on?

amitdo:
Seems like a bug in combining the two OCR engines.

psinger:
Any way to track this down further?

amitdo:
You can use GDB to see the function call chain.

Frankly, I only use --oem 1 (or 3 with best/fast traineddata), so I'm not so motivated to invest time on this issue. Sorry.

syzer:
:+1:

stweil:
I get the reported assertion with the second image (all other images work for me) and will have a look.

[...]

PR #1954, opened 2018-10-06
Fix use of wrong UNICHARSET

stweil:


stweil:
This should fix issue #1205. I found it by adding an assertion more up in the call stack. There might be more similar code using the wrong UNICHARSET, so I'll try to add more assertions to find it.

PR #1957, opened 2018-10-06
Fix some warnings from static code analyzer LGTM

stweil:
See https://lgtm.com/projects/g/tesseract-ocr/tesseract/alerts.


amitdo:
LGTM still complains:

>Resource data_ is acquired by class GenericVector<FontSpacingInfo *> but not released in the destructor. It is released from reserve on line 691, so this function may need to be called from the destructor.
This alert was introduced in04703ca 14 days ago

stweil:
That's a so called false positive, a wrong message. See [this discussion](https://discuss.lgtm.com/t/false-positive-for-resource-not-released-in-c-destructor/1386). We have to wait for a fixed LGTM.

PR #1960, opened 2018-10-06
Rework check for readable input file

stweil:
This reverts commit 1a096441d05a76e72145859cabb5413bf9464dfa and
implements an alternate check which allows input from stdin.


zdenop:
IMO correct place for testing input (file) should be calling application and not API!

stweil:
I agree. And I wonder whether handling of `stdin` and parsing of lists with image names should be part of an OCR API. I don't think so.

zdenop:
Not sure about. Parsing of lists with image filenames is solution for creating multipage pdf without conversion of images to multipage tiff.

PR #1964, opened 2018-10-08
Avoid crash with --psm 0 and LSTM traineddata

stweil:
Orientation and script detect only worked with legacy models
and crashed with LSTM models.

### DIFF
--- a/src/classify/adaptmatch.cpp
+++ b/src/classify/adaptmatch.cpp
@@ -1541,6 +1541,10 @@ void Classify::DoAdaptiveMatch(TBLOB *Blob, ADAPT_RESULTS *Results) {
                            &bl_features);
   if (sample == nullptr) return;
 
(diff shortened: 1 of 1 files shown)


amitdo:
What about using `mgr->IsBaseAvailable() ` ?

stweil:
Where do you suggest to use that?

amitdo:
I dunno... :-)

stweil:
Nor do I. :-)

amitdo:
I think this issue should be fixed upper in the call chain.

Probably here:
`bool Tesseract::init_tesseract_lang_data()`

Sorry, I don't have time for further research.

stweil:
I agree. Maybe it could even be fixed to work with LSTM, too, but I found no quick solution. That's why I added the TODO comment.

PR #1980, opened 2018-10-11
unittest: Add baseapi_test, fileio_test and qrsequence_test

stweil:


stweil:
The new test uses the Abseil API. Run `git submodule init && git submodule update` to get the code needed for `make check`.

PR #1983, opened 2018-10-12
unittest: Disable build rules for tests which still fail to build

stweil:
This fixes warnings from autogen.sh.


stweil:
Don't apply. It is bad.

stweil:
Fixed now, tested and ready to get merged.

PR #1985, opened 2018-10-13
win32: Show TIFF errors on console

stweil:
Showing them in a window (default) is not acceptable for a console
application like Tesseract which must be able to work in batch mode.

Such error messages can be triggered by TIFF files which include
vendor specific tags.

### DIFF
--- a/src/api/tesseractmain.cpp
+++ b/src/api/tesseractmain.cpp
@@ -42,6 +42,15 @@
 
 #include <tiffio.h>
 
+static void Win32ErrorHandler(const char* module, const char* fmt,
+                              va_list ap) {
+  if (module != nullptr) {
+    fprintf(stderr, "%s: ", module);
+  }
+  vfprintf(stderr, fmt, ap);
+  fprintf(stderr, ".\n");
+}
+
 static void Win32WarningHandler(const char* module, const char* fmt,
                                 va_list ap) {
   if (module != nullptr) {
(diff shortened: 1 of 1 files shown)


stweil:
@DanBloomberg, maybe a similar solution is needed for Leptonica, too.

DanBloomberg:
@stweil  
This appears to be a Windows specific mechanism for output of errors/warnings.  Should there be a unix version?

Is leptonica not giving enough error and warning information?

stweil:
Leptonica currently [sets the TIFF warning handler to a dummy function](https://github.com/DanBloomberg/leptonica/blob/master/src/tiffio.c#L2467) when it opens a TIFF file, so warning messages are suppressed. It does not change the error handler. Therefore errors (like unknown tags in TIFF files) are printed on Linux, but shown in a popup window on Windows. While thinking more about it, I think that it would be sufficient to let Leptonica do the error handling and add ` TIFFSetErrorHandler(dummyHandler); /* disable error messages */`.

DanBloomberg:
To be clear, you're suggesting adding 
TIFFSetErrorHandler(dummyHandler); /* disable error messages */.
at line 2467 of tiffio.c, below the similar call for the warning handler?
And this only disables error messages, not the actual handling of the errors?

Also, there are two other calls to TIFFSetWarningHandler().
What to do about them?

stweil:
All three locations need the same kind of handling.

http://www.libtiff.org/man/TIFFError.3t.html documents `TIFFSetErrorHandler`. The default warning and error handlers only show messages. According to the documentation, the dummy handler can be removed. It is sufficient to pass NULL (also for `TIFFSetWarningHandler`).

DanBloomberg:
OK, I'll add the error message suppression, using NULL for the arg.

stweil:
Thank you!

PR #1993, opened 2018-10-16
Fix several issues reported by Coverity Scan

stweil:


src/api/renderer.h
```diff
@@ -205,7 +205,6 @@ class TESS_API TessPDFRenderer : public TessResultRenderer {
   GenericVector<long int> pages_;    // object number for every /Page object
   std::string datadir_;              // where to find the custom font
   bool textonly_;                    // skip images if set
-  int jpg_quality_;                  // set JPEG quality
```
amitdo: Travis fails becaue of this change.

amitdo: https://github.com/tesseract-ocr/tesseract/blob/cbd09de7feca46543289d01b2d8482312ced2dd7/src/api/pdfrenderer.cpp#L188

amitdo: Travis failed because of this change.

You should also change the .cpp file.


https://github.com/tesseract-ocr/tesseract/blob/cbd09de7feca46543289d01b2d8482312ced2dd7/src/api/pdfrenderer.cpp#L188

stweil: @zdenop tried a different solution in commit cbd09de7feca46543289d01b2d8482312ced2dd7. I have to revert that commit.

stweil: Done now. The updated pull request should pass the CI tests.

amitdo: What he did was right (based on my suggestion).

He just forget to remove the part you now removed.

zdenop: thank stweil. I forget we do not need here.

PR #1997, opened 2018-10-17
Renamed GetGlyphConfidences() to GetChoices() and glyph_confidences t…

noahmetzger:
…o lstm_choice_mode

Renamed the global attribute glyph_confidences to lstm_choice_mode and the method GetGlyphConfidences() to GetChoices(). All Variables and comments contained in related methods were renamed as well.

Signed-off-by: Noah Metzger <noah.metzger@bib.uni-mannheim.de>

PR #1998, opened 2018-10-18
Rename API function for getting LSTM choices

stweil:


zdenop:
build failed:
[00:06:24] (ClCompile target) -> 
[00:06:24]   c:\projects\tesseract\src\api\baseapi.cpp(1564): error C2039: 'GetChoices': is not a member of 'tesseract::ResultIterator' [C:\projects\tesseract\build\libtesseract.vcxproj]
[00:06:24] 
[00:06:24]     1180 Warning(s)
[00:06:24]     1 Error(s)

bertsky:
I think the build fails because in baseapi.cpp in `TessBaseAPI::GetHOCRText()` the call has to be renamed as well.

bertsky:
+1 for putting LSTM in the name. Since this is still in flux, how about renaming it once more: `GetBestLSTMChoices`? 

(Rationale: this is not the complete Unicharset, only the data from `RecodeBeamSearch::SaveMostCertainChoices()`.)

stweil:
The build is fixed now, and I updated the name as suggested by @bertsky.

bertsky:
Thanks!

bertsky:
Sorry, I missed the discussion in #1997 completely. There was not much feedback in #1851, so I thought nobody was really interested. 

@amitdo, the problem with `GetSymbolChoices` is that we can have either symbols or LSTM timesteps behind that function. The level is controlled by `lstm_choice_mode` currently, and (hopefully) by an additional level parameter soon (`RIL_SYMBOL` vs `RIL_TIMESTEP`). IIRC the term "choice" is used in various places equally, but never applied to words. So keeping that ambiguity would be ok, don't you think?

egorpugin:
I think `Symbol` word must be somewhere in the name.

bertsky:
> IIRC the term "choice" is used in various places equally, but never applied to words

I was wrong there. And except for `ChoiceIterator` (which is currently limited to symbols but does not say so in its name), in fact usually the mnemonics makes the scope/level clear, i.e. there are `WordChoices`, `BlobChoices` etc.

So the only reason to keep "symbol" out of the name would be the analogy to `LTRResultIterator::GetChoiceIterator`.  

Arguably, both are about symbols, even on the timesteps level (where `duplicate` just marks the symbol boundaries).

amitdo:
https://github.com/tesseract-ocr/tesseract/search?q=%22word-%3Ebest_choices%22

https://github.com/tesseract-ocr/tesseract/search?q=%22word-%3Ebest_choice%22

https://github.com/tesseract-ocr/tesseract/search?q=%22word-%3Eraw_choice%22

amitdo:
It's not critical for me. I'll let others decide on this naming issue.

stweil:
I now renamed once more according to the last voting result.

Please wait with merging until we have seen that LGTM did check the training code.

stweil:
LGTM passed, but still did not check the training files. I think this PR can be merged. LGTM will repeat the checks for the master branch automatically, we just have to wait.

PR #2004, opened 2018-10-18
Add configuration for LGTM

stweil:
LGTM runs "make" by default, so training files are not checked.
This should be fixed by this configuration file.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,14 @@
+extraction:
+  cpp:
+    prepare:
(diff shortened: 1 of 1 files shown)


stweil:
Let's wait and see whether LGTM works as expected ... (could not test it otherwise).

zdenop:
Should I merge it or just wait?

stweil:
Please merge it, then we can see whether it works and checks `src/training`.

PR #2008, opened 2018-10-19
Fix configuration for LGTM

stweil:

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -11,4 +11,5 @@ extraction:
         - ../configure
     index:
       build_command:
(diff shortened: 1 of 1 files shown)


amitdo:
Is there a way to configure Coverity scan?

I see that it has warnings for abseil and googletest source files.

stweil:
In theory there is a way, but in real life there is a [bug in Coverity Scan](https://community.synopsys.com/s/question/0D53400003pnL0iCAE/getting-the-were-sorry-but-something-went-wrong-error) which makes that impossible since a few months.

PR #2015, opened 2018-10-21
fix unbound variables errors

mgeerdsen:
Since `set -u ` has been introduced in `tesstrain_utils.sh` several unbound variable errors have occured when using `tesstrain.sh`.
This PR assigns empty default values to several variables to avoid these errors.


zdenop:
thanks!

PR #2027, opened 2018-10-23
Fix compiler warning

stweil:
Compiler warning on macOS:

    tesscallback.h:29:7: warning:
      'TessClosure' has no out-of-line virtual method definitions;
      its vtable will be emitted in every translation unit [-Wweak-vtables]


egorpugin:
Did you try `virtual ~TessClosure() = default;` ?

PR #2030, opened 2018-10-27
Add no-copy version of SetImage

wh0:
This adds SetImageNoCopy on TessBaseAPI and ImageThresholder, which is suitable for cases where you have an image that's already in the right format and won't be using it for anything else.

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -631,6 +631,25 @@ void TessBaseAPI::SetImage(Pix* pix) {
 }
 
 /**
+ * Provide an image for Tesseract to recognize. Tesseract takes a clone of
+ * it, so it will persist until after Recognize. Returns true if successful
+ * or false if the input is not suitable.
+ */
+bool TessBaseAPI::SetImageNoCopy(Pix* pix) {
+  if (InternalSetImage()) {
+    if (pixGetSpp(pix) == 4 && pixGetInputFormat(pix) == IFF_PNG) {
+      tprintf("Pix with alpha channel passed to SetImageNoCopy. Use SetImage instead.\n");
+      return false;
+    }
+    if (!thresholder_->SetImageNoCopy(pix)) {
+      return false;
+    }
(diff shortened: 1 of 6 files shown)


zdenop:
What is real user case for this change?
Why user with "right image format" should use SetImageNoCopy  instead of SetImage?

amitdo:
I don't know if this PR is nessary, but I have a remark.

>This adds SetImageNoCopy on TessBaseAPI and ImageThresholder

For the C++ part. maybe use 
`bool TessBaseAPI::SetImage(Pix* pix, bool copypix) `

wh0:
Thanks for commenting.

> What is real user case for this change?

same as with SetImage, it's to provide an image for Tesseract to use

>  Why user with "right image format" should use SetImageNoCopy instead of SetImage?

They should use SetImageNoCopy in order to set the image without unnecessary copies and memory allocation.

> maybe use `bool TessBaseAPI::SetImage(Pix* pix, bool copypix)`

that's a good idea. might be possible to maintain backwards compatibility too with a default value

amitdo:
Question: Why does SetImage()  copy the image instead of using the ref counting mechanism of Leptonica's Pix?



This might be useful for people doing

amitdo:
The `pix_` image in `ImageThresholder` will be modified in several staged during Tesseract run. If the clone option will be used the user must aware of this fact. Also,  if this option is used the API user must not modify the image until all Tesseract's ocr phases are finished.

IMO, although less efficient it's safer to only support the copy option.

amitdo:
`void TessBaseAPI::SetImage(Pix *pix);`

With the current `SetImage()` in baseapi, you can destroy the pix right after calling this function, so it will live for a very short time. Tesseract will then use a copy of the pix.

amitdo:
BTW, it seems that in `baseapi.cpp `, after `SetImage(pix)` is called, the pix is not imminently destroyed. We can destroy it at that point in time.

PR #2031, opened 2018-10-27
Add debug configuration for LSTM

stweil:
It was provided by Jeff Breidenbach <jbreiden@google.com>.

### DIFF
--- a/tessdata/configs/lstmdebug
+++ b/tessdata/configs/lstmdebug
@@ -1,0 +1,4 @@
+stopper_debug_level 1
+classify_debug_level 1
+segsearch_debug_level 1
+language_model_debug_level 3


zdenop:
I am out but can your check if file is included in Makefile.am? I do not
sebe it there.

Dňa so 27. 10. 2018, 8:05 Stefan Weil <notifications@github.com> napísal(a):

stweil:
Do you think that file should be added to the distribution? Or is it only useful for developers who typically use git, so they can use it from the source tree?

zdenop:
IMO if it is in official report, it should be also installed and added to distribution. Otherwise it should be enough to document it on wiki.

jbreiden:
It's for developers, but I don't see a problem distributing.

PR #2037, opened 2018-10-30
Uses CMAKE_CURRENT_ variables instead of CMAKE_ variables

joelgallant:
This allows tesseract to be included as a subdirectory in other projects (my preferred way of handling 3rd party deps). I tested in a local project and everything else does seem to function as intended.


joelgallant:
Looks like the cmake config of tesseract is almost the same as leptonica (assuming it was copied at some point), similar PR there: https://github.com/DanBloomberg/leptonica/pull/387

amitdo:
The same guy (@egorpugin) maintains CMake and CPPAN for Leptonica and Tesseract.

egorpugin:
Thanks!

PR #2039, opened 2018-10-31
fix a couple minor compiler warnings

chrismamo1:


egorpugin:
Thank you!

src/ccutil/genericvector.h
```diff
@@ -983,7 +983,7 @@ template <typename T>
 bool GenericVector<T>::SkipDeSerialize(tesseract::TFile* fp) {
   uint32_t reserved;
   if (fp->FReadEndian(&reserved, sizeof(reserved), 1) != 1) return false;
-  return fp->FRead(nullptr, sizeof(T), reserved) == reserved;
+  return (uint32_t)fp->FRead(nullptr, sizeof(T), reserved) == reserved;
```
stweil: Please use C++ type casts instead of C type casts in future commits because C type casts also cause a compiler warning.

chrismamo1: Oops, sorry. I've been doing a lot of JS, Rust, and OCaml lately so I've got a few different ideas of what type casting looks like floating around in my head.

PR #2044, opened 2018-11-08
Remove dummy test

stweil:
This reverts commit 99755b0732c060440fd66a7fa58641d46f96ad00.
The dummy test is no longer needed as there exist a lot of real tests now.


stweil:
Is it sufficient to support testing only with autoconf, but not with cmake? I think yes, but maybe there are other opinions.

zdenop:
there were problems with cmake testings so I turn them off in 126ca8050c178769dc9149b8bab357ab6e3fc871

PR #2045, opened 2018-11-08
Remove gettimeofday.h

stweil:


stweil:
This is an old commit from my source tree which still does not use `std::chrono`, but eliminates the problematic `gettimeofday.h`. Switching to `std::chrono` can be done in a next step if desired.

zdenop:
I have ready std::chrono commit, so will to last check with yours code...

stweil:
Great. Then you can close this one as soon as your commit is ready.

PR #2051, opened 2018-11-14
Fix mktemp in tesstrain_utils.sh

johnlinp:
The commit 10f2c45c00 unified the usage of mktemp, but with a
incorrect bash syntax and unnecessary definition of LANG_CODE
and TIMESTAMP. This patch fixes the above problems.

### DIFF
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -25,14 +25,7 @@ else
 
 UNAME=$(uname -s | tr 'A-Z' 'a-z')
 
-case $UNAME in
-  darwin | *freebsd | dragonfly | cygwin*)
-    MKTEMP_DT="mktemp -d -t"
-    ;;
-  * )
-    MKTEMP_DT="mktemp -d --tmpdir"
-    ;;
-esac
+MKTEMP_DT="mktemp -d -t"
 FONT_CONFIG_CACHE=$(${MKTEMP_DT} font_tmp.XXXXXXXXXX)
 
 if [[ ($UNAME == *darwin*) ]]; then
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -24,18 +24,16 @@ else
  fi
 
 UNAME=$(uname -s | tr 'A-Z' 'a-z')
-LANG_CODE="ENG"
-TIMESTAMP=`date +%Y-%m-%d`
 
 case $UNAME in
   darwin | *freebsd | dragonfly | cygwin*)
-    MKTEMP_DT=$(mktemp -d -t)
+    MKTEMP_DT="mktemp -d -t"
     ;;
   * )
-    MKTEMP_DT=$(mktemp -d --tmpdir)
+    MKTEMP_DT="mktemp -d --tmpdir"
     ;;
 esac
-FONT_CONFIG_CACHE=(${MKTEMP_DT} font_tmp.XXXXXXXXXX)
+FONT_CONFIG_CACHE=$(${MKTEMP_DT} font_tmp.XXXXXXXXXX)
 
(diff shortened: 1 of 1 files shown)


zdenop:
I was looking for some good source for checking bash script quality and I found several references https://www.shellcheck.net/ which can be used online and offline.
Some of suggested changes it recognized as wrong (see inline comments), but it found several other issues. So I suggest to fix all 3 bash script with shellcheck.

amitdo:
Just like any other static analysis tool, there will be false positive alerts. 

https://github.com/koalaman/shellcheck/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+%22false+positive%22

https://github.com/koalaman/shellcheck/search?q=%22false+positive%22&type=Wikis

johnlinp:
@stweil You are right - `MKTEMP_DT` can be simplified to `mktemp -d -t` only. I pushed a new commit for that change. Please take a look at it. Thank you.

zdenop:
@johnlinp : I reply here and not in #2054, because here is right place for comments of this PR.

1. This commit fix nothing: this is right  statement, because there is no real problem that your commit solve fix. Changing syntax fix nothing.
2. You claim you fix wrong bash syntax. This is true only for `TMP_DIR=$(${MKTEMP_DT} ${LANG_CODE}-${TIMESTAMP}.XXX)`. I asked you for source (standard) that claim used syntax is wrong. I still did not get answer. Even I give you reference that new modern style for bash use `$(...)` only. And we try to modernize tesseract code.
3. If you want to fix syntax you need to fix it in all file, to be consistent. Now script use syntaxt for variables double-quotes, back-quoted and `$(...)`. Such changes are not improvement, because decrease code readability.

[Comment](https://github.com/tesseract-ocr/tesseract/issues/2054#issuecomment-438942597) report that this PR makes script even worse.

zdenop:
I am correcting myself: `TMP_DIR=$(${MKTEMP_DT} ${LANG_CODE}-${TIMESTAMP}.XXX)` and `FONT_CONFIG_CACHE=$(${MKTEMP_DT} font_tmp.XXXXXXXXXX)` seems to fix issue #2054.

But rest of my objection are IMO still valid.

johnlinp:
Hi @zdenop 

> 1. This commit fix nothing: this is right  statement, because there is no real problem that your commit solve fix. Changing syntax fix nothing.

There *is* a real problem to solve: #2054.

> 2. You claim you fix wrong bash syntax. This is true only for `TMP_DIR=$(${MKTEMP_DT} ${LANG_CODE}-${TIMESTAMP}.XXX)`. I asked you for source (standard) that claim used syntax is wrong. I still did not get answer. Even I give you reference that new modern style for bash use `$(...)` only. And we try to modernize tesseract code.

In your commit 10f2c45c00, you wrote:

```
MKTEMP_DT=$(mktemp -d -t)
```

which directly execute the command `mktemp -d -t`. However, `mktemp -d -t` is not supposed to executed directly, it should be executed with a `template` argument. Executing `mktemp -d -t` will cause warning messages like `mktemp: option requires an argument -- t`, which is the message described in #2054.

> 3. If you want to fix syntax you need to fix it in all file, to be consistent. Now script use syntaxt for variables double-quotes, back-quoted and `$(...)`. Such changes are not improvement, because decrease code readability.
> 
> [Comment](https://github.com/tesseract-ocr/tesseract/issues/2054#issuecomment-438942597) report that this PR makes script even worse.

This commit doesn't mean to fix syntax. This commit is to fix the problem like #2054.

stweil:
Merged, thank you. I added a commit to simplify the shell code even more.

zdenop:
I am sorry I fooled myself that you are just changing syntax. My bad - I should test it :-(

Issue #2054, opened 2018-11-15
Training tesseract tesstrain.sh exits with a warning


rakesh-km:
src/training/tesstrain.sh --fonts_dir /usr/share/fonts --lang eng --linedata_only \
  --noextract_font_properties --langdata_dir ../langdata \
  --tessdata_dir ./tessdata --output_dir ~/tesstutorial/engtrain

throws warning : 
mktemp: option requires an argument -- t
usage: mktemp [-d] [-q] [-t prefix] [-u] template ...
       mktemp [-d] [-q] [-u] -t prefix 

------------------------

### Environment

* **Tesseract Version**:  4.0.0
* **Commit Number**: NA
* **Platform**: Mac OS

rakesh-km:
now it throws error in language-specific.sh

=== Starting training for language 'eng'
/tesseract/src/training/language-specific.sh: line 1175: FONTS: unbound variable

zdenop:
@stweil: #2051 fix nothing

johnlinp:
> @stweil: #2051 fix nothing

Why did you say so? Is there any problem with my fix?

rakesh-km:
#2051 fixed the issue with mktemp. However the error now is the one that I have mentioned in the comment.

zdenop:
@rakesh-km: than you problem should be not with mktemp (on mac usage is the same as before patch), but with this 2 definitions (there is missing `$`):
```
line 38: FONT_CONFIG_CACHE=$(${MKTEMP_DT} font_tmp.XXXXXXXXXX)
line 212: TMP_DIR=$(${MKTEMP_DT} ${LANG_CODE}-${TIMESTAMP}.XXX)
```

stweil:
Yes, `$` was missing. Pull request #2051 fixes that, too.

stweil:
@zdenop, [line 32](https://github.com/tesseract-ocr/tesseract/blob/master/src/training/tesstrain_utils.sh#L32) is causing the first error message on macOS. The pull request fixes this line.

stweil:
The problem with error `unbound variable` seems to require lots of changes because it occurs not only on the line reported above. I can fix it by replacing `test -z "${VAR}"` by `test -z "${VAR:-}"`. Maybe `test -v VAR` would be better (at least for some cases).

I still don't know why this problem occurs with bash on macOS, but not on Linux. Hints are appreciated.

stweil:
[...]

PR #2056, opened 2018-11-15
Fix unbound variable $FONTS

johnlinp:
Fixes #2054

PR #2057, opened 2018-11-15
Fix some issues with the shell scripts for training

stweil:


stweil:
This replaces pull request #2056. It should fix issue #2054.

stweil:
Please don't apply this pull request. I just noticed that macOS uses an older version of bash which obviously does not support `test -v`. :-(

stweil:
Pull request #2058 replaces this one.

PR #2058, opened 2018-11-15
 Fix some issues with the shell scripts for training

stweil:


stweil:
This replaces #2057.

stweil:
I now updated the pull request after testing it on macOS – one more unbound variable had to be fixed.

stweil:
We should consider making a 4.0.1 release after the last fixes for the training scripts because the issue with unbound variables affects all platforms, not only macOS.

zdenop:
thanks

PR #2060, opened 2018-11-17
Fix wrong results from function streamtofloat

stweil:
The local variable k should be 10 ^ (number of digits after comma),
but will overflow when there are more than 9 digits after the comma
because an int value cannot store 10000000000.

This results in wrong double values read from .tr files for example
(or in a runtime exception if Tesseract was compiled with -ftrapv).

Using uint64_t does not fix the general problem but allows more digits
which should be sufficient for the data read by Tesseract.

### DIFF
--- a/src/ccutil/scanutils.cpp
+++ b/src/ccutil/scanutils.cpp
@@ -143,11 +143,11 @@ static uintmax_t streamtoumax(FILE* s, int base) {
 }
 
 static double streamtofloat(FILE* s) {
-  int minus = 0;
-  int v = 0;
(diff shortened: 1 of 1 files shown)


stweil:
I noticed that while working on `mastertrainer_test` which crashed while reading `-0.0035381317` in this line from `eng.Arial.exp0.tr`:

     -0.0035381317 0.2197364 0.24173248 0 0 0

Old versions of Tesseract have the same bug.

amitdo:
The `tprintf.h` header is not needed here.

PR #2061, opened 2018-11-17
Fix wrong results from function streamtofloat

stweil:
The local variable k should be 10 ^ (number of digits after comma),
but will overflow when there are more than 9 digits after the comma
because an int value cannot store 10000000000.

This results in wrong double values read from .tr files for example
(or in a runtime exception if Tesseract was compiled with -ftrapv).

Using uint64_t does not fix the general problem but allows more digits
which should be sufficient for the data read by Tesseract.

### DIFF
--- a/ccutil/scanutils.cpp
+++ b/ccutil/scanutils.cpp
@@ -144,11 +144,11 @@ uintmax_t streamtoumax(FILE* s, int base) {
 }
 
 double streamtofloat(FILE* s) {
-  int minus = 0;
-  int v = 0;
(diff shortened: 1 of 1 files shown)


stweil:
This is identical to pull request #2060, but for the 3.05 branch.

amitdo:
What's the reason for the reimplementation of fscanf()?

stweil:
It does not handle all locale dependencies, and I plan to replace it by C++ stream i/o.

PR #2065, opened 2018-11-19
fix(configure) Don't add rt on Android

guillaume-tgl:
Library rt is included in the libc on Android: https://developer.android.com/ndk/guides/stable_apis#a3


egorpugin:
Thanks!

Issue #1503, opened 2018-04-22
Port tesstrain.sh to Python or C++


egorpugin:
Since training tools are available on windows, it will be nice if we have training scripts ported too.
There are two ways:
1. Python script - which is probably the best option, and 
2. C++ program/script - my personal preference because of easier c++ dependencies integration with CPPAN - initial work (WIP) can be found in https://github.com/egorpugin/tesseract/commits/cpp_tesstrain

Contributions are appreciated.

amitdo:
I like the Python option :)

zdenop:
me too, but it increase dependacy for user that does not need python. Or we try to freezy it ;-)
Should we create extra repository for "test" like this (e.g. tools- experimental)?

egorpugin:
> Should we create extra repository for "test" like this (e.g. tools- experimental)?

I think this is not needed. We put more scripts in training dir as they are ready.
But still we can also use main repo for development of such tools to provide the best collaborative access.
I did not commit my changes to main repo just because I wanted to notify you first.
Porting of that script is pretty straightforward - prepare some data, run commands.

If we use separate repo - we could issue potential problems with integrating tess code - just extra burden.

hoangtocdo90:
vote python and yaml for conf file. C++ and shell is pain in the ass

stweil:
Python is also used for some related software, for example by the [hocr-tools](https://github.com/tmbdev/hocr-tools/). I think it would be a good choice.

PR #2076, opened 2018-11-24
RFC: Pythonize tesstrain.sh and friends

jbarlow83:
After attempting to chase down an error in tesstrain.sh I ended up converting to tesstrain.sh to Python. I also see interested was expressed by others in this conversion in #1503.

Only LSTM is converted right now. I have not done this work for the Tesseract 3 portions of training because IIRC, the old 3.x engine might get scrapped. If there's a need I can convert that too; the process is pretty mechanical. 

Python 3.6+ is required, which also ships with Ubuntu 18.04 and is always installed. I need Python 3.6 for its improved string substitution - this makes it follow the bash version more closely,

I attempted to keep source changes minimal so it would be easier to compare
bash to Python in code review and confirm equivalence. The first commit is structured so it's somewhat easier to compare to the originals in bash, e.g. keeping Python and bash the same for the rare cases where they are the same. The second is cleaned to PEP8 so it deviates further. I definitely appreciate anyone checking whether I've mistranslated something. ( #1503 suggests language_specific.py being presented in YAML, and I agree that would be better, but for now I want to avoid major restructuring.)

There are minor improvements already but it's more or less functionally equivalent for LSTM training. Python's logging library is used and the log file is handled more consistently. Console output is more terse when things are going well. There are progress bars. Temporary files are retained on failure, for inspection. It's generally easier to use.

The .cpp utilities generally depend on portions of the Tesseract code base. I don't plan to modify those.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,6 +73,7 @@ build/*
 *.a
 *.class
 *.jar
+__pycache__
 
 # tessdata
 *.traineddata
--- a/src/training/language_specific.py
+++ b/src/training/language_specific.py
@@ -1,0 +1,1294 @@
+#!/usr/bin/env python3
+# (C) Copyright 2014, Google Inc.
+# (C) Copyright 2018, James R Barlow
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Set some language specific variables. Works in conjunction with
+# tesstrain.sh
+#
+
+#=============================================================================
+# Language specific info
+#=============================================================================
+
+import os
+import logging
+log = logging.getLogger(__name__)
+
+# Array of all valid language codes.
+VALID_LANGUAGE_CODES=("afr amh ara asm aze aze_cyrl bel ben bih bod bos bul cat "
+                     "ceb ces chi_sim chi_tra chr cym cyr_lid dan deu div dzo "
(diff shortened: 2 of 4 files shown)


amitdo:
Nice work!

Maybe we should move this python code to a separate repo, `tesstrain-python`, hosted under the `tesseract-ocr` organization.

egorpugin:
> Maybe we should move this python code to a separate repo, tesstrain-python

I cannot agree. Too much repos.

amitdo:
I still prefer to separate it from the c++ code.

Can you move the python code to a new directory?

One option is `python` under the root dir, similar to the `java` dir.

amitdo:
Other options:

`/tools/training/python`
`/src/training/tools/python`

zdenop:
IMO it is nothing else than replacing bash script. if bash script was not problem with c++ why python?
For me bigger problem is that we mix library and program code (src/api/tesseractmain.cpp  + whole src/training/)

amitdo:
>if bash script was not problem with c++ why python?

With new code we can change old ways of doing things.

>For me bigger problem is that we mix library and program code (src/api/tesseractmain.cpp + whole src/training/)

This is related to my suggestion.

One possible solution:
`tools/tesseract.cpp`

jbarlow83:
Why don't we install tesstrain.py as tesstrain? The language used to write
an installed program like training tools should be an implementation
detail. That would facilitate moving other training tools to higher level
languages if desired. (It might be cleaner to move core training code into
libtesseract and have a Python wrapper that binds the APIs, some time in
the future.)

At the same time it is nice for source code files to have the proper
language suffix, hence the rename.

zdenop:
Move core training code will increase decency on external libs for all users - and most of them does not need them. (if you wanna have fun - try to use pango in python 3 on window ;-))
Separate training tools is right approach.

jbarlow83:
Python has all of the functionality of libicu baked in, and can render fonts so you cut out both dependencies by shifting to it. For example Pillow can render fonts. But if you specifically need pango and can't substitute another font library, yes, that would be painful.

I'm saying this more to put what could be a viable alternative on your radar - maybe it's a lower maintenance option long term. I won't be trying it.

PR #2079, opened 2018-11-25
Add example in README.md

ajschaeper:
Added an example in the README for newbies like me to have a place to get started.
Also added a link to the wiki where there are more helpful examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ Basic **[command line usage](https://github.com/tesseract-ocr/tesseract/wiki/Com
 
 For more information about the various command line options use `tesseract --help` or `man tesseract`.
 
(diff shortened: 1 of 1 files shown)


zdenop:
we have there in past 2 examples images and they were removed from source...

zdenop:
black&white text image: 126 KB seriously? no image size optimization?

stweil:
Please don't add new images (binaries!) to the tesseract repository, as they increase the size of the Git data too much.

ajschaeper:
Good points. 
* Removed everything, including the files
* Just left a link to the examples in the wiki

stweil:
Thank you for this update. It's merged now.

PR #2090, opened 2018-11-29
Optimize performance by using inline functions

stweil:


stweil:
This change only optimizes the calculations for best models (double precision), in my tests by a few percent for OCR. I did not run a performance test for training.

PR #2092, opened 2018-11-30
Format new ALTO code with clang-format

stweil:


stweil:
It would look like [this](https://github.com/stweil/tesseract/commit/8bcb5d5c4b545463019f017b3cd8bd06444e4efe). Do you think I should send a pull request, or do the format parameters need more fine tuning?

amitdo:
I will leave that decision to you...

PR #2095, opened 2018-11-30
Use -ffast-math for calculation of dot product

stweil:
This reduces the code size for intsimdmatrixavx2 from 2700 to 2668
and slightly improves the performance for fast models with AVX2.

### DIFF
--- a/src/arch/Makefile.am
+++ b/src/arch/Makefile.am
@@ -18,13 +18,13 @@ noinst_LTLIBRARIES = libtesseract_avx.la libtesseract_avx2.la libtesseract_sse.l
 noinst_LTLIBRARIES += libtesseract_arch.la
 
 if AVX_OPT
-libtesseract_avx_la_CXXFLAGS = -mavx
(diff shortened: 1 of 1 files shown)


stweil:
See [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html).

amitdo:
`-ffast-math` can change the result of floating point operation.

Isn't that risky? Did you run a test on many pages to see that it does not negatively affect accuracy?

If you agree that it should be tested, I suggest to test with gcc and clang.

For clang maybe you want to test with (1) libc++ and (2) libstdc++.

Maybe also test no-simd, just sse4, avx.

stweil:
With gcc 6.3 the generated code only changed for AVX2, so at least for that compiler the other cases will give identical results as before. I'll do some more tests.

stweil:
I‌ wonder whether we should use `fno-math-errno` for all code. It is the default on macOS, reduces the code size by about 4400 byte, and Tesseract does not test errno for math operations.

PR #2104, opened 2018-12-04
Fix two runtime errors

stweil:


amitdo:
>... resulting in wrong hOCR output

The hocr renderer should check the result it gets from calling that function:

```
if (result != nullptr && (*font_id != -1))  {
  // output bold, itatic etc.
}
```

stweil:
I noticed the problem with the wrong font attributes in new hOCR files: one of my machines produced all [pages with italic text](https://digi.bib.uni-mannheim.de/periodika/reichsanzeiger/ocr/film/tesseract-4.0.0-20181201/022-9903/0019.hocr).

I then used a [debug build with runtime checks](https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation#debug-builds) on another machine which only supports SSE. Here is the complete protocol of the test which I had run there:

    $ tesseract test/testing/HelloGoogle.tif /tmp/out hocr
    Tesseract Open Source OCR Engine v4.0.0 with Leptonica
    Page 1
    ../../../../../src/ccstruct/matrix.h:84:11: runtime error: null pointer passed as argument 1, which is declared to never be null
    ../../../../../src/ccstruct/matrix.h:84:11: runtime error: null pointer passed as argument 2, which is declared to never be null
    ../../../../../src/api/baseapi.cpp:1616:5: runtime error: load of value 128, which is not a valid value for type 'bool'
    ../../../../../src/api/baseapi.cpp:1627:5: runtime error: load of value 128, which is not a valid value for type 'bool'
    /usr/lib/gcc/x86_64-linux-gnu/6/include/emmintrin.h:704:24: runtime error: load of misaligned address 0x6340000154dc for type '__m64', which requires 8 byte alignment
    0x6340000154dc: note: pointer points here
      7f 87 d2 61 85 92 6e 7f  7f 7f 7f 7f 7f 7f 7f 7f  7f 7f 7f 7f 7f 7f 7f 7f  7f 7f 7f 7f 7f 7f 7f 7f
                  ^ 
    /usr/lib/gcc/x86_64-linux-gnu/6/include/emmintrin.h:704:24: runtime error: load of misaligned address 0x60e00000d24a for type '__m64', which requires 8 byte alignment
    0x60e00000d24a: note: pointer points here
     b2 bb  7f e9 e2 13 7f ba a1 ea  5b ca fb c5 81 e8 56 54  54 72 73 0b ad d6 be c9  f4 99 e5 3b 25 73
                  ^ 
    /usr/lib/gcc/x86_64-linux-gnu/6/include/emmintrin.h:704:24: runtime error: load of misaligned address 0x62200001b159 for type '__m64', which requires 8 byte alignment
    0x62200001b159: note: pointer points here
     05 fe 07  f2 d3 06 02 05 e8 05 11  0d ee ae fb 55 bb cf 95  59 43 f0 7f fe 20 13 15  18 d2 f8 20 f7

The first four runtime errors are fixed now, but the last three still have to be fixed: it looks like the SSE optimized code gets misaligned data.

PR #2118, opened 2018-12-13
protos: Remove several unused macros, functions and global variables

stweil:
The unused global variable TrainingData used a lot of runtime memory.


stweil:
The 2nd commit removes the unused config variable `MicroFeatures`. Is this acceptable, although in theory it might cause problems when it is still in use somewhere?

stweil:
Comparing sizes before and after these commits:

    2542223	  16064	2770240	5328527	 514e8f	tesseract (old)
    2540439	  16056	 410944	2967439	 2d478f	tesseract (new)

Tesseract was built with `--disable-shared`.

zdenop:
I did not find any usage (yet ;-) ) Let's try it :-)

PR #2121, opened 2018-12-15
Move code for hOCR renderer to new file 

stweil:


amitdo:
For consistrncy, I suggest to move all the remaining renderers out of baseapi.cpp.

PR #2134, opened 2018-12-27
RFC: Add support for image or image list by URL

stweil:
This allows OCR of images from the internet without downloading them first:

    tesseract http://IMAGE_URL OUTPUT ...

It uses libcurl.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -428,6 +428,14 @@ AC_CHECK_TYPES([mbstate_t],,, [#include "wchar.h"])
 # Test auxiliary packages
 # ----------------------------------------
 
+AM_CONDITIONAL([HAVE_LIBCURL], false)
+PKG_CHECK_MODULES([libcurl], [libcurl], [have_libcurl=true], [have_libcurl=false])
+if $have_libcurl; then
+  AM_CONDITIONAL([HAVE_LIBCURL], true)
+else
+  AM_CONDITIONAL([HAVE_LIBCURL], false)
+fi
+
 PKG_CHECK_MODULES([LEPTONICA], [lept >= 1.74], [have_lept=true], [have_lept=false])
 if $have_lept; then
   CPPFLAGS="$CPPFLAGS $LEPTONICA_CFLAGS"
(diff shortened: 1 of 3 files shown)


stweil:
Is this a feature which is useful for others, too?

I want to use it for distributed OCR of books and newspapers with thousands of pages to avoid
downloading of temporary image files to disk.

libcurl not only supports http: and https:, but also other schemas like ftp: or ssh:, which could be added
to the new code, too, if needed.

jbreiden:
Tesseract already accepts images on stdin. Why not use the curl binary and
pipes to avoid temp files?

jbreiden:
I also wonder if it might cause any issues for larger software that uses
Tesseract as a component, such as ffmpeg or opencv.

>

egorpugin:
Maybe it's possible to add this into tesseract binary, not library, in case if you need this on command line.
If you use tesslib it is possible to integrate curl into your app.

zdenop:
This works for me:
```
curl "http://d2jaiao3zdxbzm.cloudfront.net/wp-content/uploads/figure-65.png" | tesseract - -
```
If the feature is already available, way to add additional dependency to tesseract?

stweil:
The current implementation adds curl in baseapi,cpp, so it's part of the library. Of course I could make curl support optional with `--disable-curl` being the default configure setting.

The pipe solution needs an additional step to add the image name to the hOCR output.

stweil:
See https://groups.google.com/forum/#!topic/tesseract-dev/ZFn38UEOWM8 with a discussion to extend URL support to get traineddata models, too.

Shreeshrii:
I think this will be a useful addition, specially from command line.

The concerns raised are addressed by @stweil's suggestion:

> I could make curl support optional with --disable-curl being the default configure setting.

zdenop:
@Shreeshrii : see my comment at [tesseract-dev forum](https://groups.google.com/d/msg/tesseract-dev/ZFn38UEOWM8/VBmPEn2WAAAJ)

MakaWeli:
https://twitter.com/@xsamas12

MakaWeli:
https://twitter.com/@xsamas12

egorpugin:
Is PR ready for merging?
If yes, no need to keep it longer unmerged.

stweil:
> Is PR ready for merging?

Technically it is ready, but it currently only works with `autoconf` (CMake support is still missing), and there is also no `configure` option like `--disable-curl` which allows disabling it at build time.

Issue #1113, opened 2017-09-07
Error while running of Red Hat- DotProductAVX can't be used on Android


AzkaGilani:
I have used tesseract in my project. The project has been deployed on Red Hat server. It comiples fines on ubuntu and when i compile the same program in Red Hat it gives run time error: "DotProductAVX can't be used on Android." I have found other issues relating on internet similar to this, but the problem is that i can't update tesseract version or can not even rebuild it.
Is there any other solution by which i can solve this??
Should i make some change in my C++ code?

tesseract --version gives the following output:
`tesseract 0afd593
 leptonica-1.74
  libjpeg 8d (libjpeg-turbo 1.4.2) : libpng 1.2.54 : libtiff 4.0.6 : zlib 1.2.8

 Found AVX
 Found SSE
`

stweil:
So the build process did not add the code needed for AVX => `arch/dotproductavx.cpp` was compiled without `__AVX__` being defined => The compiler was called without `-mavx` for that file. As you did not tell us more about your build, you are currently the only one who can solve that problem. A full build log (output from configure / make) would help.

stweil:
The Tesseract version in the output from `tesseract --version` looks strange. Which code did you compile?

AzkaGilani:
I have compiled tesseract from source. 
Build is tesseract-4.0Alpha while (0afd593) is commit number on this build.

AzkaGilani:
I used following commands to build the project. It is working fine on ubuntu 16.04 but fails on Red Hat when tesseract function is called.

`sudo apt-get install autoconf automake libtool
sudo apt-get install autoconf-archive
sudo apt-get install pkg-config
sudo apt-get install libpng12-dev
sudo apt-get install libjpeg8-dev
sudo apt-get install libtiff5-dev
sudo apt-get install zlib1g-dev

wget http://www.leptonica.org/source/leptonica-1.74.tar.gz

tar -zxvf leptonica-1.74.tar.gz
cd leptonica-1.74
./configure
make
sudo make install

git clone git@gitlab.com:VisionX/core/libs/extern/tesseract.git
cd tesseract
git fetch origin
git branch -v -a
git checkout -b tesseract-4.0Alpha origin/tesseract-4.0Alpha
git pull origin tesseract-4.0Alpha
./autogen.sh
./configure --enable-debug
LDFLAGS="-L/usr/local/lib" CFLAGS="-I/usr/local/include" make
sudo make install
sudo ldconfig`

[...]

PR #2135, opened 2019-01-01
Fix and simplify SIMD tests

stweil:
The tests for SSE and AVX must only be done if the correct compiler
flags were used.


stweil:
This should fix issue #1113. The dummy implementations for SSE and AVX on Android are now removed, so we can be sure that they are never called.

amitdo:
Why do you need `-DAVX` when you already have `__AVX__`?

stweil:
`__AVX__` is set automatically by the compiler if the code is compiled with AVX support.

`simdetect.cpp` must work on hosts without AVX, too, so that code cannot be complied with AVX support. That means a macro which says that AVX is included (in another compilation unit!) is needed. I could have taken `__AVX__` as well, but macros starting with underscores are reserved for the system and the compiler. That's why I decided to use `AVX`.

PR #310, opened 2016-04-28
Character boxes in hOCR output

nickjwhite:

PR #2144, opened 2019-01-04
Fix headers

nickjwhite:
A couple of places where headers should have been included.

I found these while updating my simplemakev4 branch to build tesseract with make.

protos.cpp needs intproto.h for MAX_NUM_PROTOS.
commontraining.h needs to always use baseapi.h, as otherwise most training binaries won't build, and it doesn't rely on the HAVE_CONFIG_H it's ifdef-ed around.


zdenop:
thank. Welcome back. Can you have a look at PR #310?

PR #2146, opened 2019-01-05
Updates File::DeleteMatchingFiles

rhardih:
This adds and ANDROID specific version of DeleteMatchingFiles, without
the use of glob.h, in order to be backwards compatible with Android 6.0.

cc @theraysmith @zdenop

### DIFF
--- a/src/ccutil/fileio.cpp
+++ b/src/ccutil/fileio.cpp
@@ -19,7 +19,15 @@
 #include <io.h>
 #endif
 #else
+
+// glob.h isn't available in the Android NDK until platform version android-28
+#ifdef __ANDROID__
+#include <dirent.h>
+#include <regex.h>
+#else
 #include <glob.h>
+#endif
+
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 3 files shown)


rhardih:
I've only compile tested this, since I'm not really familiar with the training part of tesseract.

Do I need to write a test for `DeleteMatchingFiles`? I couldn't seem to find an existing one.

src/ccutil/fileio.cpp
```diff
@@ -113,16 +122,52 @@ bool File::DeleteMatchingFiles(const char* pattern) {
...
+
+  regcomp(&match, pattern, REG_BASIC);
+  dir = opendir(directory);
+
+  if (dir != NULL) {
+    while ((dp = readdir(dir)) != NULL) {
+      if (!(res = regexec(&match, dp->d_name, 0, NULL, 0))) {
+        full_path = File::JoinPath(directory, dp->d_name).c_str();
+        all_deleted &= File::Delete(full_path);
+      }
```
rhardih: Are the two cases below needed, or can we do without error messaging for this?

rhardih:
The approach was inspired by [this commit](https://github.com/tesseract-ocr/tesseract/commit/9b2906c67ea90aa4adb46f314796d53cfd6c7325) getting rid of *glob.h* as well.

src/ccutil/fileio.cpp
```diff
@@ -113,16 +122,52 @@ bool File::DeleteMatchingFiles(const char* pattern) {
...
+      else if (res == REG_NOMATCH) {
+        // No match, noop or warn?
+      }
+      else {
+        // Print error message?
+
+        //regerror(res, &match, msg, sizeof(msg));
+        //fprintf(stderr, "Regex match failed: %s\n", msg);
+        //exit(1);
```
rhardih: In case these are kept, what's the way to go. I'm assuming exiting here isn't the desired outcome.

zdenop:
I don't think training was intended for Android...

rhardih:
So maybe the solution is to exclude training from Android builds entirely?

zdenop:
IMO yes - training is focused on desktop used, so for mobile solution is does not make sense...

rhardih:
Ok, I guess this might belong behind the BUILD_TRAINING_TOOLS flag. I'll see if I can whip up another PR for that.

PR #2152, opened 2019-01-09
 Remove opencl_device_selection.h

stweil:
Always use OpenCL device selection if OpenCL is enabled.

This fixes a regression which was introduced by commit 5c6a57b
which removed the definition for USE_DEVICE_SELECTION.


stweil:
It looks like the CI checks are currently broken.

PR #2153, opened 2019-01-09
Fix documentation for page segmentation mode 2

stweil:
It never worked, so add a comment that the implementation is missing.
Add also a to-do comment.


amitdo:
What's the logic to say 'psm 2 - .not implemented', but keep the wrong code?

amitdo:
Even if we ignore the mismatch, AFAICT the code itself does not use the api correctly.

amitdo:
... and even if it was correct, it's not in the right place.

Any psm output should be implemented as 'renderer'.

stweil:
Yes.

PR #2160, opened 2019-01-15
Refactor IntSimMatrix and WeightMatrix

stweil:
* Move shaped weights from IntSimMatrix to WeightMatrix .
* Replace dynamically allocated leaked lazily initialized singleton IntSimMatrix by constant objects.
* Compile files for AVX, AVX2 or SSE only when needed.
* Allow user selected method for calculation of integer dot product (for example `-c dotproduct=sse`).
* Several small optimisation (slightly reduced code size, small performance gain).


stweil:
Builds with CMake are currently broken (unrelated to this pull request).

PR #2162, opened 2019-01-16
Fix broken builds

stweil:
Don't use CPPAN by default because it fails with an error message:

    /usr/bin/ld: cannot find -lpvt.cppan.demo.danbloomberg.leptonica


stweil:
Only CMake based builds were broken. The continuous integration tests (Appveyor, Travis) use CMake, so they were broken, too.

PR #2167, opened 2019-01-18
Fixes unicharset_test

Shreeshrii:
unichar_test still getting error

unichar_test.cc: In member function ‘virtual void {anonymous}::UnicharTest_Conversion_Test::TestBody()’:
unichar_test.cc:24:21: error: ‘char32’ was not declared in this scope
   const std::vector<char32> kUTF32Src = {'a', 0x5d0, 0xca4, 0x2a714};
                     ^~~~~~
unichar_test.cc:24:21: note: suggested alternative:
In file included from unichar_test.cc:12:0:
../src/ccutil/unichar.h:52:26: note:   ‘tesseract::char32’
 using char32 = signed int;
                          ^

### DIFF
--- a/unittest/unicharset_test.cc
+++ b/unittest/unicharset_test.cc
@@ -1,4 +1,19 @@
-#include "tesseract/ccutil/unicharset.h"
+// (C) Copyright 2017, Google Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


unittest/unicharset_test.cc
```diff
@@ -62,7 +77,7 @@ TEST(UnicharsetTest, Multibyte) {
   EXPECT_EQ(u.size(), 9);
   EXPECT_EQ(u.unichar_to_id("\u0627"), 3);
   EXPECT_EQ(u.unichar_to_id("\u062c"), 4);
-  // The first two bytes of this string is \u0627, which matches id 3;
+  // The first two bytes of this std::string is \u0627, which matches id 3;
```
stweil: That change was not needed.

stweil:
This needs #2168 to fix the build errors.

unittest/unicharset_test.cc
Shreeshrii: You are right. I had done a replace for ' string ' by ' std::string '.

PR #2175, opened 2019-01-21
Add header files for dawg_test

Shreeshrii:
Files taken from Google's ortools repo 

@stweil Please review whether they are needed.


stweil:
I was not successful in finding that code and used a different implementation therefore, so they are no longer needed.

PR #2186, opened 2019-01-23
 training: Don't terminate after processing 8 fonts or 8 images

stweil:
tesstrain_utils.sh sets the shell flag -e, so it exits immediately
if a command exits with a non-zero status.

The following command returns a non-zero status as soon as counter is a
multiple of par_factor (par_factor=8, that means as soon as 8 fonts or
images are processed):

    let rem=counter%par_factor

The new code fixes this undesired exit.


Shreeshrii:
Thanks for fixing the bug.

PR #2188, opened 2019-01-23
tesstrain.sh was failing with line 81: ENABLE_SHAPE_CLUSTERING: unbound variable

diegodlh:
It seems that to decide whether to run phase S or not, tesstrain.sh was checking if ENABLE_SHAPE_CLUSTERING variable was set to "y". However, this variable is unbound, and the script fails when it reaches this point.
It looks like whether to run shape clustering or not is decided with the --run_shape_clustering flag, which is checked in the tesstrain_utils.sh script instead.
Therefore, it seemed to me that removing this if statement from tesstrain.sh seemed appropriate to solve this bug.


stweil:
There is also a similar Python3 script for training. All fixes made for the Bash script might be needed for Python3, too. This can be done later in a separate pull request, of course.

PR #2200, opened 2019-01-26
fix and enable more unittests

Shreeshrii:


unittest/lang_model_test.cc
```diff
@@ -111,4 +214,4 @@ TEST(LangModelTest, AddACharacter) {
 }
 
 }  // namespace
-}  // namespace tesseract
+}  // namespace tesseract
```
stweil: Here you accidentally removed the terminating linefeed.

Shreeshrii: Fixed via 5e90a61eb4e27d00d70a8385369bd969dd786df4
Thanks!

src/training/tesstrain_utils.sh
```diff
@@ -246,7 +250,7 @@ generate_font_image() {
 
     local common_args="--fontconfig_tmpdir=${FONT_CONFIG_CACHE}"
     common_args+=" --fonts_dir=${FONTS_DIR} --strip_unrenderable_words"
-    common_args+=" --leading=${LEADING}"
+    common_args+=" --leading=${LEADING} --xsize=${X_SIZE}"
```
stweil: The argument is called `--xsize` here and also in the subject of this commit, but the commit implements it as `--x_size`. I think this is confusing and suggest to always use `--xsize`.

Shreeshrii: Thanks. Fixed via 94b8988b8cca3812137933db00750bd6e2e84e32

PR #2209, opened 2019-01-31
Use universal location for docbook.xsl

cjmayo:
xsltproc will use the system catalog to find the local path.
Pass --nonet option to ensure the Internet is not used.

---

docbook.xsl isn't in the same location on all distributions e.g. on Fedora (symlink) and Gentoo Linux is at:
/usr/share/sgml/docbook/xsl-stylesheets/manpages/docbook.xsl


stweil:
Thanks, it's merged now.

PR #2211, opened 2019-02-01
configure: Remove header check for ICU

stweil:
It wrongly detects old versions of ICU as valid.
Checking with pkg-config is sufficient and also sets ICU_UC_LIBS.


zdenop:
thanks

PR #2212, opened 2019-02-01
Fix build rule for manpages

stweil:
This is similar to commit 2106cba0a98a90451df835f3ab7b2aaf54826442
which fixed doc/generate_manpages.sh.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -2,9 +2,6 @@
 
 if ASCIIDOC
 
-asciidoc=asciidoc -d manpage
-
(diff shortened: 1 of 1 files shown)


stweil:
@AlexanderP, could you please test whether this fix works for the build of the Debian package, so you can remove your local patch?

stweil:
Building of the documentation requires the Debian packages `asciidoc`, `xsltproc` and `docbook-xsl`. autoconf currently only checks for `asciidoc`.

PR #2216, opened 2019-02-02
Lstmbox

Shreeshrii:
Create box files (using code similar to tsv renderer) in the format needed for LSTM training i.e. with a line for space after every word and a line with tab to mark end of line.

### DIFF
--- a/src/api/renderer.h
+++ b/src/api/renderer.h
@@ -252,7 +252,7 @@ class TESS_API TessUnlvRenderer : public TessResultRenderer {
  */
 class TESS_API TessLSTMBOXRenderer : public TessResultRenderer {
  public:
-  explicit TessLSTMBOXRenderer(const char *outputbase);
+  explicit TessLSTMBOXRenderer(const char* outputbase);
 
  protected:
   virtual bool AddImageHandler(TessBaseAPI* api);
--- a/src/api/renderer.h
+++ b/src/api/renderer.h
@@ -248,6 +248,17 @@ class TESS_API TessUnlvRenderer : public TessResultRenderer {
 };
 
 /**
+ * Renders tesseract output into a plain UTF-8 text string for LSTMBOX
+ */
+class TESS_API TessLSTMBOXRenderer : public TessResultRenderer {
+ public:
+  explicit TessLSTMBOXRenderer(const char *outputbase);
+
+ protected:
+  virtual bool AddImageHandler(TessBaseAPI* api);
+};
+
+/**
  * Renders tesseract output into a plain UTF-8 text string
  */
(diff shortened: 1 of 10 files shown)


Shreeshrii:
I have tested this for `eng` and `hin`. It will require testing for RTL languages.

Since the character level bounding boxes are NOT accurate with LSTM engine,  I changed the format to the one used by ocrd-train i.e. bounding info at TEXTLINE level for all characters on a line.

zdenop:
Sorry if I miss some information, but what is user case for this?

Shreeshrii:
When tesseract  is used with `makebox` to create box files from scanned images, the format is suitable for training for base tesseract (--oem 0) but does not have the boxes for space between words and tabs to mark end of line as needed for LSTM training.

This PR allows creation of box files from images, in the format needed by LSTM training. The box files will still need to be edited for accuracy (similar to tesseract 3). 

This can be useful when someone wants to finetune for a particular typeface which is not available as a font.

Shreeshrii:
Since there are lots of requests by users who want to train using images, I wanted to add this option. 
The code in src/api/lstmboxrenderer.cpp can be streamlined. I do not know c++ to be able to do that. 
@stweil Please review and fix. Thanks.

src/api/renderer.h
```diff
@@ -247,6 +247,17 @@ class TESS_API TessUnlvRenderer : public TessResultRenderer {
   virtual bool AddImageHandler(TessBaseAPI* api);
 };
 
+/**
+ * Renders tesseract output into a plain UTF-8 text string for LSTMBOX
+ */
+class TESS_API TessLSTMBOXRenderer : public TessResultRenderer {
+ public:
+  explicit TessLSTMBOXRenderer(const char *outputbase);
```
stweil:
```suggestion
  explicit TessLSTMBOXRenderer(const char* outputbase);
```

Shreeshrii:
[testbox.zip](https://github.com/tesseract-ocr/tesseract/files/2825674/testbox.zip)

Test results for eng, hin, ara, chi_sim and chi_tra

Both chi_sim and chi_tra are including extra spaces when the word has both Chinese and Latin script.

Shreeshrii:
https://github.com/Shreeshrii/tesseract/tree/lstmboxnew has a cleaner version of commits.

amitdo:
text2image does not output space char at EOL.

Shreeshrii:
@amitdo Thanks. Will change.

Have you checked the output for RTL eg. Hebrew?

amitdo:
The ara lstmbox in testbox.zip looks fine.

PR #2220, opened 2019-02-03
Man page description of configs and parameters

cjmayo:
Three commits here, first two should be straightforward:
* get.image is mentioned in tesseract(1), as get.images and not installed
* A typo and missing full stops in tesseract(1)

At first read of tesseract(1) I didn't find it easy to follow what a config was and the third commit is my attempt to improve that. That's more subjective so please do look at it critically.

### DIFF
--- a/doc/tesseract.1.asc
+++ b/doc/tesseract.1.asc
@@ -36,7 +36,7 @@ IN/OUT ARGUMENTS
 	The basename of the output file (to which the appropriate extension
 	will be appended).  By default the output will be a text file
 	with `.txt` added to the basename unless there are one or more
-	'configfile' options which explicitly specify the desired output.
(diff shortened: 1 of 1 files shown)


stweil:
It would be easier to have separate pull requests instead of one with several unrelated commits. The 2nd commit is fine for me, but the others are more controversial or need more care. I wonder for example whether renaming the current file `tessdata/configs/get.image` to `tessdata/configs/get.images` would be a better solution.

cjmayo:
OK. Created:

Rename get.image config to get.images and install #2223
Tidy tesseract(1) #2224

I'll rebase this one if and when the others are merged because there is a slight overlap.

cjmayo:
Rebased on master, including restoring reference to `get.images`.

PR #2223, opened 2019-02-05
Rename get.image config to get.images and install

cjmayo:
get.images is mentioned in tesseract(1) and not installed.


stweil:
Thank you.

PR #2227, opened 2019-02-08
Avoid gmake-specific pattern substitution in Makefile.am.

### LINKED ISSUE #2226: POSIX make compatibility
The rules in doc/Makefile.am are not compatible with BSD/POSIX make:

```
make: don't know how to make combine_lang_model.1 (prerequisite of: all-am)
Stop in /tmp/tesseract/doc
```

### Environment

* **Tesseract Version**: 4.0.0
* **Commit Number**: 
* **Platform**: OpenBSD

rrrapha:
Resolves #2226

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -32,13 +32,15 @@ EXTRA_DIST = $(man_MANS) Doxyfile
 
 .PHONY: html
 
-html: $(patsubst %,%.html,$(man_MANS))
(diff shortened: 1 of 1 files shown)


stweil:
I am not sure whether these changes are really needed. Personally I think that the current code is easier to understand. Are there environments used for building Tesseract which cannot provide `gmake`?

PR #2228, opened 2019-02-09
OpenCL: Remove PERF_COUNT framework

stweil:
It was rarely used, but added a lot of code and an unconditional
dependency on openclwrapper.h.


amitdo:
Why we still keep the OpenCL code (~4000 LOC) ?

stweil:
Maybe because I am just evaluating it with even more lines of code for LSTM?

amitdo:
Ok.

amitdo:
Did you hear about SYCL?
https://en.wikipedia.org/wiki/SYCL

stweil:
No, I did not up to now. Thank you for the hint.

PR #2235, opened 2019-02-12
Simplify tanh and logistic functions and precompute function tables

stweil:
Both functions are called very often, so computing the table values
at program start should be faster than computing them on demand.

### DIFF
--- a/src/lstm/functions.cpp
+++ b/src/lstm/functions.cpp
@@ -2,7 +2,6 @@
 // File:        functions.cpp
 // Description: Static initialize-on-first-use non-linearity functions.
 // Author:      Ray Smith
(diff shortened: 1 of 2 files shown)


stweil:
C++ 11 cannot compute the tables at compile time. C++ 14 would support that additional optimization. Should we consider switching to C++ 14?

zdenop:
I would prefer conditional build ([`#if __cplusplus < 201300`](https://stackoverflow.com/questions/30940504/simplest-c-example-to-verify-my-compiler-is-c14-compliant)) for some time as we did it for c++11...

amitdo:
https://github.com/google/styleguide/issues/429

Issue #2225, opened 2019-02-06
Compilling Tesseract-ocr using git or other compilation


alucard079:
After the command "make"

this error has been generated.

IOError: [Errno 32] Broken pipe
Makefile:616: recipe for target 'combine_lang_model.1' failed
make[2]: *** [combine_lang_model.1] Error 127
make[2]: Leaving directory '/home/nigel/tesseract/doc'
Makefile:495: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/nigel/tesseract'
Makefile:404: recipe for target 'all' failed
make: *** [all] Error 2

What should I do? I followed the instructions very carefully.

stweil:
That might be caused by missing required packages. See https://github.com/tesseract-ocr/tesseract/pull/2212#issuecomment-459834395 for the Debian packages which are needed to build `combine_lang_model.1`. Could you please update the Wiki if that works for you?

zdenop:
@alucard079:  did you check the latest code?

zdenop:
Closed because of missing input from reporter.

PR #2248, opened 2019-02-15
Improve robustness of manpage build

stweil:
* Don't try the build if xsltproc is missing.
* Don't abort the build if the stylesheet file is missing.


stweil:
This addresses issue #2225.

PR #2256, opened 2019-02-19
RFC: tess4train_impact_from_full.sh

Shreeshrii:


Shreeshrii:
Request Feedback:

1. Does it run successfully in your setup? You will need to provide box/tiff pair directory and change other paths if needed.

2. It is a simple script which needs to be manually modified for running and does not do any error checking.

3. Does it need to be put under src/training?

4. Are there any formatting conventions to be followed for bash? 

I am planning to make a similar one for plus-minus training. I think these two options will suffice for most users.

Shreeshrii:
See newer bash scripts in https://github.com/Shreeshrii/tess4training

See updated ocrd-train Makefile in https://github.com/OCR-D/ocrd-train/pull/66

PR #2266, opened 2019-02-25
DO NOT MERGE: Fix reversal of numerals in Arabic script

Shreeshrii:
Fixes issue https://github.com/tesseract-ocr/tesseract/issues/2263

Also addresses related issues
Add Indic numerals and missing punctuation to Arabic tesseract-ocr/langdata#131
Arabic training data has room for improvement #2047

See https://github.com/Shreeshrii/tessdata_arabic
for the finetuned traineddata file and test image and OCR output.

The OCR output looks correct when seen in notepad++ in RTL view. However I am not able to copy it in github.

PR #2270, opened 2019-02-26
Treat U_ARABIC_NUMBER as LTR

Shreeshrii:
https://www.w3.org/International/articles/inline-bidi-markup/uba-basics#numbers

> Numbers in RTL scripts run left-to-right within the right-to-left flow.

See earlier discussion in the closed PR - https://github.com/tesseract-ocr/tesseract/pull/2266

The change to ccutil/unicharset.cpp which was in earlier PR is NOT included, since U_ARABIC_NUMBER was being used there as part of condition to find if right_to_left scripts are significant in the unicharset.

https://github.com/tesseract-ocr/tesseract/blob/master/src/ccutil/unicharset.cpp#L956

### DIFF
--- a/src/ccstruct/pageres.h
+++ b/src/ccstruct/pageres.h
@@ -398,8 +398,7 @@ class WERD_RES : public ELIST_LINK {
       UNICHARSET::Direction dir =
           uch_set->get_direction(unichar_id);
       if (dir == UNICHARSET::U_RIGHT_TO_LEFT ||
-          dir == UNICHARSET::U_RIGHT_TO_LEFT_ARABIC ||
-          dir == UNICHARSET::U_ARABIC_NUMBER)
+          dir == UNICHARSET::U_RIGHT_TO_LEFT_ARABIC)
         return true;
     }
     return false;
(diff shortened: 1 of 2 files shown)


Shreeshrii:
@zdenop  Please merge this. 

Test Case:

Image: https://github.com/Shreeshrii/tessdata_arabic/blob/master/ara.Amiri.exp0-ara-Amiri-layer.png

Ground Truth: https://github.com/Shreeshrii/tessdata_arabic/blob/master/ara.Amiri.exp0-ara-Amiri-layer.testdeco.gt.txt

Traineddata (Replace a layer training): https://github.com/Shreeshrii/tessdata_arabic/blob/master/ara-Amiri-layer.traineddata

OCRed text - psm 6: https://github.com/Shreeshrii/tessdata_arabic/blob/master/ara.Amiri.exp0-ara-Amiri-layer.txt

According to @amitdo the text files display correctly in RTL format in Firefox.

I have tested with numbers with numerals in Arabic script at beginning, end and middle of lines. As seen in earlier results TOC with numbers at end was displaying correctly. This sample shows correct recognition and no reversal at end and middle. 

Where a line starts with a number, some digits are being dropped - this was something I had noticed in a different issue with Hebrew example also. I will post a link later.

Shreeshrii:
FYI - see https://github.com/tesseract-ocr/tesseract/pull/2266 for earlier version of PR and related discussion.

PR #2272, opened 2019-02-26
Allow UTF-8 variant of C locale

nijel:
It behaves same in scanf, but it allows proper handling of unicode
chars.


stweil:
Thank you, the commit is merged now. It is still not the final solution, but relaxes the current restriction a little bit.

nijel:
I hope this is not a final solution, but with this patch it is at least possible to work with unicode filenames on Python 3.6 while using tesseract...

PR #2284, opened 2019-03-02
Fix some compiler warnings

stweil:


src/training/boxchar.cpp
```diff
@@ -54,7 +53,7 @@ void BoxChar::GetDirection(int* num_rtl, int* num_ltr) const {
   std::vector<char32> uni_vector = UNICHAR::UTF8ToUTF32(ch_.c_str());
   if (uni_vector.empty()) {
     tprintf("Illegal utf8 in boxchar string:%s = ", ch_.c_str());
-    for (int c = 0; c < ch_.size(); ++c) {
+    for (unsigned c = 0; c < ch_.size(); ++c) {
```
egorpugin: Isn't size_t is better (on 64 arch, c will be 32)?

src/training/validate_indic.cpp
```diff
@@ -120,7 +120,7 @@ bool ValidateIndic::ConsumeViramaIfValid(IndicPair joiner, bool post_matra) {
         ASSERT_HOST(!CodeOnlyToOutput());
       } else {
         // Half-form with optional Nukta.
-        int len = output_.size() + 1 - output_used_;
+        unsigned len = output_.size() + 1 - output_used_;
```
egorpugin: size_t?

egorpugin:
Maybe size_t is better in several cases in this PR?

stweil:
Using `size_t` in `boxchar.cpp` is indeed reasonable, because that code already had similar code with `size_t`. I updated the pull request accordingly.

I did not update the validator code to use `size_t` now although that would be possible because a consistent use of `size_t` there would require more changes.

PR #2231, opened 2019-02-10
Add renderer to create WordStr box files from images

Shreeshrii:
@thebigwasp provided an example of WordStr box file that works in https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/tesseract-ocr/0r8QvV3j8ew/5oYrCY5_AAAJ

This PR implements it.

PR #2286, opened 2019-03-02
Rename function to TessBaseAPIGetTsvText to be consistent to Create method

Shreeshrii:


zdenop:
Why? [TSV](https://en.wikipedia.org/wiki/Tab-separated_values) is equivalent for CSV. Also C++ method use TSV.  I never see Csv or Tsv. Search for TSV file...

Shreeshrii:
Tsv is being used within tesseract already.

https://github.com/tesseract-ocr/tesseract/blob/master/src/api/renderer.cpp#L145

```
TessTsvRenderer::TessTsvRenderer(const char* outputbase)
    : TessResultRenderer(outputbase, "tsv") {
  font_info_ = false;
}

TessTsvRenderer::TessTsvRenderer(const char* outputbase, bool font_info)
    : TessResultRenderer(outputbase, "tsv") {
  font_info_ = font_info;
}

bool TessTsvRenderer::BeginDocumentHandler() {
  // Output TSV column headings
  AppendString(
      "level\tpage_num\tblock_num\tpar_num\tline_num\tword_"
      "num\tleft\ttop\twidth\theight\tconf\ttext\n");
  return true;
}

bool TessTsvRenderer::EndDocumentHandler() { return true; }

bool TessTsvRenderer::AddImageHandler(TessBaseAPI* api) {
  const std::unique_ptr<const char[]> tsv(api->GetTSVText(imagenum()));
  if (tsv == nullptr) return false;

  AppendString(tsv.get());

  return true;
}```

zdenop:
it is ok to have variable lower case (tsv). IMO `Tsv` should be corrected to `TSV`.

amitdo:
Correcting the name in the C++ code will break the API.

Shreeshrii:
My change was only for my earlier commit for C-API.

amitdo:
Yes Shree, I know.

My comment was directed to Zdenko.

zdenop:
@amitdo: We already [modified API](https://abi-laboratory.pro/index.php?view=objects_report&l=tesseract&v1=4.0.0&v2=current)
I really wonder if anybody use renderer part of API  (e.g. storing output to disk) ;-) for simple formats like TSV or txt  if they have access to ResultIterator...

zdenop:
@stwei : What is your opinion here?

stweil:
If we think that the API is not used for TSV, then removing those API functions would be a simple solution.

Regarding upper and lower case: it's already a mess, as there is also `TessAlto...`, `TessHOcr...`, `TessUnlv...`.

Shreeshrii:
I think there are developers who would use it when available. Please see https://github.com/tesseract-ocr/tesseract/pull/2231#issuecomment-468855255

PR #2287, opened 2019-03-02
Format new code with clang-format

stweil:

### DIFF
--- a/src/api/wordstrboxrenderer.cpp
+++ b/src/api/wordstrboxrenderer.cpp
@@ -23,9 +23,9 @@
 namespace tesseract {
 
 /**
- * Create a UTF8 box file with WordStr strings from the internal data structures.
- * page_number is a 0-base page index that will appear in the box file.
- * Returned string must be freed with the delete [] operator.
+ * Create a UTF8 box file with WordStr strings from the internal data
+ * structures. page_number is a 0-base page index that will appear in the box
+ * file. Returned string must be freed with the delete [] operator.
  */
 
 char* TessBaseAPI::GetWordStrBoxText(int page_number) {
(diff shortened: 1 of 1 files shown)


src/api/wordstrboxrenderer.cpp
```diff
@@ -44,31 +44,32 @@ char* TessBaseAPI::GetWordStrBoxText(int page_number) {
...
-        res_it->Next(RIL_WORD);
-     } while (!res_it->Empty(RIL_BLOCK) && !res_it->IsAtBeginningOf(RIL_WORD));
+    do {
+      wordstr_box_str +=
+          std::unique_ptr<const char[]>(res_it->GetUTF8Text(RIL_WORD)).get();
+      wordstr_box_str += " ";
+      res_it->Next(RIL_WORD);
+    } while (!res_it->Empty(RIL_BLOCK) && !res_it->IsAtBeginningOf(RIL_WORD));
   }
   wordstr_box_str.add_str_int("\n\t ", right + 1);
```
stweil: @Shreeshrii, `right`, `top` and `bottom` can be uninitialized here (reported by Coverity Scan). What would be the right fix for this?

amitdo:
>Format new code with clang-format

What about old code?

stweil:
`src/api/wordstrboxrenderer.cpp` was added recently, so there is no old code.

amitdo:
I meant what about other files?

src/api/wordstrboxrenderer.cpp
stweil: There is a similar problem in [lstmboxrenderer.cpp:82](https://github.com/tesseract-ocr/tesseract/blob/master/src/api/lstmboxrenderer.cpp#L82) and 3 more lines.

stweil:
The new code got my attention because of the [latest report from Coverity Scan](https://scan.coverity.com/projects/tesseract-ocr?tab=overview).

But you are right, the old code needs formatting, too.

zdenop:
@stweil: do you plan to fix issues from Coverity Scan?

stweil:
Not at the moment. I am still waiting for feedback from @Shreeshrii (see question above).

src/api/wordstrboxrenderer.cpp
Shreeshrii: @stweil For some reason I did not get any email regarding the comments on the commits. Got the msg re issue today.

The EOL tab needs to be added at end of a line. Currently it is identified via IsAtBeginningOf(RIL_TEXTLINE). However that would put an EOL tab box at the beginning of the page,  hence I am checking that it is NOT the first line on page.

There is probably a different way to structure the code to get rid of this problem.  Please change the code as you see fit. Thanks!

Shreeshrii:
@stweil I hadn't recd the earlier msg with the review. Please feel free to change the code. Thanks.

PR #2289, opened 2019-03-04
Added an additional optional --tmp_dir parameter to specify the tempo…

Armyke:
…rary directory in which tesstrain.py creates the training temporary files. The main reason is due to the slow R/W on HDD, if anyone wants to speed up this process can use as tmp_dir a directory on an SSDrive

### DIFF
--- a/src/training/tesstrain_utils.py
+++ b/src/training/tesstrain_utils.py
@@ -125,6 +125,7 @@ parser.add_argument(
     help="A list of fontnames to train on.",
 )
 parser.add_argument("--fonts_dir", help="Path to font files.")
+parser.add_argument("--tmp_dir", help="Path to temporary training directory.")
 parser.add_argument(
     "--lang", metavar="LANG_CODE", dest="lang_code", help="ISO 639 code."
 )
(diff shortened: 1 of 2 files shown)


zdenop:
tesstrain_utils.py should be "synced" with its bash version... 
1. First changes from tesstrain.sh and tesstrain_utils.sh should to implemented to its python version
2. And there should be also patch for tesstrain_utils.sh...

@jbarlow83: can you have a look on this so we can implement it in 4.1.0 version

Armyke:
I was currently working on that, the latest commit of about 10 mins ago is
on tesstrain_utile.sh, i tested it and i hope it's fine. Have a good day!

Il Lun 4 Mar 2019, 15:14 zdenop <notifications@github.com> ha scritto:

jbarlow83:
The Python and IIRC the bash version too, both use mktemp from libc and so
will place temporary files in the path specified by the TMPDIR environment
variable. As such there is no point in implementing this argument for
either script - it duplicates standard functionality.

zdenop:
thanks.

Issue #1270, opened 2018-01-12
Sans-serif 'I' is recognized as vertical bar, '|'


araftis:
### Environment

Version:
tesseract 4.00.00alpha
 leptonica-1.74.4
  libgif 4.2.3 : libjpeg 9b : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0

 Found AVX2
 Found AVX
 Found SSE

Commit Hash:
10a8a67ca2bde7f2fad2370ed7c3feda7400084e

Platform:
Mac OS X, High Sierra (10.13.4)
Note that support libraries were built with Mac Ports, but I pulled and built the source from scratch.

### Current Behavior:

So I tried out the new 4.0 branch, and I must say, I'm quite impressed with the improved accuracy. I'm using Tesseract to OCR subtitles form a MKV / BluRay subtitle stream, and for the most part, the accuracy was really high, except in one case, the capital "I", when it's represented in a sans-serif font. Pretty much every single time, the recognizer saw this as a vertical bar character "|" . Note that it worked correctly when the "I" was combined with other test, like "I'll". It only failed when "I" appeared alone.

### Expected Behavior:

Given that I would expect vertical bars to rarely appear in text, especially compared to the first person pronoun, I expect to get the letter "I". 

### Additional Notes:

I've attached one of the subtitle frames. Note that this is written from my test app while recognizing the images, but running tesseract on the command line gives the same results. Also of note, this particular image is an oblique font, but I have the same issue with non-oblique fonts.

![kick-ass 2 0022](https://user-images.githubusercontent.com/35355017/34856671-444b0260-f6fb-11e7-899e-5150ab8e074a.png)

Shreeshrii:
You can do post-processing after OCR to change vertical bar | to capital I.

Or, you can do plusminus type of fine-tuning on English traineddata to not include vertical bar | in the unicharset.

araftis:
Oh, I certainly do post processing. In fact, I used to do a bunch of post processing with the 3.x branch, so I was quite pleased I only had to do one substitution on the 4.0 code. I'll take a look at the traineddata, but since it's not my area of expertise, I figured it was worth it for someone with more knowledge should take a look.

Issue #960, opened 2017-05-30
user pattern/dict does not work at all


wosiu:
They do not work for me. I've been trying versions: 3.05.00 and 4.00.00alpha.
My file date.user-pattern contains one line:
2014-\d\d-\d\d
Picture is one line with date, like: 2014-03-19
I run: tesseract img.jpg stdout --user-patterns date.user-patterns -psm 8
and output: "mum-w" which obviously does not match the pattern. 
Character whitelisting helps a bit, but format from pattern is not preserve and accuracy is poor.
I also tried some other examples - does not work either.
Many people have the same problem, aggregated links under this one:
https://stackoverflow.com/questions/34560697/tesseract-ocr-user-patterns
also #403
Should we assume that this feature does not work at all? Is there any official comment on this?

wosiu:
Same problem with user dictionary:
tesseract H3.png stdout --user-patterns date.user-patterns --psm 8 --user-words date.user-words -c language_model_penalty_non_dict_word=9999999999999999999 --oem 0
I tried different language_model_penalty_non_dict_word values with no luck
Related #297, which is closed, so I assume the feature doesn't work. I think it would be better for users if those flags would be removed from command line and configurations, because it is misleading as long as they don't affect engine.

Shreeshrii:
Tested user-words option with 3.05.01 on windows (using binaries by @stweil)

Works ok. See attached test image.

bazaar config file as used (uses system dictionary + user words)
```
load_system_dawg     T
load_freq_dawg       T
user_words_suffix    user-words
user_patterns_suffix user-patterns
```

eng.user-words as used
```
Online
the
quick
brown
fox
jumped
```

Output without user-words
```
tesseract test.png stdout
shared Guruvayoor Dnline Friends's post
```

Output with user-words
```
tesseract test.png stdout  bazaar
shared Guruvayoor Online Friends's post
```

So `Online` from eng.user-words is used to improve output when using the `bazaar` config file.

Shreeshrii:
I tested for user-patterns just now with versions 3.02 and 3.05.01, both for windows so that I didn't have to worry about correct versions of leptonica. The test image is attached.

There is no change in output with the user-patterns option in both. So, if this feature worked, it would be before 3.02. 

However, just by resizing the image to 200%, the dates  are correctly recognized. 

![date](https://cloud.githubusercontent.com/assets/5095331/26753470/976e8802-4884-11e7-82aa-5cb4bc1b75e5.png)
[...]

Issue #2159, opened 2019-01-15
Option --psm 10 digits are not taken account.


ruimo:
### Environment

* **4.0.0**
* **macOS 10.13.3**

### Current Behavior:

$ tesseract 000.png stdout --psm 10 digits
Warning: Invalid resolution 0 dpi. Using 70 instead.
iL

000.png: ![000](https://user-images.githubusercontent.com/1510145/51164234-9ea68e80-18df-11e9-844c-ea5bcad93165.png)

### Expected Behavior:

Since --psm 10 is specified, the output 'iL' is invalid. A single character should be recognized.
As 'digits' are specified as a config file, 'iL' is invalid. A digit character should be recognized.

amitdo:
The content of the `digits` config file:
`tessedit_char_whitelist 0123456789-.`

See #751.

ruimo:
Thanks, I updated the expected behaviour.

bertsky:
I have looked into this via debugging (adding `DebugBeams` at the end of `RecodeBeamSearch::Decode`). For the whitelisting, I used branch #2294. 

> Since --psm 10 is specified, the output 'iL' is invalid. A single character should be recognized.

True. From what I can tell, that `PSM_SINGLE_CHAR` does not work for LSTMs. And getting two characters out of one is a [known error](https://github.com/tesseract-ocr/tesseract/issues/1465). I coined this the "diplopia" problem when presenting my analysis for #2339.

> As 'digits' are specified as a config file, 'iL' is invalid. A digit character(0123456789-.) should be recognized.

True again. As @amitdo and @Shreeshrii pointed out, whitelisting will be re-instated for LSTMs after #2294 is merged. But even with that, for this example, nothing can be decoded (empty result), because the **beam is too narrow**. 

It can be made wider, e.g. by changing `ComputeTopN` to not stop looking further than the first and second best input alternatives (which will be null and "." in this case). I do get a "1" as result with that modification, but the implications of this should be discussed in #2294.

Shreeshrii:
@ruimo 

Please try with my finetuned traineddata which is limited to digits and some punctuation (instead of digits config).

digits_layer.traineddata	
digitsall_layer.traineddata

from https://github.com/Shreeshrii/tessdata_shreetest

Issue #403, opened 2016-08-30
LSTM: User patterns do not work


Shreeshrii:
ref: https://groups.google.com/forum/#!msg/tesseract-ocr/S9CIK3jOMWw/vVBZULrJ9xcJ

I tried using bazaar config for user patterns suggested in above post ( \A\A\d\d\d\A\A
) with the latest windows binary. It does not seem to work. Does the functionality work on linux?

input, output and config files attached. I added.txt extension to bazaar and eng.user-patterns in order to upload it here.

![patterntest](https://cloud.githubusercontent.com/assets/5095331/18089323/013e7fa8-6edd-11e6-948a-65b376c7ec4f.png)

OUTPUT

```
0011917
OX345PT
PT7895M
BA409QT
OMOOKM
WE4321M

OOLI9T7
OX345PT
PT789SM
BA409QT
OMOOKMI
WE432LM

OOLI9T7
OX345PT
PT7898M
BA409QT
OMOOKMI
WE432LM


```

[patternbazaar.txt](https://github.com/tesseract-ocr/tesseract/files/444884/patternbazaar.txt)

[bazaar.txt](https://github.com/tesseract-ocr/tesseract/files/444889/bazaar.txt)
[eng.user-patterns.txt](https://github.com/tesseract-ocr/tesseract/files/444890/eng.user-patterns.txt)

Shreeshrii:
Some other reports of user-patterns and user-words not working

https://groups.google.com/forum/#!topic/tesseract-ocr/5vFqVcJmHnM

http://stackoverflow.com/questions/17209919/tesseract-user-patterns

Has anyone tried this? Does it work?

theraysmith:
Question:
There are 2 ways these things could work:
1. FORCE the output to match the provided pattern(s) and/or word(s). With this option, you **can't get anything else out**, whatever is in the image.
2. Use the user-patterns and user-words as a hint. Other things could be output, if it thinks it is more likely. The hint can be made stronger, but **there will always be inputs that produce something outside of the patterns supplied.**
Which is it to be?
Can someone familiar with the above discussions please summarize for me, and if the consensus is 1 above, then it could be made to happen, or else it might be possible to increase the strength of the hint.

BTW, this could behave differently for base tesseract vs LSTM.

amitdo:
I can tell you that in the Tesseract forum many users ask about these files. They are disappointed that there is no effect on accuracy when using them with their input.

[...]

PR #2294, opened 2019-03-07
trying to add tessedit_char_whitelist etc. again:

bertsky:
- ignore matrix outputs in ComputeTopN if they
  belong to a disabled unichar_id
- pass UNICHARSET refs to check that
- in SetBlackAndWhitelist, also update the unicharset
  of the lstm_recognizer_ instance, if any

### DIFF
--- a/src/ccmain/tesseractclass.cpp
+++ b/src/ccmain/tesseractclass.cpp
@@ -630,6 +630,12 @@ void Tesseract::SetBlackAndWhitelist() {
     sub_langs_[i]->unicharset.set_black_and_whitelist(
         tessedit_char_blacklist.string(), tessedit_char_whitelist.string(),
         tessedit_char_unblacklist.string());
+    if (sub_langs_[i]->lstm_recognizer_) {
+      UNICHARSET& lstm_unicharset = const_cast<UNICHARSET&> (sub_langs_[i]->lstm_recognizer_->GetUnicharset());
+      lstm_unicharset.set_black_and_whitelist(tessedit_char_blacklist.string(),
+                                              tessedit_char_whitelist.string(),
+                                              tessedit_char_unblacklist.string());
+    }
   }
 }
 
--- a/src/ccmain/tesseractclass.cpp
+++ b/src/ccmain/tesseractclass.cpp
@@ -619,6 +619,12 @@ void Tesseract::SetBlackAndWhitelist() {
   unicharset.set_black_and_whitelist(tessedit_char_blacklist.string(),
                                      tessedit_char_whitelist.string(),
                                      tessedit_char_unblacklist.string());
+  if (lstm_recognizer_) {
+    UNICHARSET& lstm_unicharset = const_cast<UNICHARSET&> (lstm_recognizer_->GetUnicharset());
+    lstm_unicharset.set_black_and_whitelist(tessedit_char_blacklist.string(),
+                                            tessedit_char_whitelist.string(),
+                                            tessedit_char_unblacklist.string());
+  }
   // Black and white lists should apply to all loaded classifiers.
   for (int i = 0; i < sub_langs_.size(); ++i) {
     sub_langs_[i]->unicharset.set_black_and_whitelist(
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -83,9 +83,7 @@ run_command() {
     fi
     shift
     tlog "[$(date)] ${cmd} $@"
-    "${cmd}" "$@" 2>&1 1>&2 | tee -a ${LOG_FILE}
-    # check completion status
-    if [[ $? -gt 0 ]]; then
+    if ! "${cmd}" "$@" 2>&1 1>&2 | tee -a ${LOG_FILE}; then
         err_exit "Program $(basename ${cmd}) failed. Abort."
     fi
 }
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -69,21 +69,14 @@ err_exit() {
 # if the program file is not found.
 # Usage: run_command CMD ARG1 ARG2...
 run_command() {
-    local cmd=$(which $1)
-    if [[ -z ${cmd} ]]; then
-      for d in api training; do
-        cmd=$(which $d/$1)
-        if [[ ! -z ${cmd} ]]; then
-          break
-        fi
-      done
-      if [[ -z ${cmd} ]]; then
-          err_exit "$1 not found"
-      fi
-    fi
+    local cmd
+    cmd=$(which $1 || \
+              for d in api training; do
+                  which $d/$1 && break
+              done) || err_exit "'$1' not found"
     shift
     tlog "[$(date)] ${cmd} $@"
-    if ! "${cmd}" "$@" 2>&1 1>&2 | tee -a ${LOG_FILE}; then
+    if ! "${cmd}" "$@" |& tee -a ${LOG_FILE}; then
         err_exit "Program $(basename ${cmd}) failed. Abort."
     fi
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,9 @@ You should always make sure your changes build and run successfully.
 
 For that, your clone needs to have all submodules (`abseil`, `googletest`, `test`) included. To do so, either specify `--recurse-submodules` during the initial clone, or run `git submodule update --init --recursive NAME` for each `NAME` later. If `configure` already created those directories (blocking the clone), remove them first (or `make distclean`), then clone and reconfigure. 
 
-Have a look at README.md and the [wiki page](https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation#unit-test-builds) on installation. In short, to build the library and CLI, run `make`. To test it, run `make check`. To build the training tools, run `make training`.
+Have a look at [the README](./README.md) and [testing README](./test/testing/README.md) and the [wiki page](https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation#unit-test-builds) on installation. 
+
+In short, after running `configure` from the build directory of your choice, to build the library and CLI, run `make`. To test it, run `make check`. To build the training tools, run `make training`.
 
 As soon as your changes are building and tests are succeeding, you can publish them. If you have not already, please [fork](https://guides.github.com/activities/forking/) tesseract (somewhere) on GitHub, and push your changes to that fork (in a new branch). Then [submit as PR](https://help.github.com/en/articles/creating-a-pull-request-from-a-fork).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,4 +67,12 @@ your question has been asked (and has been answered) many times before...
 
 ## For Developers: Creating a Pull Request
 
-TBD
+You should always make sure your changes build and run successfully. 
+
+For that, your clone needs to have all submodules (`abseil`, `googletest`, `test`) included. To do so, either specify `--recurse-submodules` during the initial clone, or run `git submodule update --init --recursive NAME` for each `NAME` later. If `configure` already created those directories (blocking the clone), remove them first (or `make distclean`), then clone and reconfigure. 
+
+Have a look at README.md and the [wiki page](https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation#unit-test-builds) on installation. In short, to build the library and CLI, run `make`. To test it, run `make check`. To build the training tools, run `make training`.
+
+As soon as your changes are building and tests are succeeding, you can publish them. If you have not already, please [fork](https://guides.github.com/activities/forking/) tesseract (somewhere) on GitHub, and push your changes to that fork (in a new branch). Then [submit as PR](https://help.github.com/en/articles/creating-a-pull-request-from-a-fork).
+
+Please also keep track of reports from CI (automated build status) and Coverity/LGTM (quality scan). When the indicators show deterioration after your changes, further action may be required to improve them.
--- a/src/ccutil/unicharset.h
+++ b/src/ccutil/unicharset.h
@@ -871,6 +871,7 @@ class UNICHARSET {
 
   // Return the enabled property of the given unichar.
   bool get_enabled(UNICHAR_ID unichar_id) const {
+    ASSERT_HOST(contains_unichar_id(unichar_id));
     return unichars[unichar_id].properties.enabled;
   }
 
--- a/src/lstm/recodebeam.cpp
+++ b/src/lstm/recodebeam.cpp
@@ -87,7 +87,7 @@ void RecodeBeamSearch::Decode(const NetworkIO& output, double dict_ratio,
   if (lstm_choice_mode)
     timesteps.clear();
   for (int t = 0; t < width; ++t) {
-    ComputeTopN(output.f(t), output.NumFeatures(), kBeamWidths[0], charset);
+    ComputeTopN(output.f(t), output.NumFeatures(), kBeamWidths[0]);
     DecodeStep(output.f(t), t, dict_ratio, cert_offset, worst_dict_cert,
                charset);
     if (lstm_choice_mode) {
@@ -102,7 +102,7 @@ void RecodeBeamSearch::Decode(const GENERIC_2D_ARRAY<float>& output,
   beam_size_ = 0;
   int width = output.dim1();
   for (int t = 0; t < width; ++t) {
-    ComputeTopN(output[t], output.dim2(), kBeamWidths[0], charset);
+    ComputeTopN(output[t], output.dim2(), kBeamWidths[0]);
     DecodeStep(output[t], t, dict_ratio, cert_offset, worst_dict_cert, charset);
   }
 }
@@ -456,19 +456,12 @@ WERD_RES* RecodeBeamSearch::InitializeWord(bool leading_space,
 // Fills top_n_flags_ with bools that are true iff the corresponding output
 // is one of the top_n.
 void RecodeBeamSearch::ComputeTopN(const float* outputs, int num_outputs,
-                                   int top_n, const UNICHARSET* charset) {
+                                   int top_n) {
   top_n_flags_.init_to_size(num_outputs, TN_ALSO_RAN);
   top_code_ = -1;
   second_code_ = -1;
   top_heap_.clear();
   for (int i = 0; i < num_outputs; ++i) {
-    // Decode label via recoder_.
-    RecodedCharID code;
-    code.Set(0, i);
-    int label = recoder_.DecodeUnichar(code);
-    if (label != INVALID_UNICHAR_ID && // not part of a bigger code.
-        !charset->get_enabled(label)) // disabled
-      continue;
     if (top_heap_.size() < top_n || outputs[i] > top_heap_.PeekTop().key) {
       TopPair entry(outputs[i], i);
       top_heap_.Push(&entry);
@@ -505,10 +498,10 @@ void RecodeBeamSearch::DecodeStep(const float* outputs, int t,
   if (t == 0) {
     // The first step can only use singles and initials.
     ContinueContext(nullptr, BeamIndex(false, NC_ANYTHING, 0), outputs, TN_TOP2,
-                    dict_ratio, cert_offset, worst_dict_cert, step);
+                    charset, dict_ratio, cert_offset, worst_dict_cert, step);
     if (dict_ != nullptr) {
-      ContinueContext(nullptr, BeamIndex(true, NC_ANYTHING, 0), outputs,
-                      TN_TOP2, dict_ratio, cert_offset, worst_dict_cert, step);
+      ContinueContext(nullptr, BeamIndex(true, NC_ANYTHING, 0), outputs, TN_TOP2,
+                      charset, dict_ratio, cert_offset, worst_dict_cert, step);
     }
   } else {
     RecodeBeam* prev = beam_[t - 1];
@@ -540,9 +533,8 @@ void RecodeBeamSearch::DecodeStep(const float* outputs, int t,
         // best first, but it comes before a lot of the worst, so it is slightly
         // more efficient than going forwards.
         for (int i = prev->beams_[index].size() - 1; i >= 0; --i) {
-          ContinueContext(&prev->beams_[index].get(i).data, index, outputs,
-                          top_n, dict_ratio, cert_offset, worst_dict_cert,
-                          step);
+          ContinueContext(&prev->beams_[index].get(i).data, index, outputs, top_n,
+                          charset, dict_ratio, cert_offset, worst_dict_cert, step);
         }
       }
       for (int index = 0; index < kNumBeams; ++index) {
@@ -569,7 +561,9 @@ void RecodeBeamSearch::DecodeStep(const float* outputs, int t,
 // choices for which top_n_flags[index] == top_n_flag.
 void RecodeBeamSearch::ContinueContext(const RecodeNode* prev, int index,
                                        const float* outputs,
-                                       TopNState top_n_flag, double dict_ratio,
+                                       TopNState top_n_flag,
+                                       const UNICHARSET* charset,
+                                       double dict_ratio,
                                        double cert_offset,
                                        double worst_dict_cert,
                                        RecodeBeam* step) {
@@ -632,6 +626,8 @@ void RecodeBeamSearch::ContinueContext(const RecodeNode* prev, int index,
       int unichar_id = recoder_.DecodeUnichar(full_code);
       // Map the null char to INVALID.
       if (length == 0 && code == null_char_) unichar_id = INVALID_UNICHAR_ID;
+      if (unichar_id != INVALID_UNICHAR_ID && !charset->get_enabled(unichar_id))
+        continue; // disabled by whitelist/blacklist
       ContinueUnichar(code, unichar_id, cert, worst_dict_cert, dict_ratio,
                       use_dawgs, NC_ANYTHING, prev, step);
       if (top_n_flag == TN_TOP2 && code != null_char_) {
--- a/src/lstm/recodebeam.cpp
+++ b/src/lstm/recodebeam.cpp
@@ -87,7 +87,7 @@ void RecodeBeamSearch::Decode(const NetworkIO& output, double dict_ratio,
   if (lstm_choice_mode)
     timesteps.clear();
   for (int t = 0; t < width; ++t) {
-    ComputeTopN(output.f(t), output.NumFeatures(), kBeamWidths[0]);
+    ComputeTopN(output.f(t), output.NumFeatures(), kBeamWidths[0], charset);
     DecodeStep(output.f(t), t, dict_ratio, cert_offset, worst_dict_cert,
                charset);
     if (lstm_choice_mode) {
@@ -102,7 +102,7 @@ void RecodeBeamSearch::Decode(const GENERIC_2D_ARRAY<float>& output,
   beam_size_ = 0;
   int width = output.dim1();
   for (int t = 0; t < width; ++t) {
-    ComputeTopN(output[t], output.dim2(), kBeamWidths[0]);
+    ComputeTopN(output[t], output.dim2(), kBeamWidths[0], charset);
     DecodeStep(output[t], t, dict_ratio, cert_offset, worst_dict_cert, charset);
   }
 }
@@ -456,12 +456,19 @@ WERD_RES* RecodeBeamSearch::InitializeWord(bool leading_space,
 // Fills top_n_flags_ with bools that are true iff the corresponding output
 // is one of the top_n.
 void RecodeBeamSearch::ComputeTopN(const float* outputs, int num_outputs,
-                                   int top_n) {
+                                   int top_n, const UNICHARSET* charset) {
   top_n_flags_.init_to_size(num_outputs, TN_ALSO_RAN);
   top_code_ = -1;
   second_code_ = -1;
   top_heap_.clear();
   for (int i = 0; i < num_outputs; ++i) {
+    // Decode label via recoder_.
+    RecodedCharID code;
+    code.Set(0, i);
+    int label = recoder_.DecodeUnichar(code);
+    if (label != INVALID_UNICHAR_ID && // not part of a bigger code.
+        !charset->get_enabled(label)) // disabled
+      continue;
     if (top_heap_.size() < top_n || outputs[i] > top_heap_.PeekTop().key) {
       TopPair entry(outputs[i], i);
       top_heap_.Push(&entry);
--- a/src/lstm/recodebeam.cpp
+++ b/src/lstm/recodebeam.cpp
@@ -626,7 +626,9 @@ void RecodeBeamSearch::ContinueContext(const RecodeNode* prev, int index,
       int unichar_id = recoder_.DecodeUnichar(full_code);
       // Map the null char to INVALID.
       if (length == 0 && code == null_char_) unichar_id = INVALID_UNICHAR_ID;
-      if (unichar_id != INVALID_UNICHAR_ID && !charset->get_enabled(unichar_id))
+      if (unichar_id != INVALID_UNICHAR_ID &&
+          charset != nullptr &&
+          !charset->get_enabled(unichar_id))
         continue; // disabled by whitelist/blacklist
       ContinueUnichar(code, unichar_id, cert, worst_dict_cert, dict_ratio,
                       use_dawgs, NC_ANYTHING, prev, step);
--- a/src/lstm/recodebeam.h
+++ b/src/lstm/recodebeam.h
@@ -293,7 +293,7 @@ class RecodeBeamSearch {
 
   // Fills top_n_flags_ with bools that are true iff the corresponding output
   // is one of the top_n.
-  void ComputeTopN(const float* outputs, int num_outputs, int top_n);
+  void ComputeTopN(const float* outputs, int num_outputs, int top_n, const UNICHARSET* unicharset);
 
   // Adds the computation for the current time-step to the beam. Call at each
   // time-step in sequence from left to right. outputs is the activation vector
--- a/src/lstm/recodebeam.h
+++ b/src/lstm/recodebeam.h
@@ -293,7 +293,7 @@ class RecodeBeamSearch {
 
   // Fills top_n_flags_ with bools that are true iff the corresponding output
   // is one of the top_n.
-  void ComputeTopN(const float* outputs, int num_outputs, int top_n, const UNICHARSET* unicharset);
+  void ComputeTopN(const float* outputs, int num_outputs, int top_n);
 
   // Adds the computation for the current time-step to the beam. Call at each
   // time-step in sequence from left to right. outputs is the activation vector
@@ -310,9 +310,9 @@ class RecodeBeamSearch {
   // using the given network outputs to provide scores to the choices. Uses only
   // those choices for which top_n_flags[code] == top_n_flag.
   void ContinueContext(const RecodeNode* prev, int index, const float* outputs,
-                       TopNState top_n_flag, double dict_ratio,
-                       double cert_offset, double worst_dict_cert,
-                       RecodeBeam* step);
+                       TopNState top_n_flag, const UNICHARSET* unicharset,
+                       double dict_ratio, double cert_offset,
+                       double worst_dict_cert, RecodeBeam* step);
   // Continues for a new unichar, using dawg or non-dawg as per flag.
   void ContinueUnichar(int code, int unichar_id, float cert,
                        float worst_dict_cert, float dict_ratio, bool use_dawgs,


src/ccmain/tesseractclass.cpp
```diff
@@ -619,6 +619,12 @@ void Tesseract::SetBlackAndWhitelist() {
   unicharset.set_black_and_whitelist(tessedit_char_blacklist.string(),
                                      tessedit_char_whitelist.string(),
                                      tessedit_char_unblacklist.string());
+  if (lstm_recognizer_) {
```
stweil: Similar code might be needed in the `for` loop below (after line 633) for the `sub_langs_`.

zdenop:
`sub_langs_` will be handled withing this PR (should be merged now or should we wait for additional commit)?

src/ccmain/tesseractclass.cpp
bertsky: sure, I will have a look.

bertsky:
@zdenop, I will add that to the PR of course. It should be ready by tomorrow.

bertsky:
So are there no objections to the way I use `RecodedCharID` (label mapping) and `UnicharCompress` (recoder)? I feel a little uncomfortable here.

src/ccmain/tesseractclass.cpp
bertsky: done

bertsky:
It works with sublangs now. But I just saw it does not work with Chinese characters yet. So I guess my `RecodedCharID` attempt is indeed not enough...

Shreeshrii:
Regarding recoder, please see the unittest
https://github.com/tesseract-ocr/tesseract/blob/master/unittest/lstm_test.h#L163

Shreeshrii:
Please also test that `lstmtraining`, `lstmeval` and other training programs work with the changes.

bertsky:
Thanks, @Shreeshrii, I will.

amitdo:
Maybe you should remove the recoding part until you figure out how to make it work?

bertsky:
Well it does work for simple characters, but not for things like CJK. And I cannot avoid using it – this is the component mediating between unicharset ids and output channels. But I am positive I will understand it by looking at the unittests and training tools.

Shreeshrii:
The output of unicharcompress/recoder can be seen by the following:

```
rm -rf ~/tesstutorial/kortest
bash  ~/tesseract/src/training/tesstrain.sh \
  --fonts_dir /usr/share/fonts \
  --lang kor \
  --linedata_only \
  --save_box_tiff \
  --workspace_dir ~/tmp \
  --exposures "0" \
  --maxpages 1 \
  --noextract_font_properties \
  --langdata_dir ~/langdata_lstm \
  --tessdata_dir ~/tessdata_best  \
  --fontlist "Arial Unicode MS" \
  --training_text ~/langdata_lstm/kor/kor.training_text \
  --output_dir ~/tesstutorial/kortest
```
Using 1 page of training_text input, it creates a unicharset of 659 which is compressed to 111.

```
659
NULL 0 NULL 0
Joined 7 0,69,188,255,486,1218,0,30,486,1188 Latin 1 0 1 Joined	# Joined [4a 6f 69 6e 65 64 ]a
|Broken|0|1 f 0,69,186,255,892,2138,0,80,892,2058 Common 216 10 216 |Broken|0|1	# Broken
~ 0 111,154,139,180,90,124,4,19,100,155 Common 3 10 3 ~	# ~ [7e ]
에 1 28,63,223,255,160,206,0,31,218,230 Hangul 4 0 4 에	# 에 [c5d0 ]x
서 1 28,63,224,255,169,196,0,24,218,230 Hangul 5 0 5 서	# 서 [c11c ]x
는 1 46,75,213,255,185,204,6,22,218,230 Hangul 6 0 6 는	# 는 [b294 ]x
지 1 28,63,223,255,169,194,3,29,218,230 Hangul 7 0 7 지	# 지 [c9c0 ]x
내 1 30,63,226,255,160,186,8,34,218,230 Hangul 8 0 8 내	# 내 [b0b4 ]x
고 1 58,87,198,253,179,203,6,24,218,230 Hangul 9 0 9 고	# 고 [ace0 ]x
났 1 30,63,221,255,190,211,13,17,218,230 Hangul 10 0 10 났	# 났 [b0ac ]x
습 1 30,67,219,255,185,204,6,22,218,230 Hangul 11 0 11 습	# 습 [c2b5 ]x
니 1 30,65,223,255,152,182,13,39,218,230 Hangul 12 0 12 니	# 니 [b2c8 ]x
다 1 27,63,221,255,186,203,13,25,218,230 Hangul 13 0 13 다	# 다 [b2e4 ]x
후 1 27,64,223,255,181,203,5,24,218,230 Hangul 14 0 14 후	# 후 [d6c4 ]x
촉 1 28,63,224,255,179,204,6,25,218,230 Hangul 15 0 15 촉	# 촉 [cd09 ]x
진 1 40,69,223,255,171,203,0,27,218,230 Hangul 16 0 16 진	# 진 [c9c4 ]x
- 10 103,154,114,173,55,157,14,19,86,195 Common 17 3 17 -	# - [2d ]p
```

```
0	 
110	<nul>
1	~
62,74,90	에
60,73,90	서
53,87,93	는
63,89,90	지
53,71,90	내
51,77,90	고
53,70,103	났
60,87,100	습
53,89,90	니
54,70,90	다
69,82,90	후
65,77,91	촉
63,89,93	진
2	-
```

```

**Comment by bertsky** [comment]
Found it. I believe I do understand now. It does work for Chinese characters, too. I am pretty sure it works for all other cases as well.

@Shreeshrii, I very much doubt that training will be affected, as this is a runtime variable. Also, I have never done training before. What would be required? The lstmtraining tutorial of [the training page](https://github.com/tesseract-ocr/tesseract/wiki/TrainingTesseract-4.00) in the wiki? How do I know it still works?

**Comment by bertsky** [comment]
@stweil, do you want me to squash the three commits into one (with a better changelog) on this PR?

**Comment by Shreeshrii** [comment]
@bertsky I applied the PR to my local installation and I am getting the following error with lstmeval 

```
(gdb) run
Starting program: /usr/local/bin/lstmeval --verbosity -1 --model /home/ubuntu/tesstutorial/IAST_PLUS/plusminus_checkpoint --traineddata /home/ubuntu/tesstutorial/IAST_PLUS/iast/iast.traineddata --eval_listfile /home/ubuntu/tesstutorial/IAST_eval_1/iast.training_files.txt
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/powerpc64le-linux-gnu/libthread_db.so.1".
/home/ubuntu/tesstutorial/IAST_PLUS/plusminus_checkpoint is not a recognition model, trying training checkpoint...
[New Thread 0x3fffb6e5f100 (LWP 30849)]
Loaded 116/116 pages (1-116) of document /home/ubuntu/tesstutorial/IAST_eval_1/iast.Gentium_Basic_Italic.exp0.lstmf
[Thread 0x3fffb6e5f100 (LWP 30849) exited]
[New Thread 0x3fffb6e5f100 (LWP 30850)]
Loaded 116/116 pages (1-116) of document /home/ubuntu/tesstutorial/IAST_eval_1/iast.Gentium_Book_Basic_Italic.exp0.lstmf
[Thread 0x3fffb6e5f100 (LWP 30850) exited]
[New Thread 0x3fffb6e5f100 (LWP 30851)]
[New Thread 0x3fffb665f100 (LWP 30852)]
[New Thread 0x3fffb5e5f100 (LWP 30853)]

Thread 1 "lstmeval" received signal SIGSEGV, Segmentation fault.
0x00003fffb7dc76b4 in UNICHARSET::get_enabled (this=0x0, unichar_id=<optimized out>) at ../../src/ccutil/unicharset.h:874
874         return unichars[unichar_id].properties.enabled;
(gdb) backtrace
#0  0x00003fffb7dc76b4 in UNICHARSET::get_enabled (this=0x0, unichar_id=<optimized out>) at ../../src/ccutil/unicharset.h:874
#1  tesseract::RecodeBeamSearch::ContinueContext (this=0x11078440, prev=0x0, index=<optimized out>, outputs=0x181d1eb0, top_n_flag=<optimized out>, charset=0x0,
    dict_ratio=1, cert_offset=0, worst_dict_cert=-20, step=0x164c63c0) at recodebeam.cpp:629
#2  0x00003fffb7dc7ec4 in tesseract::RecodeBeamSearch::DecodeStep (this=0x11078440, outputs=0x181d1eb0, t=<optimized out>, dict_ratio=1, cert_offset=0,
    worst_dict_cert=-20, charset=0x0, debug=<optimized out>) at recodebeam.cpp:501
#3  0x00003fffb7dc849c in tesseract::RecodeBeamSearch::Decode (this=0x11078440, output=..., dict_ratio=1, cert_offset=0, worst_dict_cert=-20, charset=0x0,
    lstm_choice_mode=<optimized out>) at recodebeam.cpp:92
#4  0x00003fffb7da32dc in tesseract::LSTMRecognizer::LabelsViaReEncode (this=0x3fffffffdc80, output=..., labels=0x3fffffffd928, xcoords=0x3fffffffd948)
    at lstmrecognizer.cpp:436
#5  0x00003fffb7da4084 in tesseract::LSTMRecognizer::LabelsFromOutputs (this=0x3fffffffdc80, outputs=..., labels=0x3fffffffd928, xcoords=<optimized out>)
    at lstmrecognizer.cpp:423
#6  0x00003fffb7dabbd4 in tesseract::LSTMTrainer::PrepareForBackward (this=0x3fffffffdc78, trainingdata=0x3fffb0000910, fwd_outputs=0x3fffffffdb58,
    targets=0x3fffffffdbe8) at lstmtrainer.cpp:867
#7  0x000000001000a198 in tesseract::LSTMTester::RunEvalSync (this=0x3fffffffebc0, iteration=<optimized out>, training_errors=<optimized out>, model_mgr=...,
    training_stage=<optimized out>, verbosity=<optimized out>) at lstmtester.cpp:101
#8  0x0000000010004914 in main (argc=<optimized out>, argv=<optimized out>) at lstmeval.cpp:79
(gdb)
```

**Comment by Shreeshrii** [comment]
`make check` also fails on certain unitests.

PASS: qrsequence_test
../config/test-driver: line 107: 31583 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: recodebeam_test
PASS: rect_test
PASS: resultiterator_test
PASS: shapetable_test
PASS: stats_test
PASS: stringrenderer_test
PASS: tablefind_test
PASS: tablerecog_test
PASS: tabvector_test
PASS: tfile_test
PASS: commandlineflags_test
../config/test-driver: line 107: 31801 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: lstm_recode_test
../config/test-driver: line 107: 31825 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: lstm_squashed_test
../config/test-driver: line 107: 31849 Segmentation fault      (core dumped) "$@" > $log_file 2>&1
FAIL: lstm_test
PASS: lstmtrainer_test

**Comment by stweil** [comment]
> @stweil, do you want me to squash the three commits into one (with a better changelog) on this PR?

@bertsky, first of all thank you for addressing this issue with your pull request. We can decide about squashing as soon as the changes are ready for the production code. Then squashing can optionally be done either by you or by the maintainer who merges the pull request.

I hope that I'll have finished some other work on Tesseract soon, then I can help with fixing the problems reported above by @Shreeshrii.

**Comment by bertsky** [comment]
@Shreeshrii thanks a lot! I will look into it. As for the lstmeval test, will I have to go through tesstutorial on the wiki for that?

I was not aware that `make check` is supposed to work at all. (Also I think it should be named `make test` as for most automake projects.) The problem is, it does not find the googletest submodule and thus fails to build for me. Is this a missing dependency? (`googletest/googletest/src` is empty.) And is this related to https://github.com/tesseract-ocr/test perhaps?

**Comment by Shreeshrii** [comment]
Try `git submodule update --init --recursive` to include `googletest`, `test` and `abseil` submodules when you build tesseract.

`make check` works but it has a lot of dependencies in terms of having all three repos of tessdata, tessdata_best, tessdata_fast and the `test` repo for the testdata files. 

Regarding `LSTM training` - yes, tesstutorial is the way to go, though most people complain about it being difficult to follow. As you go through it, please note/make any changes that can be made to improve the wiki.

@zdenop It has been suggested to have a separate repo for `tesstutorial` with all required dependencies in it. I can upload the files for it if you create a new repo.

**Comment by Shreeshrii** [comment]
[tesstutorial-eng.sh.txt](https://github.com/tesseract-ocr/tesseract/files/2946038/tesstutorial-eng.sh.txt)

@bertsky Please see the attached file (bash script) which has the commands for tesstutorial (for the first few types of training). You can use it as a base, modify for your environment and test.

**Comment by stweil** [comment]
@bertsky, `make check` also expects clones of `traineddata`, `traineddata_best` and `traineddata_fast` parallel to the clone of `tesseract`.

**Comment by bertsky** [comment]
@stweil, are you sure `check` depends on them (not the tesstutorial)? And didn't you mean `tessdata`, `tessdata_best` and `tessdata_fast`? And where would they have to be placed?

**Comment by bertsky** [comment]
@Shreeshrii, thanks, that worked (after removing the existing directories, which had been filled with dependencies by configure already, and then re-instating them after cloning). I can see the segfaults now. I will address them first, maybe it's actually the same problem for lstmeval.

I think I should create another PR with some minimal notes in the dev section of `CONTRIBUTING.md` after this.

**Comment by bertsky** [comment]
Oh, right, just found [where check expects](https://github.com/tesseract-ocr/tesseract/blob/bd95c9d2b8c2ac8ffb7d0894f8c3b2b0e92a19b4/unittest/Makefile.am#L6) the other repos. That's not such a nice way of maintaining dependencies! Wouldn't it be better to add them as submodules, too? Or at least provide some option to configure for where they are located?

**Comment by stweil** [comment]
Yes, sorry, the names were wrong. I fixed my comment now. The directory with the model files are required for the tests run with `make check`. The directories `tessdata`, ... are expected in the same directory hierarchy as the `tesseract` source tree.

**Comment by bertsky** [comment]
Why not integrate them as submodules overtly, just like `test`, `googletest`, and `abseil`?

**Comment by stweil** [comment]
The answer is simple: 8 GB. That's a lot of disk space on my notebook which I cannot afford. The model files are really large, and their git repository is even larger.

**Comment by stweil** [comment]
Some tests don't need those models, and others only use a few of them. So the better solution would be to skip each test with missing dependencies automatically.

**Comment by bertsky** [comment]
I can see that. But a submodule does not _have to_  be cloned and initialized, unless one wants to (for testing). At the moment, no outsider gets a fair chance of knowing what is needed and where.

**Comment by stweil** [comment]
You are right. The only excuse is that many of the tests were added only recently, so both the code is still incomplete and the information on the [Wiki](https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation) is still missing.

**Comment by stweil** [comment]
There is now at least a little bit documentation [here](https://github.com/tesseract-ocr/tesseract/wiki/Compiling-%E2%80%93-GitInstallation#running-unit-tests).

**Comment by bertsky** [comment]
Should I open an issue about those dependencies then? 

With your permission, I would like to add `langdata_lstm` to the documentation on the wiki, as well as `--recurse-submodules` (which is also necessary to get `abseil`, `googletest` and `test`).

**Comment by stweil** [comment]
Sure, feel free to fix or enhance the documentation on the Wiki. Are you sure that recursion is needed? I never used it, and all three submodules don't contain submodules, so it won't make a difference.

There is already an open issue for unit tests (#1627), so it's possible to add things there. Or you open a new one.

**Comment by bertsky** [comment]
Sorry, no, of course your `submodules update` will suffice.

Thanks, that will be an interesting read. As I can see, you already advocated for submodules there.

**Comment by bertsky** [comment]
> Sorry, no, of course your `submodules update` will suffice.

...unless, of course, `tessdata` ... `langdata_lstm` were added as submodules to `test` – which would probably be even better.

**Comment by bertsky** [comment]
So. 3912cb1 was needed against the true regressions. (And giving segfaults is a _deliberate escalation strategy_ in `ccutil/errcode.cpp` for debug builds.) The rest seems to be outdated or wrong unittests and tesstutorial shortcomings. I am trying to help with that as well, but really the repo should always be left in a clean, regression-free state...

**Comment by bertsky** [comment]
I have 2 remaining regressions:
- `lstm_test` cannot do `InitCharSet` – my guess is that `FLAGS_test_tmpdir` is wrong
- `stringrenderer_test` complains that its `DoesStripUnrenderableWords` does not strip "والفكر 123 والصراع abc" to " 123  abc" – this does not relate to my change either

**Comment by Shreeshrii** [comment]
what about `lstmeval`?

**Comment by bertsky** [comment]
@Shreeshrii, as to `tesstutorial`, I get many memory leak reports. But when building without the leakage sanitizers, (and adapting to all the filesystem relative path assumptions in the script you provided,) after 349558d and acaa100, I am at least able to get through. So 3912cb1 was indeed the cause for both check and lstmeval failures. 

Does that settle it? Should I post the results/log?

**Comment by Shreeshrii** [comment]
@bertsky Thank you. I will give it a try.

**Comment by Shreeshrii** [comment]
Both `make check` and `lstmeval` run without errors. Thanks!

**Comment by Shreeshrii** [comment]
The error with `lstm_test ` is related to size of unicharset and is probably related to the tessdata dir being used for testing.

```
[ RUN      ] LSTMTrainerTest.TestLayerAccess
Config file is optional, continuing...
Failed to read data from: /home/ubuntu/langdata_lstm//eng/eng.config
Warning: given outputs 1 not equal to unicharset of 113.
Num outputs,weights in Series:
  1,32,0,1:1, 0
Num outputs,weights in Series:
  C5,5:25, 0
  Ft16:16, 416
Total weights = 416
  [C5,5Ft16]:16, 416
  Mp2,2:16, 0
  Lfys32:32, 6272
  Lbx128:256, 164864
  Fc113:113, 29041
Total weights = 200593
Built network:[1,32,0,1[C5,5Ft16]Mp2,2Lfys32Lbx128Fc113] from request [1,32,0,1 Ct5,5,16 Mp2,2 Lfys32 Lbx128 O1c1]
Training parameters:
  Debug interval = 0, weights = 0.1, learning rate = 0.0005, momentum=0.9
null char=2
Loaded 520/520 pages (1-520) of document /home/ubuntu/tesseract/test/testdata/eng.Arial.exp0.lstmf
[INFO]  Setup network:SQU-lstm
lstm_test.cc:217: Failure
Expected equality of these values:
  kNumWeights[i]
    Which is: 28784
  trainer_->GetLayer(layers[i])->num_weights()
    Which is: 29041
[  FAILED  ] LSTMTrainerTest.TestLayerAccess (1009 ms)
[----------] 11 tests from LSTMTrainerTest (1172280 ms total)
```

```
  // Expected number of weights.
  const int kNumWeights[kNumLayers] = {0,
                                       0,
                                       16 * (25 + 1),
                                       0,
                                       32 * (4 * (32 + 16 + 1)),
                                       128 * (4 * (128 + 32 + 1)),
                                       128 * (4 * (128 + 32 + 1)),
                                       112 * (2 * 128 + 1)};
```

   112 * (2 * 128 + 1)}; = 28784

  113 * (2 * 128 + 1)}; = 29041

File: src/training/tesstrain_utils.sh [hunk 2]
```diff
@@ -83,9 +83,7 @@ run_command() {
     fi
     shift
     tlog "[$(date)] ${cmd} $@"
-    "${cmd}" "$@" 2>&1 1>&2 | tee -a ${LOG_FILE}
-    # check completion status
-    if [[ $? -gt 0 ]]; then
+    if ! "${cmd}" "$@" 2>&1 1>&2 | tee -a ${LOG_FILE}; then
```
stweil:
```suggestion
    if ! "${cmd}" "$@" 2>&1 | tee -a ${LOG_FILE}; then
```

bertsky: yes, I had overlooked that. And even worse, prior to that, `which` status was also not working (see 
91a7e86).

Shreeshrii:
Verified - fixes issue https://github.com/tesseract-ocr/tesseract/issues/1270 by use of blacklist

```
ubuntu@tesseract-ocr:~/TEST$ tesseract 1270.png - -c tessedit_char_blacklist=\|
Warning: Invalid resolution 0 dpi. Using 70 instead.
Estimating resolution as 533
DAVE: I always wondered
why no one tried to be
a superhero before me.
ubuntu@tesseract-ocr:~/TEST$ tesseract 1270.png -
Warning: Invalid resolution 0 dpi. Using 70 instead.
Estimating resolution as 533
DAVE: | always wondered
why no one tried to be
a superhero before me.
```

Shreeshrii:
@bertsky Do these changes also fix https://github.com/tesseract-ocr/tesseract/issues/960 ?

bertsky:
> Do these changes also fix #960 ?

It does not seem so. Results for your pattern example from #403 are still unaffected, regardless of whether I use a config file or the `--user-patterns` option. Stracing confirms that Tesseract never attempts to open the pattern file, it just goes straight after the output file, once the traineddata itself is loaded. 

Looking into this, it appears that `LSTMRecognizer::Load` is responsible, and it does not call `LoadDictionary` unless its first option (`lang`) is non-null, which in turn `Tesseract::init_tesseract_lang_data` will not give unless `lstm_use_matrix=1`. But that does not help either!

Looking deeper, only `Dict::LoadLSTM` is in the current callgraph, but we would need `Dict::Load` to read the user words and user patterns, build a trie from them and add to the other dawgs. This can only come from `Tesseract::init_tesseract_lm`, and that from `TessBaseAPI::InitLangMod`, which has a nice disclaimer comment above:

> //TODO(amit): Adapt to lstm

bertsky:
Should I rather comment on #403 directly to give that a boost?

Shreeshrii:
@bertsky Thank you for the investigation regarding #960 and #403.  Yes, you could comment on #403, though it would be best if you suggest a solution via a PR :-)  

I would also suggest that you split out the commits regarding unittests, tesstrain etc. as separate PRs from the one regarding whitelist/blacklist. Hopefully then the review and merge would happen faster.

Shreeshrii:
>only Dict::LoadLSTM is in the current callgraph, but we would need Dict::Load to read the user words and user patterns, build a trie from them and add to the other dawgs. This can only come from Tesseract::init_tesseract_lm, and that from TessBaseAPI::InitLangMod, which has a nice disclaimer comment above:

//TODO(amit): Adapt to lstm

@amitdo Is that your TODO?

@bertsky Since base tesseract and LSTM tesseract use different unicharsets, `Dict::LoadLSTM` would need to read to read the user words and user patterns, build a trie from them and add to the other lstm dawgs ...

bertsky:
I now believe `TessBaseAPI::InitLangMod` / `Tesseract::init_tesseract_lm` are actually dead ends and should be removed. As to `LSTMRecognizer::LoadDictionary`, it would simply be a matter of replacing `Dict::LoadLSTM` with the old `Dict::Load`, but there is one _missing link_: The `LSTMRecognizer` never gets to see the runtime variables of the `Tesseract` instance, and `CCUtil` has no interface to set or initialize its `params_` member. 

Perhaps it would be best to pass `tesseract_->params()` to the constructor of `LSTMRecognizer`, and add a (delegating) constructor to both `LSTMRecognizer` and `CCUtil` which takes a `ParamsVectors* `.

Or is there some reason to keep the member params of `Tesseract` and `LSTMRecognizer` different? If so, which params besides `user_patterns_file` and `user_words_file` should be copied?

zdenop:
@theraysmith : Ray, can you reply to @bertsky ? This is important fix for tesseract 4.x... (cc: @jbreiden )

amitdo:
>@amitdo Is that your TODO?

Yes.

Shreeshrii:
@bertsky Regarding https://github.com/tesseract-ocr/tesseract/pull/2294/commits/01f2aa630d71b0b0791690b5ef9659409a080445

I had made the `const char* FLAGS_test_tmpdir = "./tmp";`
so that the files generated during a test run can be cleaned out easily before next test run since they will be there in a separate directory. Probably the first time the directory has to be created.

If you change it to `TESTDATA_DIR` then all temporary files are also kept with files that are required before the tests.

stweil:
I recently added a Makefile rule to create  `tmp` automatically.

Shreeshrii:
>stringrenderer_test complains that its DoesStripUnrenderableWords does not strip "والفكر 123 والصراع abc" to " 123 abc" – this does not relate to my change either.

Please see https://github.com/tesseract-ocr/test/blob/master/testing/README.md

> Fonts required for unittests
> The following fonts should be copied to test/testing folder for stringrenderer_test to run.
> 
> Arab https://packages.ubuntu.com/trusty/fonts-arabeyes - GNU General Public License
> 
> Lohit Hindi https://releases.pagure.org/lohit/lohit-hindi-ttf-2.4.3.tar.gz - GNU General Public License, version 2
> 
> UNBatang https://packages.ubuntu.com/trusty/fonts-unfonts-core - GPL-2
> 
> Verdana https://packages.ubuntu.com/trusty/ttf-mscorefonts-installer - /usr/share/doc/ttf-mscorefonts-installer/copyright
>

Shreeshrii:
>lstm_test cannot do InitCharSet – my guess is that FLAGS_test_tmpdir is wrong

It was failing for me because of wrong testdata. I have uploaded a new version of files to test repo. It runs ok with current code from master branch.

bertsky:
Sorry everyone, I was ill the previous days. 

@Shreeshrii 
> Regarding [01f2aa6](https://github.com/tesseract-ocr/tesseract/commit/01f2aa630d71b0b0791690b5ef9659409a080445)
> 
> I had made the `const char* FLAGS_test_tmpdir = "./tmp";`
> so that the files generated during a test run can be cleaned out easily before next test run since they will be there in a separate directory. Probably the first time the directory has to be created.

Sorry, I had not looked into the history thoroughly. My immediate concern was getting the tests working "again". You are right of course, and with the additional b7279f6d6 from @stweil it should work better. 

Should I add a revert here, or purge the change from the history of that branch and push forced?

bertsky:
Thanks @Shreeshrii for clarifying why `stringrenderer_test` failed for me, and updating `lstm_test`.

bertsky:
> Please see https://github.com/tesseract-ocr/test/blob/master/testing/README.md

I wonder if that pointer could be made more prominent. Maybe it should go into `CONTRIBUTING.md` as of this PR?

bertsky:
done.

Shreeshrii:
Verified use of --whitelist (using example from [stackoverflow](https://stackoverflow.com/questions/45622435/special-character-whitelist-with-tesseract-ocr).)

-c tessedit_char_whitelist="0123456789,.$¢" 

```
tesseract ~/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic.exp0.tif ~/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic  
tesseract ~/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic.exp0.tif ~/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic-WL  -c tessedit_char_whitelist="0123456789,.$¢" -c lstm_use_matrix=1

diff -u    ~/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic.txt ~/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic-WL.txt
```

Output:
```
--- /home/ubuntu/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic.txt     2019-03-20 19:45:44.228812479 +0000
+++ /home/ubuntu/TEST/USER/WHITELIST/eng.Brush_Script_MT_Italic-WL.txt  2019-03-20 19:45:45.708823579 +0000
@@ -3,11 +3,11 @@
 6.526.08
 4.90.59.310.38
 100.00
-0.05%
+0.05¢
 501.99
 99.99
 $99,909.15
 $7.11.813.04
 $1456569.59
-20. 000%
+20.000¢

\ No newline at end of file
```

Shreeshrii:
OK for English for [blacklist](https://github.com/tesseract-ocr/tesseract/pull/2294#issuecomment-471661380) and [whitelist](https://github.com/tesseract-ocr/tesseract/pull/2294#issuecomment-475016949)

I have not yet tested for any other language.

Shreeshrii:
OK for whitelist - Devanagari Script numbers.

   -c tessedit_char_whitelist="०१२३४५६७८९.,"

Changes visible in Zero - 0 vs ०

```
Tesseract Open Source OCR Engine v4.1.0-rc1-131-g632a2 with Leptonica
Page 1
Tesseract Open Source OCR Engine v4.1.0-rc1-131-g632a2 with Leptonica
Page 1
--- /home/ubuntu/TEST/USER/deva1/san.Ek_Mukta.exp0.txt  2019-03-21 08:10:00.574883387 +0000
+++ /home/ubuntu/TEST/USER/deva1/san.Ek_Mukta.exp0-WL.txt       2019-03-21 08:10:02.002894097 +0000
@@ -1,6 +1,6 @@
 ००१
 ५०१
-0०००
+०००
 १,००,०००.००
 ००९
 ९,००,०००.००
/home/ubuntu/TEST/USER/deva1/san.Samanata.exp0
Tesseract Open Source OCR Engine v4.1.0-rc1-131-g632a2 with Leptonica
Page 1
Tesseract Open Source OCR Engine v4.1.0-rc1-131-g632a2 with Leptonica
Page 1
--- /home/ubuntu/TEST/USER/deva1/san.Samanata.exp0.txt  2019-03-21 08:10:03.374904387 +0000
+++ /home/ubuntu/TEST/USER/deva1/san.Samanata.exp0-WL.txt       2019-03-21 08:10:04.758914767 +0000
@@ -1,6 +1,6 @@
 ००१
 ५०१
-0००
+००
 १,००,०००.००
 ००९
 ९,००,०००.००

```

Shreeshrii:
@bertsky  Regarding the config file `lstm.train` - I don't think the full path should be given in the training scripts, because these files are installed along with other configs such as pdf, hocr, etc and users who are installing from a distribution and not building from source won't have the files in  $(dirname $0)/../../tessdata/configs/lstm.train .

bertsky:
@Shreeshrii As stated above, I made acaa1003 specifically to get tesstutorial running. That and `test/testdata/README.md` are the only places I know of referencing tesstrain.sh (besides the [wiki](https://github.com/tesseract-ocr/tesseract/wiki/Training-Tesseract-%E2%80%93-tesstrain.sh)), and they don't work with `./lstm.train`. So if I revert that change, how can we make sure users will change to the right directory, or copy lstm.train appropriately, before calling tesstrain.sh?

Shreeshrii:
If tesstutorial has `./lstm.train`, that is incorrect. It should be plain `lstm.train` and should be found via TESSDATA_PREFIX eg. in /usr/local/share/tessdata/configs/lstm.train

I think some blog had suggested copying `lstm.train` to where `tesseract` is.  It is quite possible that I tried it out at sometime. But I don't think that is necessary.

bertsky:
As far as I can see, this gets passed as CLI parameter to tesseract after the image path, the image output path base, and before the langdata config path (should it exist), i.e. as one of the `configs` passed down  `main()` - `TessBaseAPI::Init()` - `Tesseract::init_tesseract()` - `Tesseract::read_config_file`. So it can be both relative to the current working directory or `TESSDATA_PREFIX/configs` or `TESSDATA_PREFIX/tessconfigs`. 

It thus appears that README, tesstutorial and wiki are all failing to mention that `lstm.train` must be put there.

Shreeshrii:
>It thus appears that README, tesstutorial and wiki are all failing to mention that lstm.train must be put there.

Won't the makefile here take care of that?
https://github.com/tesseract-ocr/tesseract/blob/master/tessdata/configs/Makefile.am

bertsky:
> Won't the makefile here take care of that?

You are right, `make install` should copy the file along with the others into TESSDATA_PREFIX. 

The problem is that tesstutorial uses a clone of `tessdata_best` via `--tessdata_dir`, and README a clone of `tessdata`, which both obviously have no config files. And the [wiki](https://github.com/tesseract-ocr/tesseract/wiki/TrainingTesseract-4.00) does not mention `lstm.train` either.

Shreeshrii:
>The problem is that tesstutorial uses a clone of tessdata_best via --tessdata_dir, and README a clone of tessdata, which both obviously have no config files. And the wiki does not mention lstm.train either.

In my local clones of tessdata_best, tessdata_fast and tessdata repos, I have copied the configs and other files from tesseract/tessdata, hence the problems are not obvious to me. Thanks for bringing it to fore.

Shreeshrii:
I think that the PR can be merged. 
cc: @stweil @zdenop @amitdo

Shreeshrii:
@bertsky Please see issue https://github.com/tesseract-ocr/tesseract/issues/2159

Wiith ` tesseract digits.png stdout --psm 6 --dpi 300 -c lstm_choice_mode=2 hocr` I get the correct value `1` as one of the choices - choice_1_1_4.

```
       <span class='ocrx_cinfo' id='lstm_choices_1_1_1'>
       <span class='ocr_glyph' id='choice_1_1_1' title='x_confs 53'>]</span>
       <span class='ocr_glyph' id='choice_1_1_2' title='x_confs 14'>i</span>
       <span class='ocr_glyph' id='choice_1_1_3' title='x_confs 11'>j</span>
       <span class='ocr_glyph' id='choice_1_1_4' title='x_confs 5'>1</span>
       <span class='ocr_glyph' id='choice_1_1_5' title='x_confs 4'>l</span>
       <span class='ocr_glyph' id='choice_1_1_6' title='x_confs 3'>J</span>
       <span class='ocr_glyph' id='choice_1_1_7' title='x_confs 2'>|</span>
       <span class='ocr_glyph' id='choice_1_1_8' title='x_confs 1'>}</span>
       <span class='ocr_glyph' id='choice_1_1_9' title='x_confs 1'>)</span>
       <span class='ocr_glyph' id='choice_1_1_10' title='x_confs 0'>.</span>
       <span class='ocr_glyph' id='choice_1_1_11' title='x_confs 0'>I</span>
```
But using ` tesseract digits.png stdout --psm 6 --dpi 300 -c lstm_choice_mode=2 -c tessedit_char_whitelist="0123456789" hocr` OR `tesseract digits.png stdout --psm 6 --dpi 300 -c lstm_choice_mode=2 hocr digits` OR ` tesseract digits.png stdout --psm 6 --dpi 300  -c tessedit_char_whitelist="0123456789" hocr` gives NO result whatsoever.

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title></title>
  <meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
  <meta name='ocr-system' content='tesseract 4.1.0-rc1-131-g632a2' />
  <meta name='ocr-capabilities' content='ocr_page ocr_carea ocr_par ocr_line ocrx_word ocrp_wconf'/>
 </head>
 <body>
  <div class='ocr_page' id='page_1' title='image "digits.png"; bbox 0 0 9 13; ppageno 0'>
  </div>
 </body>
</html>
```

bertsky:
> Please see issue #2159

@Shreeshrii Thanks for pointing out, I now [have](https://github.com/tesseract-ocr/tesseract/issues/2159#issuecomment-475912191). Essentially, if only digits are allowed, no alternatives besides null appear in the beam, because it is too narrow. So whitelisting itself does work correctly, but the latter issue is wider. We have seen it surface with [user patterns](https://github.com/tesseract-ocr/tesseract/pull/2328#issuecomment-474724763), too. And of course it applies for [lattice output](https://github.com/tesseract-ocr/tesseract/issues/2339).

BTW, the new `lstm_choice_mode` you are using here together with hocr output does not at all give correct data (both symbols and scores). It ignores incremental character encoding (`UnicharCompress`) and duplicate coding (`NodeContinuation`). Moreoever, in setting 1 it bypasses CTC altogether (so the user would have to do the decoding), and in settings 2 and 3 it merely yields the alternative character outputs along the best CTC path – without actually aggregating scores for the other paths (so only the best scores can be trusted). 

Should I open a new issue about beam input narrowness, or can we discuss that in #2339?

Shreeshrii:
> the new lstm_choice_mode you are using here together with hocr output does not at all give correct data (both symbols and scores).

@bertsky  You should bring it to the attention of @noahmetzger, @stweil, @amitdo. I do not know enough about tesseract internals to provide useful input.

>settings 2 and 3 it merely yields the alternative character outputs along the best CTC path 

For using `whitelist` it maybe enough to filter the  alternative character outputs along the best CTC path. It might mean using a different algorithm rather than the one via the beam search.

bertsky:
> You should bring it to the attention of @noahmetzger, @stweil, @amitdo. I do not know enough about tesseract internals to provide useful input.

I did that already, we have been in touch via email and phone. 

I am going to prepare a special issue (to be referenced by the 4.1 planning RFC and planning wiki) focussing on the recurring big problems of the LSTMs: beam narrowness, confidence incompatibility/incommensurability with pre-LSTM models, and CTC diplopia. Hopefully, we can get some good discussion about how to go about these.

> For using `whitelist` it maybe enough to filter the alternative character outputs along the best CTC path. It might mean using a different algorithm rather than the one via the beam search.

Maybe you are right. But it would not work for choice iterator with correct confidences and for user patterns.

zdenop:
@bertsky @Shreeshrii : so what is status of this PR?

Shreeshrii:
@zdenop In my opinion, this can be merged because it does improve the results compared to current code. However, the related issues can still be left open since other changes are needed to fully fix the issue.

stweil:
The pull request includes unrelated modifications for `CONTRIBUTING.md` and `tesstrain_utils.sh`. What about those changes, especially the 2nd one (is it still needed)?

stweil:
I just fixed the merge conflicts in this pull request.

bertsky:
@stweil, thanks for wrapping up the changesets. 

> The pull request includes unrelated modifications for `CONTRIBUTING.md` and `tesstrain_utils.sh`. What about those changes, especially the 2nd one (is it still needed)?

These are fallout from attempting to fix unittests and lstmtutorial along with validating this PR, which I added by request (see conversation above). Yes, they are still needed, but could also be split off.

zdenop:
thanks to all participants!

Issue #2298, opened 2019-03-08
Issue 13590: tesseract-ocr/fuzzer-api: Heap-buffer-overflow in GenericVector<int>::size


stweil:
OSS Fuzz reports a security issue.

stweil:
This crash happens very often in the fuzzing test. It seems to be caused by Tesseract using the wrong unicharset, so tessdata_best, tessdata_fast and the initial revision of tessdata are not affected. It can occur with traineddata which includes different unicharsets for legacy and LSTM recognizers. The fuzzer uses tessdata/eng.traineddata where eng.lstm-unicharset has 112 entries while eng.unicharset has 113 entries. It tries to get entry 112 which only exists in the legacy unicharset.

Obviously PR #1954 did not fix all code locations.

Related open issue: #1222, #1307, #2301.

stweil:
Pull request #2305 fixed this issue.

PR #2305, opened 2019-03-10
Fix Heap-buffer-overflow in GenericVector<int>::size (issue #2298)

stweil:
Credit to OSS-Fuzz:
This fixes a security issue which was reported by OSS-Fuzz, see details at
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13590.

Add also some assertions to catch similar bugs.


stweil:
The link to the full bug report currently only works for registered users. All bugs found by OSS-Fuzz have a 90 day disclosure deadline (or 30 days after a bug fix was released).

PR #2309, opened 2019-03-11
 Fix several runtime errors (found by OSS-Fuzz)

stweil:


stweil:
The modifications in `intmatcher.cpp` avoid the runtime errors without changing the resulting values. That helps for now as it allows the fuzzing process to continue, but I am not sure whether it is the right fix.

stweil:
Pushed two more commits which fix runtime errors found by fuzzing.

PR #2316, opened 2019-03-13
`accumulated_timesteps` is not a pointer but a vector

vidiecan:
in case a ChoiceIterator is used without `lstm_choice_mode`, tesseract crashes (or similar) because the check is true and we reference not existing item


noahmetzger:
I'm sorry, my bad. Forget to check if it still runs without lstm_choice_mode :(

Issue #2313, opened 2019-03-12
tesstrain continues on text2image error


Shreeshrii:
```
Stripped 5 unrenderable words
Error in pixCreateNoInit: pix_malloc fail for data
Error in pixCreateTemplateNoInit: pixd not made
Error in pixGetData: pix not defined
Error in pixGetWpl: pix not defined
Error in composeRGBPixel: &pixel not defined
Rendered page 56 to file /tmp/tam-2019-03-12.PQP/tam.Noto_Sans_Tamil_Bold.exp0.tif
Stripped 1 unrenderable words
ERROR: Program text2image failed. Abort.
Rendered page 57 to file /tmp/tam-2019-03-12.PQP/tam.TAMu_Kadambri_Regular.exp0.tif
Stripped 50 unrenderable words
Rendered page 57 to file /tmp/tam-2019-03-12.PQP/tam.Noto_Sans_Tamil_Bold.exp0.tif
Stripped 3 unrenderable words
ERROR: Program text2image failed. Abort.
```

Even on error the bash script continues and dies only at end of phase:

```
Stripped 2 unrenderable words
Rendered page 117 to file /tmp/tam-2019-03-12.PQP/tam.e-Grantamil.exp0.tif
Rendered page 118 to file /tmp/tam-2019-03-12.PQP/tam.e-Grantamil.exp0.tif
Rendered page 119 to file /tmp/tam-2019-03-12.PQP/tam.e-Grantamil.exp0.tif
ERROR: /tmp/tam-2019-03-12.PQP/tam.Arial_Unicode_MS.exp0.box does not exist or is not readable
```

*Expected* - End script at the first error itself so that problem can be investigated early.

zdenop:
Please provide instruction/test case how to generate this problem...

Shreeshrii:
I updated to latest code from master branch without any additional PRs. Now  I get the following error:

```
Rendered page 27 to file /tmp/tam-2019-03-12.p1P/tam.Arial_Unicode_MS.exp0.tif
Rendered page 26 to file /tmp/tam-2019-03-12.p1P/tam.Lohit_Tamil_Regular.exp0.tif
Stripped 9 unrenderable words
Rendered page 27 to file /tmp/tam-2019-03-12.p1P/tam.FreeSerif.exp0.tif
Stripped 4 unrenderable words
[...]

PR #2335, opened 2019-03-17
Changes to tesstrain.py - max_workers=8, distort_image=false

Shreeshrii:


Shreeshrii:
Suggestion from @jbarlow83  - see https://github.com/tesseract-ocr/tesseract/issues/2313#issuecomment-473593598

mohdbm:
I have an issue of ERROR: Program text2image failed. Abort. while generating the training data for Arabic fonts
any ideas on how can I overcome this...

PR #2319, opened 2019-03-14
tesstrain: check failure of subjobs

bertsky:
in an attempt to fix #2313

PR #2341, opened 2019-03-22
Fix

Shreeshrii:


src/ccstruct/publictypes.h
```diff
@@ -161,22 +161,22 @@ enum TextlineOrder {
  * so that the inequality test macros below work.
 */
 enum PageSegMode {
-  PSM_OSD_ONLY,       ///< Orientation and script detection only.
-  PSM_AUTO_OSD,       ///< Automatic page segmentation with orientation and
+  PSM_OSD_ONLY,       ///< 0 - Orientation and script detection only.
```
stweil:
```suggestion
  PSM_OSD_ONLY = 0,       ///< Orientation and script detection only.
```

stweil: I suggest to add the values to the definition, not in the comment.

Shreeshrii: Done via 259d5af6b1d9d193053ec579e42a2c6a5fcb9b63

src/training/tesstrain_utils.sh
```diff
@@ -164,6 +165,9 @@ parse_flags() {
             --my_boxtiff_dir)
                 parse_value "MY_BOXTIFF_DIR" ${ARGV[$j]:-}
                 i=$j ;;
+            --distort_image)
```
stweil: Isn't `--distort_image` a boolean argument? Then it should not expect an additional value and simply assign `DISTORT_IMAGE=true`.

Shreeshrii: tesstrain.sh uses 0 and 1 for all the boolean arguments so I have changed it to `DISTORT_IMAGE=1`.

Shreeshrii: Should all the booleans be changed to use `=true` and `=false` rather than `=1` and `=0`?

```
SAVE_BOX_TIFF=0
DISTORT_IMAGE=0
OVERWRITE=0
LINEDATA=0
RUN_SHAPE_CLUSTERING=0
EXTRACT_FONT_PROPERTIES=1
```

zdenop: `=true` and `=false` is more clear than `=1` and `=0` (which can somebody unterstand as integer and not boolean )

Shreeshrii: Done.

Also sorted the list of arguments alphabetically and changed the following defaults to match those required for LSTM training.

```
EXTRACT_FONT_PROPERTIES=false
LINEDATA=true
```

Shreeshrii: Codacy error:

LINEDATA appears unused. Verify use (or export if used externally).
LINEDATA=true ;;

It is used in tesstrain.sh 

https://github.com/tesseract-ocr/tesseract/blob/master/src/training/tesstrain.sh#L84

Shreeshrii:
I had not tested the change from 0/1 to false/true. The script is failing now.

stweil:
Please don't revert. I'll send a fix.

stweil:
@Shreeshrii, please test latest Git master.

Shreeshrii:
@stweil Thanks. That works now. 

https://github.com/tesseract-ocr/tesseract/pull/2319 should also be merged.

PR #2345, opened 2019-03-24
Fix compiler warnings caused by ASSERT_HOST

stweil:
The modified definition avoids warnings caused by redundant semicolons.
Now a semicolon is required when using the macro, so a few code locations
had to be updated.


zdenop:
thanks

PR #2348, opened 2019-03-26
Modernize C++ code

stweil:


stweil:
This is still far from being complete, but at least a next step.

zdenop:
thanks!

src/arch/intsimdmatrixavx2.cpp
```diff
@@ -85,7 +85,7 @@ static inline void ExtractResults(__m256i& result, __m256i& shift_id,
                                   const int8_t*& wi, const double*& scales,
                                   int num_out, double*& v) {
   for (int out = 0; out < num_out; ++out) {
-    int32_t res =
+    auto res =
```
amitdo: Not sure if the use of `auto` is clear in this case.

stweil: Maybe it would be a little bit clearer if left and right side of the assignment were on the same line of code.

amitdo: The type casting to `int32_t` is 6 lines beneath the `auto` It's too far...

amitdo: You might want to use this check:
https://clang.llvm.org/extra/clang-tidy/checks/google-readability-casting.html

stweil: It fixes a lot of old type casts. I now committed a small selection (see commit 36a1a30c223c6397bb4859601ee3c6fe2304cadc, the others need more manual fixes) and also changed the [code for intsimdmatrixavx2.cpp](https://github.com/tesseract-ocr/tesseract/blob/master/src/arch/intsimdmatrixavx2.cpp#L88).

PR #2362, opened 2019-03-31
Remove unused config variable tessedit_single_match

stweil:
Replace also TRUE, FALSE by true, false.

### DIFF
--- a/src/classify/classify.cpp
+++ b/src/classify/classify.cpp
@@ -60,10 +60,8 @@ namespace tesseract {
 Classify::Classify()
     : BOOL_MEMBER(allow_blob_division, true, "Use divisible blobs chopping",
                   this->params()),
-      BOOL_MEMBER(prioritize_division, FALSE,
+      BOOL_MEMBER(prioritize_division, false,
                   "Prioritize blob division over chopping", this->params()),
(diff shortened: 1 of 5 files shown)


stweil:
This is an API change, but I think it is uncritical because the config variable was unused.

zdenop:
Thanks

stweil:
@egorpugin, Appveyor CI is currently broken. Do you have an idea how to fix that?

stweil:
@zdenop, there is another config variable `tessedit_module_name` which was added in commit 4523ce9f7d9a46faff9e8bca5432293bdd3a968b (only for Windows) but is unused in the Tesseract code.

Can it be removed, too? Then we could also get rid of the macro `WINDLLNAME`.

zdenop:
yes, remove it.

stweil:
Removed in commits 77a5f2623ebf9b0a3df51838c07ec1a3d9de42a9 and ab009fae9436bb41723a5a34d3fcb5b5b3aa708a.

egorpugin:
Yes, I'm working on some sw updates, so build is unstable.
(I've did some fixes, one build was green, then I made more changes and it's broken again.) :)

PR #2366, opened 2019-04-03
Modernize code with "using"

stweil:


src/classify/adaptive.h
```diff
@@ -69,49 +63,44 @@ typedef struct
   LIST TempProtos;
   ADAPTED_CONFIG Config[MAX_NUM_CONFIGS];
 } ADAPT_CLASS_STRUCT;
-typedef ADAPT_CLASS_STRUCT *ADAPT_CLASS;
+using ADAPT_CLASS = ADAPT_CLASS_STRUCT*;
 
-typedef struct
-{
+typedef struct {
```
egorpugin: `typedef struct` is a C thing.
Probably we could replace them with just `struct`.

src/arch/simddetect.h
```diff
@@ -22,7 +22,7 @@
 namespace tesseract {
 
 // Function pointer for best calculation of dot product.
-typedef double (*DotProductFunction)(const double* u, const double* v, int n);
+using DotProductFunction = double (*)(const double*, const double*, int);
```
amitdo: The fix drops the var names, is it ok?

src/classify/adaptive.h
stweil: The main focus of this PR is `modernize-use-using`. Yes, those `typedef struct` statements should be replaced by C++ `struct` or `class` statements where possible (not in `capi.h`), That's one of the many things left for future commits.

src/arch/simddetect.h
stweil: Technically it is fine, but nevertheless that is a side effect of `clang-tidy` which I did not intend. Maybe `clang-tidy` can be told to keep the names for the function parameters, or I have to add them manually again if we prefer to have those names.

Personally I want the names at least in public API header files.

amitdo: Yes, I agree.

amitdo:
>@egorpugin on 4 Apr 2019
>
>typedef struct is a C thing.
Probably we could replace them with just struct.

https://github.com/tesseract-ocr/tesseract/search?q=typedef+struct

Egor, I'm still waiting for your PR (or git push).
:-)

egorpugin:
Done https://github.com/tesseract-ocr/tesseract/commit/efd17e205a6617f3729ac04a60c854fa28485014

PR #2368, opened 2019-04-04
disable-legacy build: Do not include unused headers

amitdo:


zdenop:
Is there a preferred way how 3rd party projects can check whether `disable-legacy` was used for tesseract build:
1. during build/configure of project?
2. at runtime?

This question is not related only to this build options, but IMO should be answered during committing this kind of changes.

amitdo:
There isn't, but I think the situation is not different than disabling graphics/openmp/opencl.

zdenop:
Well, not exactly: `disable-legacy` has influence on API (baseapi.h) the others not (which does not mean there is no need for improvement).

amitdo:
yes, you're right.

PR #2383, opened 2019-04-10
Fix for MSVC

prodeveloper0:
LoadDataFromFile/SaveDataToFile use fopen with unsupported file mode 'e' in MSVC.


egorpugin:
Thank you!

PR #2394, opened 2019-04-18
unittest: Fix and enable params_model_test

stweil:
This needs the latest test submodule.

The test uses LoadFromFile which is not used otherwise, so remove that
function from class ParamsModel.


egorpugin:
FYI: appveyor will be broken again couple of days.

PR #2396, opened 2019-04-19
 Replace sscanf by std::stringstream

stweil:
Using std::stringstream allows working with the C locale, independent
of the current locale settings.


zdenop:
thanks!

PR #2400, opened 2019-04-20
Replace more snprintf by std::stringrenderer

stweil:
This makes more code independent of the locale settings.


stweil:
More code locations which depend on the locale settings can be found with `git grep "%[0-9.]*[fg]"`. Many of those lines are calls of `tprintf`. I don't think that such `tprintf` calls must be fixed to be independent of the locale settings as they are not part of the external Tesseract API.

zdenop:
thanks!

Issue #2390, opened 2019-04-15
Tesseract min and max macro


Bleach665:
Windows 10 x64, CMake 3.14, VS 2017, Tesseract rev 332a1a980970b71bff1d97058c5ae926318492a3.
Tesseract redefine min and mac macro, therefore after including baseapi.h we receive errors like

````
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(581): error C2589: '(': illegal token on right side of '::'
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(581): error C2062: type 'unknown-type' unexpected
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(581): error C2059: syntax error: ')'
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(582): warning C4003: not enough arguments for function-like macro invocation 'max'
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(582): error C2589: '(': illegal token on right side of '::'
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(582): error C2062: type 'unknown-type' unexpected
1>e:\0_localwork\improcqcv\improcqcv\imgcoreqcv.cpp(582): error C2059: syntax error: ')'
````

when we use default std min and max macro. 
Also this errors received when other lib (opencv) was build with Tesseract.
Adding 
````
#undef min
#undef max
````
to the end of baseapi.h fix it.
Create appropriate pull request?

stweil:
It's not Tesseract, but the Windows header files which define `min` and `max` macros. We have to fix the public Tesseract headers to avoid including `windows.h`.

zdenop:
@Bleach665: Errors comes from your code... Can you provide us test code that demonstrate this problem?

Bleach665:
@zdenop. Just try to use std min or max macro after tesseract/baseapi.h.
`std::min(2, 3);`

Bleach665:
@zdenop. Just try to use std min or max after include tesseract/baseapi.h

zdenop:
try to add `#include <tesseract/host.h>` after `#include <tesseract/baseapi.h>`

stweil:
That might work because `host.h` undefines both macros, but I consider that a hack.

zdenop:
[...]

PR #2401, opened 2019-04-21
Don't include windows.h from platform.h

stweil:
This partially reverts commit c150b9832d7d07761ec7869a625e6c46ba62206a.
Now params.cpp includes host.h which also gets the definition for MAX_PATH.

### DIFF
--- a/src/ccutil/params.cpp
+++ b/src/ccutil/params.cpp
@@ -25,9 +25,9 @@
 #include <sstream>          // for std::stringstream
 
 #include "genericvector.h"
+#include "host.h"           // platform.h, windows.h for MAX_PATH
 #include "tprintf.h"
 #include "params.h"
-#include "platform.h"  // MAX_PATH
 
 #define PLUS          '+'        //flag states
 #define MINUS         '-'
--- a/src/ccutil/platform.h
+++ b/src/ccutil/platform.h
@@ -20,7 +20,6 @@
 
 #define DLLSYM
 #ifdef _WIN32
-#  include <windows.h>   // MAX_PATH
 #  ifndef NOMINMAX
 #    define NOMINMAX
 #  endif /* NOMINMAX */


stweil:
This addresses issue #2390.

zdenop:
Do you think it there reason to keep `host.h` and `platform.h` separated?  
`host.h` is used only in few places:
```
src/ccmain/paragraphs.cpp:#include "host.h"                 // for NearlyEqual
src/ccmain/recogtraining.cpp:#include "host.h"  // for NearlyEqual
src/ccstruct/blobbox.cpp:#include "host.h"        // for NearlyEqual
src/ccstruct/blobbox.h:#include "host.h"              // for PRId32
src/ccstruct/boxword.cpp:#include "host.h"       // for NearlyEqual
src/ccstruct/imagedata.cpp:#include "host.h"        // for PRId64
src/ccstruct/ocrpara.cpp:#include "host.h"  // For NearlyEqual()
src/ccstruct/pdblock.cpp:#include "host.h"  // for PRId32
src/ccstruct/polyblk.cpp:#include "host.h"  // for PRId32
src/dict/dawg.h:#include "host.h"               // for PRId64
```

stweil:
> Do you think it there reason to keep host.h and platform.h separated?

Yes, see [my comment](https://github.com/tesseract-ocr/tesseract/issues/2390#issuecomment-485280003).

The more difficult question is whether we may remove host.h from the 4.1 API.

zdenop:
Do not worry about that - I can ran `abi-compliance-checker` locally so I will backport all possible changes from master.

Also I found [nice extension for Visual studio](https://marketplace.visualstudio.com/items?itemName=Wumpf.IncludeToolbox) that eliminate not needed  includes (which is based on [include-what-you-use](https://include-what-you-use.org/)), so maybe we can better optimize includes.

PR #2403, opened 2019-04-22
Remove host.h from public API and clean platform.h

stweil:


stweil:
Maybe `M_PI` and `MAX_PATH` should be removed from `platform.h`, too.

zdenop:
So public/API part will be in `platform.h` and internal part will be in `host.h`?

PR #2418, opened 2019-05-01
Backport changes from master to 4.1 branch

stweil:


stweil:
@zdenop, those changes should be API compatible. My local API compatibility test does not work (always shows 100 %), so maybe you can test it in your setting.

Can we also apply commit db08786d2d478b2b457a15350e0dabdfc7f97fa1? It removes `host.h` from the API.

zdenop:
I am just testing it (removal of host.h).

zdenop:
Removal of host.h should be ok according my tests.

zdenop:
thanks will you backport also rest of changes?

PR #2422, opened 2019-05-02
tesscallback: Remove unused code

stweil:


stweil:
`tesscallback.h` is part of the public API. It is the 2nd largest source file of Tesseract.

It looks like a generated file, but I have no information how it was built. Nor do I know why it was made at all. Maybe we can get rid of it.

zdenop:
thanks.

zdenop:
Seems like no problem for API/ABI checker compatibility for the moment...

PR #2423, opened 2019-05-07
Fix CPPFLAGS configuration for icu4c and libarchive

jbarlow83:
./configure did not set CPPFLAGS correctly on macOS due to problems with how CPPFLAGS were being set.

This caused build issues like missing `<archive.h>`


zdenop:
thanks!

PR #2430, opened 2019-05-12
Fix reading of parameter from traineddata normproto component and make function independent of locale

stweil:
The NonEssential parameter was wrongly derived from linear_token instead
of essential_token and therefore always set to true.


zdenop:
thanks

PR #2432, opened 2019-05-14
Add different classes to hocr output depending on BlockType

nickjwhite:
These classes are taken from the hOCR specification, and seem
to map well onto the BlockType types. There are probably more that
could be added.

Note that I haven't read the hOCR specification closely, so there's a
chance I misunderstood something. @kba, can you take a look and
check that the hOCR produced is valid and sensible?


zdenop:
@nickjwhite : thanks. Can have a look at issue 2303 if you are able to fix it?

stweil:
@nickjwhite, @zdenop, I just updated branch [4.1](https://github.com/tesseract-ocr/tesseract/tree/4.1). Would you suggest to add PR #2432 to that branch, too?

zdenop:
Yes, all commits that does not effect API should be cherry-picked to 4.1 branch

stweil:
I updated 4.1 accordingly, thank you.

PR #2434, opened 2019-05-14
configure: Fix for latest developer tools on macOS

stweil:
AX_CHECK_COMPILE_FLAG fails if it is used with -Werror and the compiler
raises error -Wunused-macros. Add -Wno-unused-macros to disable those
errors if possible.

Simplify also the setting of several conditionals (AVX, AVX2, ...).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -111,31 +111,24 @@ case "${host_os}" in
         ;;
 esac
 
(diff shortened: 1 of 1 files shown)


zdenop:
thanks.

PR #2441, opened 2019-05-16
Fix unittest build without legacy code and use locale for most unittests

stweil:


amitdo:
* intfeaturemap_test.cc
* mastertrainer_test.cc
* osd_test.cc
* params_model_test.cc
* shapetable_test.cc

Why not to conditionaly disable them in the build system?

stweil:
> Why not to conditionaly disable them in the build system?

That would be possible, too. Initially I was not sure whether some tests would remain usable even without the legacy code. For tests which cannot be run there is always the choice to silently ignore them or to report that they were skipped, no matter why they cannot be run. I thought that explicitly saying that they were skipped is better.

PR #2448, opened 2019-05-20
Remove local definition of M_PI

stweil:
It is defined for all platforms when math.h or cmath is included
after defining the macro _USE_MATH_DEFINES.

Define _USE_MATH_DEFINES before any include statement to make sure
that M_PI gets defined. It is not necessary to define it conditionally
only for Windows.


zdenop:
thanks.

PR #2453, opened 2019-05-21
Remove SavePixForCrash and related code

stweil:
That debugging code uses very much memory and is no longer useful.

     815	      0	 262144	 262959	  4032f	src/ccutil/globaloc.o

Remove also the function err_exit which was only used in ccmain/reject.cpp.


stweil:
I never used that code as there are other ways to analyze crashes. Did anyone else need it?

PR #2464, opened 2019-05-25
Do not link librt on OpenBSD

rrrapha:
This prevents a link error on OpenBSD:
```
warning: could not find a rt library
Link error: rt not found!
```


zdenop:
thanks.

PR #2474, opened 2019-05-31
Use C++17 compiler if possible

stweil:
This allows using new features of C++17 conditionally.
Simplify also the code which checks and sets the C++ version.


stweil:
C++-17 is required for [std::transform_reduce](https://en.cppreference.com/w/cpp/algorithm/transform_reduce) which might be useful to calculate the dot product on certain hosts.

PR #2475, opened 2019-05-31
Add dot product implementation using std::inner_product

stweil:


stweil:
Instead of a local implementation this code uses the dot product implementation of the standard C++ library. In the future `std::transform_reduce` (a parallelized computation) could also be used similarly (requires C++17 library).

zdenop:
thanks.

PR #2476, opened 2019-05-31
Detect and enable C++17 support for CMake builds

ismail:
This commit is the cmake mirror of commit ca885da5d361e5b59d7f9ae254f7d410b8c6a85a

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,9 @@ endif()
 
 # Check for C++ standard to use
 get_property(known_features GLOBAL PROPERTY CMAKE_CXX_KNOWN_FEATURES)
-if(cxx_std_14 IN_LIST known_features)
+if(cxx_std_17 IN_LIST known_features)
+    set(CMAKE_CXX_STANDARD 17)
+elseif(cxx_std_14 IN_LIST known_features)
     set(CMAKE_CXX_STANDARD 14)
 else()  # minimum required standard
     set(CMAKE_CXX_STANDARD 11)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,8 +131,6 @@ if (CYGWIN)
 endif()
 
 if (UNIX)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --std=c++11")
-
     if (NOT ANDROID)
         set(LIB_pthread pthread)
     endif()


stweil:
Thank you!

stweil:
@ismail, Travis CI on Linux has [problems with the 2nd commit](https://travis-ci.org/tesseract-ocr/tesseract/jobs/539678818) and fails now in Git master.

ismail:
> @ismail, Travis CI on Linux has [problems with the 2nd commit](https://travis-ci.org/tesseract-ocr/tesseract/jobs/539678818) and fails now in Git master.

gcc 4.8 is just too old, is it supposed to be supported?

zdenop:
> gcc 4.8 is just too old...

IMO this is not right question: cmake should check for the highest available c++ standard supported by compiler and use it. We require c++11 so any compiler meeting this requirement should work/be supported. Personally I am fan of support "old" sw as long as it make sense (e.g. at the moment we there is minimum code requiring c++17, so there is not sense to drop support because of it).

> We need cmake 3.8
1. I tested the master repo on raspberry PI where is cmake 3.7.2 and gcc 6.3.0 and it choose C++11 for compilations. So I am afraid problem is somewhere else...
2. We should use sw default version available in distribution. E.g. for travis build we Ubuntu Trusty  14.04.3 which has [end of life in 2022](https://wiki.ubuntu.com/Releases). Maybe we can require Ubuntu Xenial, but I have no clue, what is usage of Trusty or Xenial... just because of c++17 it does not make sense for me at the moment.

zdenop:
PR #2478 should fix it. Let's wait for travis...

stweil:
PR #2478 looks complicated. A simpler solution is to revert the commit which caused the regression for now – until we use a newer Ubuntu for Travis CI. I did that for Git master to fix CI.

zdenop:
@stweil: PR#2478 is just my test to find out real problem and real solution. It will not be merged ;-)

zdenop:
@stweil: I do not think it is correct to use c++11 for all unix build with cmake (you already committed c++14 code). Especially if we know that problem is with old g++ version (e.g. we do not know status for clang version).  I propose solution in #2479, because [gcc 6.1 and above should fully support c++14](https://gcc.gnu.org/projects/cxx-status.html#cxx14) .

c++17 is partly supported in several g++ version, so IMO checking of needed features would be more safe that testing '-std=c++17' - and this is valid for automake tools too.

ismail:
FWIW #2479 looks sane to me.

PR #2477, opened 2019-05-31
Fix OpenMP builds on Unix

ismail:
First of all do not use -Xclang -fopenmp, just use -fopenmp as this feature is stable so long now. Also on UNIX you'll need to link to pthread to get OpenMP link correctly.


ismail:
CI fails seems to be due to other changes in the master.

PR #2479, opened 2019-06-01
cmake: using c++ standard for g++

zdenop:


stweil:
Thank you for fixing this.

PR #2492, opened 2019-06-13
Fix text2image compilation on C++17 compilers

jbarlow83:
C++17 drops support for `std::random_shuffle`, breaking C++17 compilers that run to compile text2image.cpp. std::shuffle is valid on C++11 through C++17, so use std::shuffle instead.

### DIFF
--- a/src/training/text2image.cpp
+++ b/src/training/text2image.cpp
@@ -31,6 +31,7 @@
 #include <algorithm>
 #include <iostream>
 #include <map>
+#include <random>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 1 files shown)


391861737:
I've encountered the same issue and did exactly the same fix. But there might be one more thing that should be added to this PR. 
Since `std::shuffle` is introduced in C++11, there should be a MACRO guarding the original lines which are to work under versions before C++11.

391861737:
And I just intended to propose a PR to this issue before I went into yours.

jbarlow83:
I believe C++11 is the baseline requirement for Tesseract now.

zdenop:
Thanks! Yes [c++11](https://github.com/tesseract-ocr/tesseract/blob/3c507100c6e6a7a614a88cbc30584985bfe5c362/configure.ac#L359) is must for compiling tesseract >4.

zdenop:
@jbarlow83: can you please check also 4.1 branch?

PR #2493, opened 2019-06-14
[4.1] Fix text2image compilation on C++17 compilers

jbarlow83:
C++17 drops support for `std::random_shuffle`, breaking C++17 compilers
that run to compile text2image.cpp. std::shuffle is valid on C++11
through C++17, so use std::shuffle instead.

Due to the use `std::random_shuffle`, `text2image --render_ngrams`
would not give consistent results for different compilers or platforms.
With the current change, the same random number generator is used for
all platforms and initialized to the same seed, so training output
should be consistent.

### DIFF
--- a/src/training/text2image.cpp
+++ b/src/training/text2image.cpp
@@ -33,6 +33,7 @@
 #include <algorithm>
 #include <iostream>
 #include <map>
+#include <random>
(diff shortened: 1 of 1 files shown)


zdenop:
thanks

PR #2499, opened 2019-06-17
Remove code for embedded build

stweil:
That code is unrelated to Tesseract and can be easily implemented
by external projects which require it.


zdenop:
thanks

PR #2510, opened 2019-06-21
Simplify indirect call of LMPainPoints::GeneratePainPoint

stweil:
It does neither need a temporary TessResultCallback2 nor the function
LMPainPoints::GenerateForBlamer.

This also allows removing more code from tesscallback.h.


zdenop:
thanks.

PR #2514, opened 2019-06-22
Move LSTMTrainer from libtesseract to libtesseract_training

stweil:
LSTMTrainer is only used for training, so the shared library for
Tesseract can be made smaller.


egorpugin:
Seems `sw.cpp` must be changed also.

stweil:
Yes, indeed. I missed that file. Now the pull request was updated, so let's see whether the Windows build works again.

egorpugin:
kMinCertainty is missing from recodebeam.cpp/.h
These files are completely missing from sw.cpp.

stweil:
It is part of libtesseract which is included using a wildcard pattern.

egorpugin:
Oh, yes. I mean that symbol is not public - not `TESS_API`.

stweil:
Rewriting `src/lstm/recodebeam.*` simplified the code and fixed that.

PR #2531, opened 2019-07-01
add natvis file for visualstudio debugger.

### LINKED ISSUE #2530: add natvis file for visualstuduo user
### Expected Behavior:

STRING and GenericVector can debug user friendly.

ref: https://github.com/g-truc/glm/blob/master/util/glm.natvis
ref: https://docs.microsoft.com/en-us/visualstudio/debugger/create-custom-views-of-native-objects?view=vs-2019

zhuangzhuang:
fix #2530


util/tesseract.natvis
```diff
@@ -0,0 +1,18 @@
...
+    <Expand>
+      <Item Name="[size]" ExcludeView="simple">size_used_</Item>
+      <Item Name="[capacity]" ExcludeView="simple">size_reserved_</Item>
+      <ArrayItems>
+        <Size>size_used_</Size>
+        <ValuePointer>data_</ValuePointer>
+      </ArrayItems>
+    </Expand>
+  </Type>
+</AutoVisualizer>
```
stweil: Please add a terminating line feed.

stweil:
You added a new directory `util` for the new file, but as far as I have understood Visual Studio allows any directory and filename. I'd prefer using an existing directory (maybe `doc/tesseract.natvis`) or having an invisible `.tesseract.natvis` in the root directory.

stweil:
The longterm solution would be replacing proprietary data types by standard data types: `std::vector` instead of `GenericVector` and `std::string` instead of `STRING`.

stweil:
Could you please also add links to the .natvis documentation either in the commit message or as a comment in the .natvis file (or to both)? That helps maintaining that file.

egorpugin:
> The longterm solution would be replacing proprietary data types by standard data types: std::vector instead of GenericVector and std::string instead of STRING.

Can we do this for 5.0.0?
cc: @zdenop

util/tesseract.natvis
zhuangzhuang: has updated

egorpugin:
Ideal place for natvis is `src/tesseract.natvis`.

zhuangzhuang:
> Ideal place for natvis is `src/tesseract.natvis`.
There is no standard, refer to other projects.
imgui place in misc folder  https://github.com/ocornut/imgui/tree/master/misc/natvis
gsl place in root folder   https://github.com/microsoft/GSL
eigen place in debug folder  https://bitbucket.org/eigen/eigen/src/default/debug/msvc/

stweil:
Thank you. I merged it now. We can move the file later as soon as we find a better place for it.

stweil:
@zhuangzhuang, please [set your real name](https://help.github.com/en/articles/setting-your-username-in-git) in your Git configuration for future pull requests.

PR #2538, opened 2019-07-04
fix read wrong tiff page.

### LINKED ISSUE #2537: Issue using "-c tessedit_page_number=1" to process only second page of a multipage tiff G3/G4 compression
### Environment

* **Tesseract Version**: 
tesseract 4.0.0
 leptonica-1.76.0 (May 30 2019, 11:18:56) [MSC v.1916 LIB Release x86]
  libgif 5.1.4 : libjpeg 6b (libjpeg-turbo 2.0.1) : libpng 1.6.37 : libtiff 4.0.10

* **Platform**: 
Windows 10 Enterprise 1903 10.0.18362.175 Client

### Current Behavior:
I am using tesseract to extract text from a multi page tiff image (G3 or G4), but I only want to process the second page. I am using the following command:

tesseract.exe FILE.TIF OUT --tessdata-dir ."\tessdata" -l eng --psm 1 --oem 1 -c tessedit_page_number=1

For some reason it always processes the first page no matter what page number I put in the option. If I remove that option, it processes all the pages. I am attaching a zipped sample tiff file that reproduces the issue.

### Expected Behavior:
I would expect the page I am selecting in the option to be processed.
[18.zip](https://github.com/tesseract-ocr/tesseract/files/3355049/18.zip)

zhuangzhuang:
fix #2537

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1034,11 +1034,22 @@ bool TessBaseAPI::ProcessPagesMultipageTiff(const l_uint8 *data,
   int page = (tessedit_page_number >= 0) ? tessedit_page_number : 0;
   size_t offset = 0;
   for (; ; ++page) {
-    if (tessedit_page_number >= 0)
+    if (tessedit_page_number >= 0) {
       page = tessedit_page_number;
-    pix = (data) ? pixReadMemFromMultipageTiff(data, size, &offset)
-                 : pixReadFromMultipageTiff(filename, &offset);
-    if (pix == nullptr) break;
+      int pages_read = 0;
+      do {
+        pix = (data) ? pixReadMemFromMultipageTiff(data, size, &offset)
+                     : pixReadFromMultipageTiff(filename, &offset);			  
+        pages_read++;
+      } while (pix != nullptr && pages_read < (page + 1));
+    }
+    else {
+      pix = (data) ? pixReadMemFromMultipageTiff(data, size, &offset)
+                   : pixReadFromMultipageTiff(filename, &offset);		  
+    }    
+    if (pix == nullptr) {
+      break;
+    }
     tprintf("Page %d\n", page + 1);
     char page_str[kMaxIntSize];
     snprintf(page_str, kMaxIntSize - 1, "%d", page);


src/api/baseapi.cpp
```diff
@@ -1034,11 +1034,22 @@ bool TessBaseAPI::ProcessPagesMultipageTiff(const l_uint8 *data,
...
-                 : pixReadFromMultipageTiff(filename, &offset);
-    if (pix == nullptr) break;
+      int pages_read = 0;
+      do {
+        pix = (data) ? pixReadMemFromMultipageTiff(data, size, &offset)
+                     : pixReadFromMultipageTiff(filename, &offset);			  
+        pages_read++;
+      } while (pix != nullptr && pages_read < (page + 1));
+    }
+    else {
```
stweil: This is not the Tesseract coding style.

egorpugin: There could be set automatic style check based on clang format. Many projects have such thing.

stweil: While thinking more about that code and after reading the discussion on issue #233, I now think more changes would be reasonable. The new code loops reading a pix for each page until it reaches the desired page. It won't stop the loop at the end of the TIFF file if the user gave a too large page number, but starts again with reading the first page. The read data is never released, so it is easy to get out of memory:

    $ time bin/debug/x86_64-linux-gnu/src/api/tesseract issue/2537/18.TIF - -c tessedit_page_number=1000000
    Error in pixCreateNoInit: pix_malloc fail for data
    Error in pixCreate: pixd not made
    Error in pixReadFromTiffStream: pix not made

    real    1m26,521s
    user    0m26,758s
    sys     0m13,830s

In the past, `pixReadTiff` was used to read the n-th image. I think that is better than calling `pixReadFromMultipageTiff` in a loop, but I am not sure why it was replaced for the case of a single image (for handling all or a sequence of images, `pixReadFromMultipageTiff` is superior).

zhuangzhuang: > This is not the Tesseract coding style.

i will careful when next commit.

PR #2539, opened 2019-07-04
Replace tesscallback.h and related proprietary data types by C++-11 functionals

stweil:
The C++-11 functionals (using std::function and std::bind) do the same in a more elegant way.


egorpugin:
If we have C++14 lambdas would be better than bind which is not recommended at all.

stweil:
Thank you for this information. So would you suggest to drop C++-11 and require C++-14 support for Git master now and use lambdas? Or can we use bind now and replace it at some time later?

egorpugin:
It's better to use lambdas starting with this PR.
iif C++14 allowed. 
@zdenop said it's not time for C++17, but what about requiring C++14?

src/api/baseapi.h
```diff
@@ -77,8 +78,8 @@ using DictFunc = int (Dict::*)(void *, const UNICHARSET &, UNICHAR_ID, bool) con
 using ProbabilityInContextFunc = double (Dict::*)(const char *, const char *, int, const char *, int);
 using ParamsModelClassifyFunc = float (Dict::*)(const char *, void *);
 using FillLatticeFunc = void (Wordrec::*)(const MATRIX &, const WERD_CHOICE_LIST &, const UNICHARSET &, BlamerBundle *);
-typedef TessCallback4<const UNICHARSET &, int, PageIterator *, Pix *>
-    TruthCallback;
+typedef std::function<void(const UNICHARSET&, int, PageIterator*, Pix*)>
```
egorpugin: using = ...

stweil: This is fixed now.

zdenop:
I believe we can require c++14 in master (a.k.a tesseract 5.0.0). On other hand we should keep support for widely used compilers (e.g. VS2015  or default compilers in LTS Debian and Ubuntu or Mac OS; I have no clue about IOS, Android compiler options).
4.1.0 we can keep as c++11...
Maybe it would make sense to make overview of compilers version used on different platforms and check supported features.

stweil:
I suggest to merge this pull request as it is for now because the new code is rather close to the old code.

@egorpugin, I still have not figured out how to use lambdas instead of std::bind. Do you have a URL which explains why std::bind should not be used and how it can be replaced by lambdas?

I'd wait with dropping C++-11 in Git master until we really have new code which requires it.

zdenop:
Thanks. Agreed.

egorpugin:
Google for `c++ lambda instead of bind` or `c++ lambda vs bind`.

Also from there: https://stackoverflow.com/questions/17363003/why-use-stdbind-over-lambdas-in-c14

zdenop:
[travis linux build](https://travis-ci.org/tesseract-ocr/tesseract/builds/554595277) was failing from this commit... It used Ubuntu Trusty with g++ (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4  
I changed it to Xenial as [Trusty end of life was April, 2019](https://endoflife.software/operating-systems/linux/ubuntu). 
Now we should try to switch/require c++14...

PR #2542, opened 2019-07-04
Fix handling of single pages from multipage TIFF files (issue #2537) 

stweil:


stweil:
Travis CI failed in `apt-get`, so that failure is unrelated to my code change.

zdenop:
Just quick note: `pixReadMemFromMultipageTiff` was implemented because of  #233 with "[This patch reduces multipage TIFF seeks from O(n^3) to O(n)](https://github.com/tesseract-ocr/tesseract/issues/233#issuecomment-246744997)" and improved with https://github.com/stweil/tesseract/commit/54fafc4e2e9e2941b643f6cef67a7ec7e0b8bb49. 

So now we are back and #233 should be rechecked...

stweil:
Tesseract still uses `pixReadMemFromMultipageTiff` when all pages are processed (that was the problem reported in #233), so this use case did not change.

PR #2543, opened 2019-07-05
Fix crash in Tesseract::classify_word_and_language()

db4:
This
https://github.com/tesseract-ocr/tesseract/blob/d8494f3215c107308467c961fec05e1bf04dbe0e/src/ccmain/control.cpp#L1394
invalidates `word` pointer


zdenop:
thanks.

stweil:
The same code is also used in Git master, so ideally the pull request should have targeted Git master instead of the 4.1 branch. Should I cherry-pick the commit from 4.1 to master?

@db4, how can the crash be reproduced?

zdenop:
Yes, I reproduced it on master too.

Issue #1252, opened 2017-12-30
Can't create training data for font with unicode name on windows


wareya:
### Environment

commit  bc668da built by https://ci.appveyor.com/project/zdenop/tesseract/
Windows 7 64-bit
running tesstrain from a unicode-compatible mingw environment

### Current Behavior:

![image](https://user-images.githubusercontent.com/585488/34453485-8964e7b6-ed24-11e7-8016-86b971c4a6a3.png)

### Expected Behavior:

Not failing to find the font.

### Suggested Fix:

I know unicode filename access on windows is hard. You have to convert whatever the text encoding is to MS's bastardized version of UTF-16 and use _wfopen() instead of fopen(). I don't know whether a third party library is responsible for accessing font files in text2image, that would make things harder.

amitdo:
A workaround: 
Create a copy of the font in another folder, rename this copy so it will only include ASCII, and then install the font.

zdenop:
Can you please provide font (link) and short training text for reproducing the problem?

wareya:
The font I ran into the problem with is a commercial one that can't be shared, and I don't have any other fonts on my system that show up with Unicode names when I run --list_available_fonts. I'll update the issue if/when I find an affected font that I can share.

zdenop:
Maybe you can create some easy custom font just for demonstration (there are some online tools for generating font e.g. http://www.glyphrstudio.com/, https://www.prototypo.io/, https://fontstruct.com/ - I did not test it ;-) ), or you can adjust name of some free china font with [fontforge](https://fontforge.github.io)

zdenop:
Similar report with test case [in forum](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/tesseract-ocr/yg-4dmmbpwQ/NjvRFl_eBAAJ)

amitdo:
I didn't test it, but my guess is that it is a Windows-specific issue, due to the filenames encoding.

amitdo:
Probably something like this is needed in text2image:

https://github.com/tesseract-ocr/tesseract/blob/d75ef80f1224150fc2ab94f959374f926bf09388/src/api/baseapi.cpp#L1501-L1516

zdenop:
I am not sure about that (windows only issue): on linux I also get crash:
```
ERROR: Illegal UTF8 encountered.
Index 0 char = 0xffffffa2
Index 1 char = 0xffffffd2
Index 2 char = 0xffffffd4
[...]

PR #2546, opened 2019-07-08
fix windows stdout messy code

### LINKED ISSUE #2545: set output to console cause messy code on windows
### Environment

* **Platform**:  win10 

### Suggested Fix:
use `SetConsoleOutputCP`   https://docs.microsoft.com/en-us/windows/console/setconsoleoutputcp
command use 
`tesseract E:\test\single_ch_line.jpg stdout -l chi_sim `

zhuangzhuang:
fix #2545

### DIFF
--- a/src/api/tesseractmain.cpp
+++ b/src/api/tesseractmain.cpp
@@ -69,6 +69,31 @@ static void Win32WarningHandler(const char* module, const char* fmt,
 }
 
 #endif /* HAVE_TIFFIO_H */
+
+class AutoWin32ConsoleOutputCP {
+ public:
+  explicit AutoWin32ConsoleOutputCP(UINT codeCP) : changed_(false) {
+    auto oldCP = GetConsoleOutputCP();
+    if (oldCP == codeCP) {
+      return;
+    }
+    oldCP_ = oldCP;
+    changed_ = true;
+    SetConsoleOutputCP(codeCP);
+  }
+  ~AutoWin32ConsoleOutputCP() {
+    if (changed_) {
+      SetConsoleOutputCP(oldCP_);
+    }
+  }
+
+ private:
+  bool changed_;
+  UINT oldCP_;
+};
+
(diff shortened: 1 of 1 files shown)


egorpugin:
Please, remove global and create this object explicitly in main.
I think simple code below is enough:
```
#ifdef _WIN32
SetConsoleOutputCP(CP_UTF8);
#endif
```

src/api/tesseractmain.cpp
```diff
@@ -69,6 +69,31 @@ static void Win32WarningHandler(const char* module, const char* fmt,
...
+      SetConsoleOutputCP(oldCP_);
+    }
+  }
+
+ private:
+  bool changed_;
+  UINT oldCP_;
+};
+
+static AutoWin32ConsoleOutputCP autoWin32CosnoleOutputCP(CP_UTF8);
```
stweil: Typo in name

zhuangzhuang:
> I think simple code below is enough for main():
> 
> ```
> #ifdef _WIN32
> SetConsoleOutputCP(CP_UTF8);
> #endif
> ```

@egorpugin 
i think this must handle on exit.
just  like https://github.com/microsoft/vcpkg/blob/master/toolsrc/src/vcpkg.cpp#L322

egorpugin:
Sorry, did not understand what you mean.

src/api/tesseractmain.cpp
```diff
@@ -69,6 +69,31 @@ static void Win32WarningHandler(const char* module, const char* fmt,
 }
 
 #endif /* HAVE_TIFFIO_H */
+
+class AutoWin32ConsoleOutputCP {
+ public:
+  explicit AutoWin32ConsoleOutputCP(UINT codeCP) : changed_(false) {
+    auto oldCP = GetConsoleOutputCP();
```
egorpugin: Maybe it's possible to remove this? Just always set and unset, even if the same CP.

zdenop:
@zhuangzhuang: should not something like this help to solve #1252?

PR #2549, opened 2019-07-08
Avoid using experimental C++14/17 support in CMake

theirix:
This commit points CMAKE_CXX_STANDARD to the latest non-experimental standard.

CMake announces C++14 and C++17 support even if the
compiler supports it only experimentally (c++1y and c++1z).
It breaks cmake standard detection and requires workarounds
for old compilers.

It heals build on clang 3.9, 4.0, gcc 4.8 and possible for other compilers.


stweil:
Thank you!

PR #2561, opened 2019-07-12
Implement double dot product using Intel FMA and reduce number of include files

stweil:


stweil:
This implements the FMA based dot product. It is not enabled by default, but can be used with `-c dotproduct=fma`. In my tests it performed similar to AVX.

Please note that the code is only used for `tessdata_best` (which uses double values).

zdenop:
thanks.

PR #2563, opened 2019-07-13
RFC: resultiterator_test - Enable subscripts/superscripts test

Shreeshrii:
@stweil Please review. 

I enabled the subtest with a replacement image, but tesseract (legacy engine, I think) is not recognizing all superscripts correctly.

I have commented out the check for the test to pass... Is that ok or should the test fail?


zdenop:
@stweil: any comments?

PR #2569, opened 2019-07-15
Replace propriatary thread data type by std::thread

stweil:


zdenop:
thanks!

egorpugin:
Btw, sw paths are changed (new version) and will be unstable during some time.
So, no artifacts on appveyor.

zdenop:
Would it be possible with sw to use (in tesseract) `#include <leptonica/allheaders.h>`?

stweil:
That include statement would work in most cases, but autoconf builds would fail if Leptonica was installed in a non standard directory, for example in $HOME.

    ~$ pkg-config --cflags lept
    -I/home/include/leptonica

`leptonica/allheaders.h` would not be found.

egorpugin:
> Would it be possible with sw to use (in tesseract) #include <leptonica/allheaders.h>?

1) It depends on leptonica configuration. But I control it, so if you need to, we can do it.
2) I don't know about default leptonica include statement (is it <allheaders.h> or <leptonica/allheaders.h>), so current way simply <allheaders.h>. If in distributions (using autotools/make) leptonica is installed under `leptonica` folder (and tess wants it too), it means my current leptonica config is wrong and I need to fix it.

zdenop:
Maybe we can move to dev forum ;-) 
My reasoning in this case is not about something that does not work, but about code (readability) improvement - that could break some builds (which is IMO fine for 5.0 version ;-)):
-  `allheaders.h` is too general filename.
-  using convention `#include "allheaders.h"`:
a) is misleading users to expect that this file is part of tesseract and not external library. This was pointed in past by other users, when they wanted to integrate tesseract to their project as library
b) [instructing preprocessor](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html) to first [look at the current working directory over system headers](https://stackoverflow.com/questions/3162030/difference-between-angle-bracket-and-double-quotes-while-including-heade)). So may have effect build speed also.

egorpugin:
So you want to include `allheaders.h` as `#include <leptonica/allheaders.h>`?
Ok, I'll do it.

zdenop:
Yes, this would be great. But I do not want to break travis and appveyor builds...

zdenop:
@stweil : it seems that [this PR broke travis build](https://travis-ci.org/tesseract-ocr/tesseract/builds/558903822) for linux: 
```
Scanning dependencies of target lstmeval
[ 98%] Building CXX object src/training/CMakeFiles/lstmeval.dir/lstmeval.cpp.o
[ 98%] Linking CXX executable ../../bin/lstmeval
/usr/bin/ld: libunicharset_training.a(lstmtester.cpp.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
//lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
src/training/CMakeFiles/lstmeval.dir/build.make:95: recipe for target 'bin/lstmeval' failed
make[2]: *** [bin/lstmeval] Error 1
CMakeFiles/Makefile2:729: recipe for target 'src/training/CMakeFiles/lstmeval.dir/all' failed
make[1]: *** [src/training/CMakeFiles/lstmeval.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
The command "make" exited with 2.
```

stweil:
Yes, it looks like the C++ thread implementation for Linux needs the POSIX thread library. Git master is fixed now, see commit 23ef93ac4d60932faf6a4ac86ed6ffa117ed650d.

PR #2571, opened 2019-07-16
fix cmake warning.

zhuangzhuang:
mismatch if /endif


stweil:
Thank you for fixing this.

PR #2572, opened 2019-07-16
Give word's bounds to callback also during second pass

Robyer:


zdenop:
thanks

stweil:
@Robyer, could you please add a short explanation what was fixed or improved by this change?

Robyer:
@stweil The `progress_callback2` callback informs about current processing progress and uses currently processed word's rectangle as argument. But for some unknown reason it was forcing `0` during second pass. This PR changes it to always give the correct word bounds to the callback.

This modified callback is used for example in tess-two Android library. Applications that uses it (e.g. TextFairy) can show progress visually to user by showing which word is currently being processed on the processed image. Giving `0` during second pass doesn't make sense and make user experience worse as there is no visual progress during that time.

PR #2579, opened 2019-07-17
scanutils.cpp: include sys/types.h

ffontaine:
Include sys/types.h (if available) to avoid the following build failure:

scanutils.cpp: In function 'int tvfscanf(FILE*, const char*, va_list)':
scanutils.cpp:217:3: error: 'off_t' was not declared in this scope
   off_t start_off = ftell(stream)

Fixes:
 - http://autobuild.buildroot.org/results/bdfbafa105ddf0b9b9399d94b557d318ad587d79

### DIFF
--- a/src/ccutil/scanutils.cpp
+++ b/src/ccutil/scanutils.cpp
@@ -26,6 +26,10 @@
 #include <cstring>
 #include <limits>       // for std::numeric_limits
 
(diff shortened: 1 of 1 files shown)


zdenop:
What is your OS and compiller version?

stweil:
There is in easier solution: replace off_t by long – now done in Git master.

Issue #2551, opened 2019-07-08
Feature request: embed pdf.ttf to tesseract library


zdenop:
If somebody want to distribute just tesseract library, also tessdata/pdf.tff must be shipped for providing searchable pdf functionality.

IMO [pdf.ttf](https://github.com/tesseract-ocr/tesseract/blob/master/tessdata/pdf.ttf) is the only file from[ tessdata directory](https://github.com/tesseract-ocr/tesseract/tree/master/tessdata), that can not be replaced (all config files are just list of variables that can be handled now with -c option of tesseract executable or API).

jbreiden:
That makes it a little harder to change the font, but we haven't changed it
for years.
So sounds reasonable to me. Are you thinking of uint8 pdf_ttif[] = { 0x00,
0x01, ... } ?

stweil:
Is the current Tesseract library useful without any external data, not even traineddata? I could imagine a Tesseract library which gets external data from the Internet on demand (maybe with local caching), but that is not the current state.

For `pdf.ttf` an integration into the code could be fine because it is small (only 572 bytes), and the codesize could be reduced by the file handling overhead.

zdenop:
You need traningdata for tesseract initialization, so without them library does not work.
Other topics is OSD: I found quite a lot of people use tessseract of orientation and deskew detection... Is more accurate than routine available in leptonica...

PR #2602, opened 2019-08-04
Clean ambigs.h and replace octal characters by UTF-8 string in universalambigs

stweil:


stweil:
The `LoadUniversal` method is always called for each language (plus typically for `osd`) and requires significant resources (time and memory) although the `unichar_ambigs` are often unused.

So there remains potential for optimizations.

Shreeshrii:
Is unichar_ambigs used only by legacy engine/models?

zdenop:
thanks!

egorpugin:
VS now says
```
C:/projects/tesseract/src/ccutil/universalambigs.cpp(7321): fatal error C1091: compiler limit: string exceeds 65535 bytes in length
```

zdenop:
What it version of VS? I just try clang 9 with VS compatibility and build is without problem.

egorpugin:
This is from appveyor (VS2017).
Any VS has such string limit. Other compilers obviously don't.
And clang usually do not repeat bad parts of VS in compat mode.

zdenop:
OK. I see....  On SO there is [suggestion to use array instead of long string](https://stackoverflow.com/questions/2481998/how-do-i-include-extremely-long-literals-in-c-source) .

stweil:
That VS compilers are always good for a surprise. They accept long character arrays if those arrays are initialized using integers (like in the old code), but not if they are initialized by a string (like in the new code), although that is the more natural form.

I already considered to move the string into an external config file. That would reduce the size of the tesseract binary, but cost a little extra time for the file handling. I think that might be acceptable if it is only done when it is really used – I still have to find out that.

zdenop:
VS standard is to use resource file for such purposes. Similar technique is used by Qt library (on all supported platforms). But there is nothing like that in pure C or C++ that would be portable...

Unfortunately on your idea to move it to external file is in conflict with [my idea](https://github.com/tesseract-ocr/tesseract/issues/2551) ;-) to have just library where is everything needed excluding language data. So packagers do not need to distribute config files and care about their location.

egorpugin:
I do similar embedding of needed files.
Cppan/Sw unpacks them when running.

amitdo:
The 'temporary' hack done in b8079d8ce18996a should be fixed somehow.

amitdo:
The langdata repo also contains unicharambigs files:
https://github.com/tesseract-ocr/langdata/blob/master/eng/eng.unicharambigs

amitdo:
The content of `universalambigs.cpp` looks strange to me.

PR #2604, opened 2019-08-05
Add glib and gobject libs for text2image (workaround for Pango 1.43)

rezso:
training/text2image requires glib and gobject:
/usr/bin/ld: CMakeFiles/text2image.dir/pango_font_info.cpp.o: undefined reference to symbol 'g_free'
/usr/bin/ld: /usr/lib/libglib-2.0.so.0: error adding symbols: DSO missing from command line

/usr/bin/ld: CMakeFiles/text2image.dir/pango_font_info.cpp.o: undefined reference to symbol 'g_object_unref'
/usr/bin/ld: /usr/lib/libgobject-2.0.so.0: error adding symbols: DSO missing from command line

### DIFF
--- a/src/training/CMakeLists.txt
+++ b/src/training/CMakeLists.txt
@@ -271,6 +271,7 @@ pkg_check_modules(Cairo REQUIRED cairo)
 pkg_check_modules(PangoFt2 REQUIRED pangoft2)
 pkg_check_modules(PangoCairo REQUIRED pangocairo)
 pkg_check_modules(FontConfig REQUIRED fontconfig)
+pkg_check_modules(Glib REQUIRED glib-2.0 gobject-2.0)
 endif()
 
 set(text2image_src
@@ -306,6 +307,7 @@ target_link_libraries       (text2image
     ${PangoCairo_LIBRARIES}
     ${PangoFt2_LIBRARIES}
     ${FontConfig_LIBRARIES}
+    ${Glib_LIBRARIES}
 )
 endif()
 if (CPPAN_BUILD)


zdenop:
What is your environment?
At the moment glib is not needed for linux, mac and windows. Increasing dependencies is not desired.

egorpugin:
Pango uses glib, but it should not be explicitly specified for tess.
https://software-network.org/org.sw.demo.gnome.pango.pango/1.44.3/dependencies

rezso:
g_object_unref: https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L213
https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L235
https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L240
https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L310
https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L334
... and some more

g_free:
https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L419

So both glib and gobject required.

zdenop:
Please reply to my question. 
Your point to source code does not reveal anything regarding your PR. See [linux build](https://travis-ci.org/tesseract-ocr/tesseract/jobs/568378276) and [maxos build](https://travis-ci.org/tesseract-ocr/tesseract/jobs/568378277): they works without your changes perfectly.

stweil:
@rezso, how can I reproduce the linker errors? Which OS and compiler do you use?

rezso:
I compiled tesseract under linux, with gcc 9.
But if I build tesseract with the training tools, tesseract causes an segmentation fault in geeqie.
Without training tools, tesseract works fine, so I disabled the training tools.

stweil:
How are Tesseract and Geequie related?

I still do not understand how to reproduce the linker error.

rezso:
geeqie supports rendering thumbnails for video files with ffmpeg-thumbnailer, which requires ffmpeg, and ffmpeg supports tesseract (--enable-libtesseact configure option).

Flamefire:
I support this PR. The reasoning in https://github.com/tesseract-ocr/tesseract/pull/2604#issuecomment-518815976 is absolutely correct by itself. The failure can be reproduced if you do a "normal" build with CMake being able to find pkgconfig and ICU.

Note that the autotools based build already does that: https://github.com/tesseract-ocr/tesseract/blob/f268e6615e619a7140de78d9db3f8240fa8c68bd/src/training/Makefile.am#L239

stweil:
> Note that the autotools based build already does that.

That's a bug, or at least code which should use `$(pango_LIBS)` instead. Both CMake and autotools should use the result from pkg-config:

    $ pkg-config --libs pango
    -lpango-1.0 -lgobject-2.0 -lglib-2.0

stweil:
> The failure can be reproduced if you do a "normal" build with CMake being able to find pkgconfig and ICU.

That works for me:

    [...]
    [100%] Building CXX object src/training/CMakeFiles/text2image.dir/tlog.cpp.o
    [100%] Linking CXX executable ../../bin/text2image
    [100%] Built target text2image

Flamefire:
Ok let me put it differently: 
- The file `pango_font_info.cpp` is part of `text2image`
- There is a usage of `g_object_unref` at https://github.com/tesseract-ocr/tesseract/blob/c58efee4ba991551d0eae801cc33d1b79d8b11cb/src/training/pango_font_info.cpp#L310
- This function is defined in libgobject-2.0
- Hence the target `text2image` directly depends on libgobject-2.0 and needs to link against it

`pkg-config --libs pango` results for me in `-L/scratch/ws/s3248973-EasyBuild/easybuild/eb/Pango/1.43.0-GCCcore-8.2.0/lib -lpango-1.0`.

Am I correct assuming you don't have pango installed as a shared library? This could explain pkgconfig spitting out private dependencies which explains why it works for you.

src/training/CMakeLists.txt
```diff
@@ -271,6 +271,7 @@ pkg_check_modules(Cairo REQUIRED cairo)
 pkg_check_modules(PangoFt2 REQUIRED pangoft2)
 pkg_check_modules(PangoCairo REQUIRED pangocairo)
 pkg_check_modules(FontConfig REQUIRED fontconfig)
+pkg_check_modules(Glib REQUIRED glib-2.0 gobject-2.0)
```
stweil: `pango.pc` already includes this requirement, and line 269 requires `pango`, so this should not be needed:

    Requires: glib-2.0 gobject-2.0

src/training/CMakeLists.txt
```diff
@@ -306,6 +307,7 @@ target_link_libraries       (text2image
     ${PangoCairo_LIBRARIES}
     ${PangoFt2_LIBRARIES}
     ${FontConfig_LIBRARIES}
+    ${Glib_LIBRARIES}
```
stweil: Those libraries are normally already part of `${Pango_LIBRARIES}`. Please verify why this is not the case for you.

stweil:
> Am I correct assuming you don't have pango installed as a shared library?

I use default installations on Ubuntu and Debian which install shared libraries:

    $ ldd bin/text2image | grep pango
	libpango-1.0.so.0 => /lib/x86_64-linux-gnu/libpango-1.0.so.0 (0x00007f8942a9f000)
	libpangocairo-1.0.so.0 => /lib/x86_64-linux-gnu/libpangocairo-1.0.so.0 (0x00007f89427ee000)
	libpangoft2-1.0.so.0 => /lib/x86_64-linux-gnu/libpangoft2-1.0.so.0 (0x00007f89427d5000)

src/training/CMakeLists.txt
Flamefire: Actually not always. From my pango.rc:

```
Name: Pango
Description: Internationalized text handling
Version: 1.43.0
Requires.private: glib-2.0 >=  2.38.0, gobject-2.0 >=  2.38.0, fribidi >=  0.19.7, harfbuzz >=  1.4.2, fontconfig >=  2.11.91, freetype2, xrender, xft >=  2.0.0, cairo >=  1.12.10
Libs: -L${libdir} -lpango-1.0
Libs.private: -lm
Cflags: -I${includedir}/pango-1.0
```

What version of Pango and pkgconfig are you using?

Anyway: It is required because there is a **direct** dependency from this target to glib. Relying on a dependency dragging in another dependency is wrong if you require that other dependency directly

Why do you oppose this change so much?

stweil:
> pkg-config --libs pango results for me in -L/scratch/ws/s3248973-EasyBuild/easybuild/eb/Pango/1.43.0-GCCcore-8.2.0/lib -lpango-1.0

That's causing the problem. Could it be that your pango library was linked with static versions of glib and gobject? Otherwise the output from pkg-config looks incomplete.

Flamefire:
> That's causing the problem.

To be pedantic: This is *showing* the problem. 

> Could it be that your pango library was linked with static versions of glib and gobject? 

Just checked: `ldd` on libpango shows dependencies on glib and gobject.

> Otherwise the output from pkg-config looks incomplete.

I guess it could depend on the version of pkg-config. Maybe yours doesn't know `Requires.private`? See https://github.com/tesseract-ocr/tesseract/pull/2604#discussion_r365334663

Anyway: The target has a direct dependency to glib and hence need to link against it.

stweil:
> The target has a direct dependency to glib and hence need to link against it.

It has not . Otherwise the code would have to include `glib-object.h` or other glib header files.
The code depends on pango which requires glib not only privately, but also in the code when pango objects must be released using `g_free` or `g_object_unref`.

We could add the suggested libraries, but I still think that would be a workaround for a bug in your local installation.

Flamefire:
> It has a direct dependency because it uses pango, but not otherwise: it would have to include glib-object.h or other glib header files.

If you directly use a symbol from a library in your library then this is a direct dependency. Not including `glib-object.h` directly although you use symbols from it can be considered a bug (and is by static analysers like IWYU) You only get that symbol because you happen to include one of the headers from pango including a glib header. If you move that out to some other file but keep the g_free here it won't build because of that bug.

> pango which requires glib not only privately

However this is correct. Which pango version do you have? The private depencency is a bug in Pango < 1.44: https://gitlab.gnome.org/GNOME/pango/commit/d0cb6be7431d1a3c711bd45bcf05b34601604037

It is likely the bug was introduced in 1.43: https://gitlab.gnome.org/GNOME/pango/commit/453c85215cb7f562866b0daef6a9b24a6957bf73

> We could add the suggested libraries, but I still think that would be a workaround for a bug in your local installation.

So yes and no. It is a bug in Pango 1.43.x to not declare glib as a dependency. However due to the above mentioned direct use of a symbol it certainly makes sense to declare the direct dependency.

What I don't get: Why the whole discussion? Are there any drawbacks to this? It fixes a bug and is certainly correct and minimal. There is no way this could introduce an issue, is there?

stweil:
Thank you for investigating. So this issue is caused by a bug in Pango 1.43. Older and newer versions work fine.

You can either use a Pango version without that bug or fix it locally, for example by adding the required libraries to pango.rc.

Or a maintainer can use this PR as a workaround. Then I suggest to add a comment which explains the reason for those new statements.

stweil:
@rezso, which Linux distribution and which version of pango do you use?

Flamefire:
> You can either use a Pango version without that bug or fix it locally, for example by adding the required libraries to pango.rc.

That's what I'm doing now. Not great but well, if you don't want that patch.

amitdo:
>If you directly use a symbol from a library in your library then this is a direct dependency. Not including glib-object.h directly although you use symbols from it can be considered a bug (and is by static analysers like IWYU)

IWYU is this tool:
https://include-what-you-use.org/

egorpugin:
Feel free to reopen and update if the issue still exists for you.

PR #2613, opened 2019-08-12
Remove unused code

stweil:
Several source files were removed because they only contained unused code.

PR #2614, opened 2019-08-12
Move source files which are used for training only to src/training

stweil:
They are moved from src/classify and src/lstm to src/training.

This reduces the size of the Tesseract library.


stweil:
Together with pull request #2613 this completes the symbol cleanup for the Tesseract library which now only contains symbols needed for the Tesseract executable and for the C API.

The Tesseract library is reduced significantly in size by both pull requests:

    3590092	  47504	  15616	3653212	 37be5c	/tmp/libtesseract.so.5.0.0-old
    3451063	  45808	  15616	3512487	 3598a7	/tmp/libtesseract.so.5.0.0-new

stweil:
Updated PR to fix `CMakeLists.txt` and `src/training/CMakeLists.txt`, so CI should pass now.

src/training/CMakeLists.txt
```diff
@@ -76,6 +76,22 @@ set(common_training_src
...
+    intfeaturemap.h
+    mastertrainer.cpp
+    mastertrainer.h
+    networkbuilder.cpp
+    networkbuilder.h
+    sampleiterator.cpp
+    sampleiterator.h
+    trainingsampleset.cpp
+    trainingsampleset.h
 )
```
stweil: @egorpugin, please review my changes for CMake. I don't know why there are three libraries for training and added the moved code to one of them:

    src/training/libunicharset_training.a
    src/training/libtessopt.a
    src/training/libcommon_training.a

Wouldn't it be simpler to have a single library for training?

egorpugin: > Wouldn't it be simpler to have a single library for training?

Only in case when files are used in >1 target.
Otherwise personally I prefer to keep files near users (when use count ==1).

egorpugin:
D:/dev/tesseract/src/classify/trainingsample.cpp(27): fatal error C1083: Cannot open include file: 'intfeaturemap.h': No such file or directory

stweil:
@egorpugin, yes, some files in `src/classify` require include files which are now in `src/training`. I just notices that `sw.cpp` also needs some changes. Should I try to fix that?

egorpugin:
Yes, please. Now it is very dummy implementation copied from cmake and not makefiles (which seems is in better shape).
Are you using win or linux?

egorpugin:
I've removed moved files from the installation, but that single error is still there.

PR #2620, opened 2019-08-16
simd: Check whether the OS supports FMA, AVX, ...

stweil:

### DIFF
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -120,18 +120,19 @@ SIMDDetect::SIMDDetect() {
   __cpuid(cpuInfo, 0);
   max_function_id = cpuInfo[0];
   if (max_function_id >= 1) {
+    bool xmm_ymm_state_enabled = (_xgetbv(0) & 6) == 6;
     __cpuid(cpuInfo, 1);
 #if defined(SSE4_1)
     sse_available_ = (cpuInfo[2] & 0x00080000) != 0;
 #endif
 #if defined(FMA)
-    fma_available_ = (cpuInfo[2] & 0x00001000) != 0;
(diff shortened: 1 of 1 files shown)


stweil:
@egorpugin, I could not test this with MSVC, and Appveyor CI still fails.

egorpugin:
Isn't it will be better to use https://github.com/google/cpu_features
instead of our manual hand-crafted implementation?

stweil:
A short look on that code shows that it might be used, but it looks a little bit oversized because it implements lots of things which are not needed for Tesseract. The current Tesseract code uses the Intel intrinsics which I think is better than the assembler code used by _cpu_features_.

stweil:
@egorpugin, what is causing the build failure with Appveyer CI?

egorpugin:
I have to update sw.
(Such errors are because of different ABI between sw.exe and compiled .dll module - build script.)

egorpugin:
Fixed.

AllofLife:
hi stweil thank you for your pay,if i want to deploy the tesseract to a arm linux like nvidia-jetson nano ubuntu18.04 ,what should i need to do to reduse the running time? 
If you can reply, I will be very grateful.

stweil:
There are currently no special settings for Jetson Nano to make it faster.

AllofLife:
We now understand that the reason why the deployment of jetson becomes slower.
Specifically in the same recognition rate, x86 is about eight times faster than arm.
 may be that the arm architecture does not have matrix operation acceleration instructions such as AVX.
 or it may be because we recognize Chinese and Chinese is too complex to bring slow speed.
 Can you give suggestions for optimization?  thank!

PR #2643, opened 2019-09-10
Fix several runtime errors detected by UndefinedBehaviorSanitizer

stweil:
Those errors can be reproduced by running the unit tests from a build with sanitizers:

    ./configure '--enable-debug' '--disable-openmp' '--disable-shared' 'CXX=clang++-7' \
      'CXXFLAGS=-g -O0 -D_GLIBCXX_DEBUG -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=fuzzer-no-link,address,undefined -fsanitize-address-use-after-scope'
    make training check
    grep UndefinedBehaviorSanitizer unittest/*log

### DIFF
--- a/src/ccmain/osdetect.cpp
+++ b/src/ccmain/osdetect.cpp
@@ -105,7 +105,7 @@ void OSResults::update_best_script(int orientation) {
       second = scripts_na[orientation][i];
     }
   }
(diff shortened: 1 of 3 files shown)


zdenop:
thanks.

PR #2652, opened 2019-09-15
Disabled leagcy engine build: Disable code related to ambigs.

amitdo:


amitdo:
**Please don't merge yet!**

amitdo:
It was compiled fine with --disable-shared.

With --disable-static it gives me a linker error.

amitdo:
>/usr/bin/ld: ../textord/.libs/libtesseract_textord.a(cjkpitch.o): relocation R_X86_64_PC32 against symbol `_ZN13GenericVectorIfE5clearEv' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
Makefile:540: recipe for target 'libtesseract.la' failed

amitdo:
@stweil, can you help me with the linker error?

stweil:
It works for me without a linker error (Debian stable, g++).

src/ccutil/Makefile.am
```diff
@@ -10,6 +10,10 @@ AM_CXXFLAGS += -fvisibility=hidden -fvisibility-inlines-hidden
 AM_CPPFLAGS += -DTESS_EXPORTS
 endif
 
+if DISABLED_LEGACY_ENGINE
+AM_CPPFLAGS += -DDISABLED_LEGACY_ENGINE
+endif
+
```
stweil: This is required in `textord`, too.

src/ccutil/ambigs.h
```diff
@@ -215,4 +217,19 @@ class UnicharAmbigs {
 
 }  // namespace tesseract
 
+
+#else
```
stweil: The `#else` part looks like a hack which can be avoided.

src/ccutil/Makefile.am
stweil: ... and for `lstm` as well. Done in my additional commit.

src/ccutil/ambigs.h
stweil: I removed it in my additional commit.

stweil: @amitdo, do we already have more hacks like that? Then they should be examined closer, too.

src/ccutil/Makefile.am
amitdo: > This is required in `textord`, too.

Before asking for your help, I tried adding that code to textord, but there was still a linker error.

> ... and for `lstm` as well.

This one (together with the others) solves the issue.

src/ccutil/ambigs.h
amitdo: Yes, that code was meant to be temporary.

My plan was to fix it later. You already fixed it yourself. Thanks!

amitdo:
The PR is ready to be merged now.

stweil:
Thank you. It is merged now.

Issue #2584, opened 2019-07-21
user_words_suffix not working


TAQBIBT:
We are trying to provide a user words file via available control params. Unfortunately I am getting below error : Error: failed to load user-words

Environment
tesseract v5.0.0-alpha.20190708
 leptonica-1.78.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.3) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
 Found AVX2
 Found AVX
 Found SSE
 Found libarchive 3.3.2 zlib/1.2.11 liblzma/5.2.3 bz2lib/1.0.6 liblz4/1.7.5

tesseract eng.exp0697.tif out -l eng --dpi 300 --psm 8 --oem 1 bazaar

and bazaar file in the directory tessdata/configs

bertsky:
I believe I can confirm this. Tesseract does not bother to look for the user words and patterns files in tessdata anymore, or globbing the configured suffix. It just looks for the values configured as `user_words_suffix` and `user_patterns_suffix` verbatim in the CWD. This is a regression. Trying to find out more...

bertsky:
It seems that the only part _not_ working is the `_suffix` mechanism – whereas the `_file` mechanism does work (both via config file and with the CLI options `--user-patterns` and `--user-words`.

It looks like the `CCUtil` instance in `LSTMRecognizer` does not get the same `language_data_path_prefix` as in the `Tesseract`...

bertsky:
Found it: In 297d7d86ceb3b3365876b56f08a18cba066bf23f, I forgot to also bridge the non-LSTM CCUtil's runtime knowledge of `language_data_path_prefix` (which is used to concatenate with suffixes for additional files).

stweil:
@TAQBIBT, could you please try the latest code and close the issue if it was fixed?

anergui:
close #2584

PR #2663, opened 2019-09-19
fix langdata (user words/patterns) file suffixes for LSTMs:

bertsky:
- add another constructor for LSTMRecognizer
  which takes the language_data_path_prefix configured/selected
  at runtime and passes it to the internal CCUtil
- use this in Tesseract::init_tesseract_lang_data when LSTMs
  are available

(this was missing from 297d7d86ceb3b3365876b56f08a18cba066bf23f)

### DIFF
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -175,7 +175,7 @@ bool Tesseract::init_tesseract_lang_data(
       tessedit_ocr_engine_mode == OEM_TESSERACT_LSTM_COMBINED) {
 #  endif  // ndef DISABLED_LEGACY_ENGINE
     if (mgr->IsComponentAvailable(TESSDATA_LSTM)) {
-      lstm_recognizer_ = new LSTMRecognizer;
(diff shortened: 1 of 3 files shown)


bertsky:
Fixes #2584

bertsky:
But maybe anyone has a better idea of how to make this work with LSTMs? 

(The story is the same as for the configuration variables in #2328 – we have to get the prefix into the shallow `CCUtil` / `Dict` inside `LSTMRecognizer`. For the config variables I decided to pass the existing `ParamsVectors` via `LSTMRecognizer::LoadDictionary()`, resetting the defaults only for the specific variables to be shared, via `Param::ResetFrom()`. But here we are dealing with the combined runtime settings for tessdata prefix and language. We could alternatively use `TessdataManager` for this, maybe by giving it an additional member `language_data_path_prefix_` to be set in the constructor. I even thought about using its existing member `data_file_name_` and just truncate the `.traineddata` suffix, but I guess this is bound to fail under certain conditions I am unable to identify.)

zdenop:
thanks.

PR #2668, opened 2019-09-22
Remove STRING from the public Tesseract API

stweil:


egorpugin:
Is it possible to switch to std::string?

stweil:
Yes, I think the final goal should be replacing `STRING` by `std::string` or using `char` pointers for simple cases.

egorpugin:
...or std::string_view in the future with C++17.

PR #2672, opened 2019-09-23
Remove member functions STRING::string and StringParam::string

stweil:
They were redundant because there exist member functions 'c_str' which do the same.


stweil:
This is a lengthy patch, but hopefully it helps for the migration towards `std::string` which has a similar `c_str` member function.

egorpugin:
Great job!

PR #2677, opened 2019-09-25
Remove vecfuncs.cpp and vecfunc.h

stweil:
Replace the macros which were declared in vecfuncs.h by member functions
and move a function which was only used in chop.cpp to that file.


amitdo:
Thanks!

PR #2686, opened 2019-10-02
Extend function BoxFileName to handle more common image names

stweil:
The function derives the file name for the .box file from an image name.

For training from existing line images, it is useful to directly support
the image names which are commonly used.

While generated images for Tesseract training typically use the name
pattern NAME.tif, other ground truth sets use NAME.bin.png for binarized
or NAME.nrm.png for grayscale images.

BoxFileName is also now a local function as it is only used locally.

### DIFF
--- a/src/ccstruct/boxread.cpp
+++ b/src/ccstruct/boxread.cpp
@@ -17,9 +17,10 @@
  **********************************************************************/
 
 #include "boxread.h"
-#include <cstring>          // for strchr, strcmp, strrchr
+#include <cstring>          // for strchr, strcmp
 #include <locale>           // for std::locale::classic
 #include <sstream>          // for std::stringstream
+#include <string>           // for std::string
 #include "errcode.h"        // for ERRCODE, TESSEXIT
 #include "fileerr.h"        // for CANTOPENFILE
 #include "genericvector.h"  // for GenericVector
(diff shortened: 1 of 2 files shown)


src/ccstruct/boxread.cpp
```diff
@@ -32,9 +33,26 @@
...
 
+// Returns the box file name corresponding to the given image_filename.
+static std::string BoxFileName(const char* image_filename) {
+  std::string box_filename = image_filename;
+  size_t length = box_filename.length();
+  std::string last = (length > 8) ? box_filename.substr(length - 8) : "";
+  if (last == ".bin.png" || last == ".nrm.png") {
+    box_filename.resize(length - 8);
+  } else {
+    const char* lastdot = strrchr(box_filename.c_str(), '.');
```
egorpugin: Use str.find('.')?

stweil: Good idea. `str.find_last_of('.')` might even be better. I'll update the PR tomorrow.

egorpugin: Not sure about `find_last_of()`.

stweil:
I updated and rebased the pull request and think that it is ready for merging now.

zdenop:
thanks

PR #2690, opened 2019-10-04
Optional speed optimisation

vidiecan:
Allows for disabling automatic inverting of lines when tesseract feels the inverting could be of help.


zdenop:
thanks!

src/ccmain/tesseractclass.cpp
```diff
@@ -66,6 +66,8 @@ Tesseract::Tesseract()
       BOOL_MEMBER(tessedit_dump_pageseg_images, false,
                   "Dump intermediate images made during page segmentation",
                   this->params()),
+      BOOL_MEMBER(tessedit_do_invert, true,
```
stweil: Would the shorter form `tessedit_invert` or a more descriptive name `tessedit_auto_invert` be better?

PR #2710, opened 2019-10-14
Generate unicharset from training text instead of Box files - tesstrain.sh

Shreeshrii:
fixes a number of issues related to Normalization failed for string mainly for Indic languages

### DIFF
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -345,12 +345,12 @@ phase_I_generate_image() {
 
 # Phase UP : Generate (U)nicharset and (P)roperties file.
 phase_UP_generate_unicharset() {
-    tlog "\n=== Phase UP: Generating unicharset and unichar properties files ==="
+    tlog "\n=== Phase UP: Generating unicharset and unichar properties files from Training Text ==="
 
     local box_files=$(ls ${TRAINING_DIR}/*.box)
     UNICHARSET_FILE="${TRAINING_DIR}/${LANG_CODE}.unicharset"
     run_command unicharset_extractor --output_unicharset "${UNICHARSET_FILE}" \
-      --norm_mode "${NORM_MODE}" ${box_files}
+      --norm_mode "${NORM_MODE}" ${TRAINING_TEXT}
     check_file_readable ${UNICHARSET_FILE}
 
     XHEIGHTS_FILE="${TRAINING_DIR}/${LANG_CODE}.xheights"
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -345,12 +345,24 @@ phase_I_generate_image() {
 
 # Phase UP : Generate (U)nicharset and (P)roperties file.
 phase_UP_generate_unicharset() {
-    tlog "\n=== Phase UP: Generating unicharset and unichar properties files from Training Text ==="
+    tlog "\n=== Phase UP: Generating unicharset and unichar properties files from Training Text and Box Files ==="
 
     local box_files=$(ls ${TRAINING_DIR}/*.box)
     UNICHARSET_FILE="${TRAINING_DIR}/${LANG_CODE}.unicharset"
-    run_command unicharset_extractor --output_unicharset "${UNICHARSET_FILE}" \
+    UNICHARSET_FILE_FROM_BOXES="${TRAINING_DIR}/${LANG_CODE}.boxes.unicharset"
+    UNICHARSET_FILE_FROM_TEXT="${TRAINING_DIR}/${LANG_CODE}.text.unicharset"
+    run_command unicharset_extractor --output_unicharset "${UNICHARSET_FILE_FROM_TEXT}" \
       --norm_mode "${NORM_MODE}" ${TRAINING_TEXT}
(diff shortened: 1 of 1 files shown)


Shreeshrii:
See related issue in forum https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/tesseract-ocr/VSIP2mbUIXc/WN5wWOuWDgAJ for Sinhla language

amitdo:
If text2image can't render a word it will drop it, so the training text and the box file might have different set of unichars.

Shreeshrii:
@amitdo That is a valid concern. It can be addressed by using find_fonts with text2image with 100% coverage.

What is happening currently is that a lot of unichars are getting dropped from the unicharset being generated by text2image even though they are there in the text as well as in the box file. It seems to be dependent on the text, font and norm_mode.

E.g. using langdata/sin/sin.training_text and  "Noto Sans Sinhala" font, the box files generate a unicharset of size 106 while the text file reports size of 121. Many of extra characters are combining marks for vowels,  of the kind which apply to both before and after a consonant. The characters being dropped are the ones getting the errors such as:

```
Invalid start of grapheme sequence:M=0xdd1
Normalization failed for string 'ෑ'
Invalid start of grapheme sequence:M=0xddc
Normalization failed for string 'ො'
Invalid start of grapheme sequence:M=0xdcf
Normalization failed for string 'ා'
Invalid start of grapheme sequence:M=0xdd0
Normalization failed for string 'ැ'
Invalid start of grapheme sequence:M=0xdcf
Normalization failed for string 'ා'
Invalid start of grapheme sequence:M=0xdd0
Normalization failed for string 'ැ'
Invalid start of grapheme sequence:D=0xd82
Normalization failed for string 'ං'
```

These characters are not being found at beginning of any words.

Shreeshrii:
This error has also been reported for Tamil, Bangla, etc.

zdenop:
@Shreeshrii : if the characters being dropped, there is a reason for it (bug?) that should be investigated. Your workaround will hide it (at least), so I would suggest to make it optional (or create both version and inform/warn user that process will use unicharset file with more characters...).

Shreeshrii:
> if the characters being dropped, there is a reason for it (bug?) that should be investigated. 

The issues have been reported multiple times for different languages before. text2image  is not producing correct box files. eg. in this case the combing marks are being listed separately rather than being combined with the consonants.

```
දෙ 111 4640 173 4685 0
න 172 4652 219 4685 0
  219 4652 245 4685 0
ස 245 4652 317 4684 0
ෞ 317 4652 342 4681 0
න් 343 4652 391 4702 0
ද 391 4638 421 4684 0
ර් 420 4651 454 4701 0
ය 455 4651 494 4683 0
ය 496 4651 535 4683 0
  535 4650 562 4693 0
ජෛ 562 4650 663 4693 0
න 665 4650 712 4682 0
ා 711 4650 728 4679 0
ග 729 4650 773 4682 0
ම 774 4650 815 4693 0
  815 4640 842 4693 0
ල 842 4640 915 4681 0
ෝ 916 4649 934 4699 0
ක 934 4649 983 4681 0
යේ 984 4648 1057 4698 0
  1057 4635 1082 4698 0
ද 1082 4635 1112 4680 0
ැ 1110 4641 1128 4670 0
ඩි 1126 4648 1168 4697 0
	 1168 4696 1169 4697 0
```

Notice all marks with the dotted circle.

Shreeshrii:
> so I would suggest to make it optional (or create both version and inform/warn user that process will use unicharset file with more characters...).

Ok. Will change PR to create both versions.

Shreeshrii:
Please see attached zip file with sample Sinhala text and related files.
[unicharerror.zip](https://github.com/tesseract-ocr/tesseract/files/3729445/unicharerror.zip)

@stweil Thoughts?

amitdo:
Does text2image render those chars/signs in words when it say "invalid..."?

Shreeshrii:
As far as I can make out, the image is rendered correctly and the characters are there in the box file. It is how the Consonant Vowel mark combinations split that is causing the problem.

Shreeshrii:
@zdenop I have added a second commit with changes as suggested by you. However, I think that just the first commit should suffice.

stweil:
The current situation with a shell script and a Python script for training is confusing for developers and for users, especially because updates not always synchronize both. How should we solve that? Maybe remove the shell script in favour of the Python script? We could also move such scripts to the tesstrain repository now.

Shreeshrii:
>Maybe remove the shell script in favour of the Python script? We could also move such scripts to the tesstrain repository now.

That's ok by me. The training tutorial by Ray will need to be updated to reflect the usage of recommended scripts.

amitdo:
Hi Shree,

What about using the wordstrbox config? It might solve the issue.

Shreeshrii:
The issue related to Indic language normalization still exists. I came across it again in response to a user having issues while training for Bengali language. Ref: https://groups.google.com/d/msgid/tesseract-ocr/0b64b093-fbad-46b7-b604-56b4fb51c9e1n%40googlegroups.com.

I would like that only my original commit in this PR be merged (not the 2nd one which just adds to the complexity and confusion).

egorpugin:
Then remove the second one. Or close this PR, open a new one and put there a link to this discussion.

PR #2714, opened 2019-10-16
Generate Unicharset from Training Text - tesstrain.py

Shreeshrii:
See https://github.com/tesseract-ocr/tesseract/pull/2710 for discussion

### DIFF
--- a/src/training/tesstrain_utils.py
+++ b/src/training/tesstrain_utils.py
@@ -421,8 +421,6 @@ def phase_I_generate_image(ctx, par_factor=None):
 def phase_UP_generate_unicharset(ctx):
     log.info("=== Phase UP: Generating unicharset and unichar properties files from Training Text ===")
 
-    box_files = pathlib.Path(ctx.training_dir).glob("*.box")
-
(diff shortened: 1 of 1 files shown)


src/training/tesstrain_utils.py
```diff
@@ -419,7 +419,7 @@ def phase_I_generate_image(ctx, par_factor=None):
 
 # Phase UP : Generate (U)nicharset and (P)roperties file.
 def phase_UP_generate_unicharset(ctx):
-    log.info("=== Phase UP: Generating unicharset and unichar properties files ===")
+    log.info("=== Phase UP: Generating unicharset and unichar properties files from Training Text ===")
 
     box_files = pathlib.Path(ctx.training_dir).glob("*.box")
```
stweil, line 424: `box_files` is now unused and should be removed.

stweil (changes requested):

src/training/tesstrain_utils.py
Shreeshrii: Thanks @stweil. I have made the requested change.

PR #2722, opened 2019-10-22
Remove 1870 warnings out of 1897 (27 left)

whosoonhwang:
Environment: Window 10 Home (x64)
Processor: Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz  2.71GHz
Memory: 16GB
IDE: Microsoft Visual Studio Community 2019 (Version 16.3.4)

I built it to use tesseract, but there are a lot of warnings in my build environment that make it hard to distinguish them from the warnings I've modified. So I removed the warnings in a way that would not affect existing routines. If this PR makes sense, please merge it and ignore it if it doesn't make sense in other environments.

### DIFF
--- a/src/ccmain/paragraphs.cpp
+++ b/src/ccmain/paragraphs.cpp
@@ -52,9 +52,9 @@ namespace tesseract {
 
 // Special "weak" ParagraphModels.
 const ParagraphModel *kCrownLeft
-    = reinterpret_cast<ParagraphModel *>(0xDEAD111F);
+    = reinterpret_cast<ParagraphModel *>(static_cast<__int64>(0xDEAD111F));
 const ParagraphModel *kCrownRight
-    = reinterpret_cast<ParagraphModel *>(0xDEAD888F);
+    = reinterpret_cast<ParagraphModel *>(static_cast<__int64>(0xDEAD888F));
 
 // Do the text and geometry of two rows support a paragraph break between them?
 static bool LikelyParagraphStart(const RowScratchRegisters &before,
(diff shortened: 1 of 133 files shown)


src/ccmain/paragraphs.cpp
```diff
@@ -51,10 +51,10 @@
 namespace tesseract {
 
 // Special "weak" ParagraphModels.
-const ParagraphModel *kCrownLeft
-    = reinterpret_cast<ParagraphModel *>(static_cast<__int64>(0xDEAD111F));
-const ParagraphModel *kCrownRight
-    = reinterpret_cast<ParagraphModel *>(static_cast<__int64>(0xDEAD888F));
+const ParagraphModel *kCrownLeft =
+    reinterpret_cast<ParagraphModel*>(static_cast<int64_t>(0xDEAD111F));
+const ParagraphModel *kCrownRight =
```
stweil: This must be `uintptr_t`.

src/ccmain/paragraphs.cpp
```diff
@@ -51,10 +51,10 @@
...
+    reinterpret_cast<ParagraphModel*>(static_cast<int64_t>(0xDEAD888F));
```
stweil: `uintptr_t`

whosoonhwang: Thank you for your advice.

stweil:
Such a large pull request cannot be reviewed and merged. Please try to make smaller pull requests which address single issues (like casting integers to pointers for example).

PR #2724, opened 2019-10-23
src/ccutil/*.header file remove casting warning

whosoonhwang:
Environment: Window 10 Home (x64)
Processor: Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz 2.71GHz
Memory: 16GB
IDE: Microsoft Visual Studio Community 2019 (Version 16.3.4)

header file first.

### DIFF
--- a/src/ccutil/bits16.h
+++ b/src/ccutil/bits16.h
@@ -31,21 +31,21 @@ class DLLSYM BITS16 {
 
   void turn_on_bit(       // flip specified bit
       uint8_t bit_num) {  // bit to flip 0..7
-    val = val | 01 << bit_num;
+    val = static_cast<uint16_t>(val | 01 << bit_num);
   }
 
   void turn_off_bit(      // flip specified bit
       uint8_t bit_num) {  // bit to flip 0..7
-    val = val & ~(01 << bit_num);
+    val = static_cast<uint16_t>(val & ~(01 << bit_num));
   }
 
   void set_bit(         // flip specified bit
       uint8_t bit_num,  // bit to flip 0..7
       bool value) {     // value to flip to
     if (value)
-      val = val | 01 << bit_num;
+      val = static_cast<uint16_t>(val | 01 << bit_num);
     else
-      val = val & ~(01 << bit_num);
+      val = static_cast<uint16_t>(val & ~(01 << bit_num));
   }
 
   bool bit(                     // access bit
--- a/src/ccutil/genericvector.h
+++ b/src/ccutil/genericvector.h
@@ -210,7 +210,7 @@ class GenericVector {
   // This function uses memcpy to copy the data, instead of invoking
   // operator=() for each element like double_the_size() does.
   static T* double_the_size_memcpy(int current_size, T* data) {
-    T* data_new = new T[current_size * 2];
+    T* data_new = new T[static_cast<size_t>(current_size) * 2];
     memcpy(data_new, data, sizeof(T) * current_size);
     delete[] data;
     return data_new;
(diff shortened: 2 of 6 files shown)


src/ccutil/unicharset.h
```diff
@@ -55,18 +55,18 @@ class CHAR_FRAGMENT {
...
-  inline void set_all(const char *unichar, int pos, int total, bool natural) {
-    set_unichar(unichar);
-    set_pos(pos);
-    set_total(total);
-    set_natural(natural);
+  inline void set_all(const char *unichar_, int pos_, int total_, bool natural_) {
+    set_unichar(unichar_);
+    set_pos(pos_);
+    set_total(total_);
+    set_natural(natural_);
```
whosoonhwang: This part is related to c4458.

src/ccutil/bits16.h
```diff
@@ -31,21 +31,21 @@ class DLLSYM BITS16 {
...
   void set_bit(         // flip specified bit
       uint8_t bit_num,  // bit to flip 0..7
       bool value) {     // value to flip to
     if (value)
-      val = val | 01 << bit_num;
+      val = static_cast<uint16_t>(val | 01 << bit_num);
     else
-      val = val & ~(01 << bit_num);
+      val = static_cast<uint16_t>(val & ~(01 << bit_num));
   }
```
stweil: The fixes for `bits16.h` are fine. Please send a separate pull request for them.

src/ccutil/genericvector.h
```diff
@@ -210,7 +210,7 @@ class GenericVector {
   // This function uses memcpy to copy the data, instead of invoking
   // operator=() for each element like double_the_size() does.
   static T* double_the_size_memcpy(int current_size, T* data) {
-    T* data_new = new T[current_size * 2];
+    T* data_new = new T[static_cast<size_t>(current_size) * 2];
```
stweil: I don't think this is the right longterm solution. Instead of that, size variables like `current_size` should use an unsigned data type, typically `size_t`.

src/ccutil/unicharset.h
```diff
@@ -573,13 +573,13 @@ class UNICHARSET {
                       int min_bottom, int max_bottom,
                       int min_top, int max_top) {
     unichars[unichar_id].properties.min_bottom =
-        ClipToRange<int>(min_bottom, 0, UINT8_MAX);
+        static_cast<uint8_t>(ClipToRange<int>(min_bottom, 0, UINT8_MAX));
```
stweil: This is a typical example where the type casts hide a problem in the code. The values which are clipped here are read from a file. Simply clipping them to 0...255 hides problems in that file which should only provide values in that range.

Therefore I consider such type casts as unwanted.

General rule: Don't add a type cast just to silence a compiler warning.

src/ccutil/bits16.h
whosoonhwang: Ok. I will fix this to send a separate pull request.

src/ccutil/genericvector.h
whosoonhwang: I see. Thanks for your advice.

src/ccutil/unicharset.h
whosoonhwang: I now understand the rules for casting.  Thank you so much.

src/ccutil/genericvector.h
whosoonhwang: @stweil 
I have a question. Changing from one source to the size_t data type and the other to referencing sources encountered the following situation: I have a routine that saves to a file, and I wonder what to do.

1. No modification should be made in this regard.
2. Implement a structure that is compatible with older files.
3. You don't need to make the old file compatible, but you can ignore it after it's upgraded.

I hope to let you know if a solution is in the above category or have a different approach.

ex: (Assume)
If I change type of `used` filed to size_t, I wonder what the effect is.
(https://github.com/tesseract-ocr/tesseract/blob/cede5b34e78ffccc2aa25e19ea80aad31c4354b7/src/ccutil/genericvector.h#L507)

stweil: For that example changing the type of `used` from `int` to `size_t` would create an incompatible file. Writing to file and reading from file should always use data types with a known size, for example `uint32_t`.

whosoonhwang: @stweil I was impressed with your kind explanation. Thank you always.

Issue #1255, opened 2018-01-01
FilterMusic has false positives in documents with tables


akhudek:
### Environment

* **Tesseract Version**: v4 commit number ebbfc3ae8df85c351002000a76900e3086375e7b
* **Platform**: Darwin 17.3.0 Darwin Kernel Version 17.3.0: Thu Nov  9 18:09:22 PST 2017; root:xnu-4570.31.3~1/RELEASE_X86_64 x86_64

### Current Behavior:
The table header and parts of the cell contents are not recognized due to the binarized version of the document having critical parts of the table removed.
![test-table](https://user-images.githubusercontent.com/147633/34470838-5eae68cc-ef08-11e7-9a0b-fe66f195c25c.png)
![testb](https://user-images.githubusercontent.com/147633/34470846-897c6996-ef08-11e7-9deb-b726e60a87ca.png)

### Expected Behavior:
If you disable FilterMusic, it works fine.
![testb2](https://user-images.githubusercontent.com/147633/34470863-e6ca0900-ef08-11e7-8078-3b48f02ff817.png)

### Suggested Fix:
It's not clear to me what the motivation was for trying to find and identify sheet music. Is it possibly for the table detector? If so, given that the table detection is fairly poor anyways and the extraction isn't complete, I'd suggest just removing this music filter code. Alternately, maybe expose an option to disable it?

zdenop:
Now it is possible to use `pageseg_apply_music_mask` e.g. with `-c pageseg_apply_music_mask=0`

stweil:
I wonder whether the current default is reasonable. If most documents give better results when this parameter is `false`, then that should be the default.

zdenop:
IMO it will need more investigations:
* regarding tables - PR 2732 seems to be still just part solution (some lines are removed in tables)
* _maybe_ current setup helps in other kind of documents - it should be tested on wider range of documents (unfortunately we do not have such test base at the moment). I have internal info from developer that use tesseract in thousands of docs and their experience is that if you "fix" page segmentation on one type of document - you break it on other types. So they suggest to be very careful when playing with page segmentation. BTW: they recognize tesseract page segmentation as best available opensouce general page segmentation library...

bertsky:
[...]

PR #2732, opened 2019-10-24
Add pageseg_apply_music_mask option to allow disabling the music mask

tenatus:
I have an image with a very large, lined table where most of the text was being dropped by the FilterMusic function, which was falsely recognizing parts of the table as a music staff.

I have added a new parameter called pageseg_apply_music_mask (default of true to preserve the current behavior) to allow this function to be disabled.

I am a little unsure of the proper name for the parameter so suggestions are welcome.

### DIFF
--- a/src/ccmain/pagesegmain.cpp
+++ b/src/ccmain/pagesegmain.cpp
@@ -209,7 +209,7 @@ int Tesseract::AutoPageSeg(PageSegMode pageseg_mode, BLOCK_LIST* blocks,
 
   ColumnFinder* finder = SetupPageSegAndDetectOrientation(
       pageseg_mode, blocks, osd_tess, osr, &temp_blocks, &photomask_pix,
-      &musicmask_pix);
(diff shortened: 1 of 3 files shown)


tenatus:
#1255

zdenop:
Basically I have no problem with other option, but can you provide image that demonstrate this problem?

I tried `tesseract i1255.png i1255 get.image` and I got different result that reported at #1255:

tessinput_org
[i1255.txt](https://github.com/tesseract-ocr/tesseract/files/3774814/i1255.txt)

When I tried `tesseract i1255.png i1255_new -c pageseg_apply_music_mask=0 get.image` I got the same result as in my previous test (e.g. the same output => `pageseg_apply_music_mask` has no effect on this image)

tenatus:
Here is part of the image I was using. This demonstrates the problem.
table

tessinput_without_music_mask
tessinput_with_music_mask

zdenop:
Thanks for example

PR #2735, opened 2019-10-28
Move all public headers to include/tesseract

amitdo:
Also:
* Fix #include path of public headers
* Fix autotools build


.gitignore
```diff
@@ -36,7 +36,7 @@ config_auto.h
 /doc/*.xml
 
 # generated version file
-/src/api/tess_version.h
+/include/tesseract/tess_version.h
```
stweil: The file name is now always preceded by the `tesseract` directory, so we could simply use `tesseract/version.h`.

amitdo:
This does not include the needed changes to CMake and sw, so the CI will fail.

I tested it with autotools and it works fine. 

@egorpugin, please push more commits to this PR or first merge and do the CMake fix later.

Please don't **squash** my commits.

.gitignore
amitdo: True

amitdo:
See https://github.com/tesseract-ocr/tesseract/issues/594#issuecomment-535632016

egorpugin:
@stweil @zdenop 
Is this ok for you?
Waiting for the approval/confirmation for such changes.

include/tesseract/Makefile.am
```diff
@@ -0,0 +1,54 @@
+AM_CPPFLAGS += \
```
stweil: Is `AM_CPPFLAGS` really needed here?

src/ccutil/Makefile.am
```diff
@@ -1,5 +1,5 @@
 AM_CXXFLAGS =
-AM_CPPFLAGS =
+AM_CPPFLAGS += -I$(top_srcdir)/include
```
stweil:
```suggestion
AM_CPPFLAGS = -I$(top_srcdir)/include
```

include/tesseract/Makefile.am
amitdo: Not sure. You can try to remove it and see if it compiles.
Remove (usr/local)/include/tesseract in your system first.

src/ccutil/Makefile.am
amitdo: Actually I think that this is correct, and the previous 'AM_CPPFLAGS=' was wrong.

amitdo: I mean that the '+=' here is correct, IMO.

include/tesseract/Makefile.am
stweil: It works without it because no files are compiled there. But I suggest an even more radical modification, namely no `include/tesseract/Makefile.am` at all. Instead use this code:

    diff --git a/Makefile.am b/Makefile.am
    index 4323f8fc..19cd43f8 100644
    --- a/Makefile.am
    +++ b/Makefile.am
    @@ -65,6 +65,24 @@ doc-clean:
     pkgconfigdir = $(libdir)/pkgconfig
     pkgconfig_DATA = tesseract.pc
     
    +pkginclude_HEADERS = include/tesseract/apitypes.h
    +pkginclude_HEADERS += include/tesseract/baseapi.h
    +pkginclude_HEADERS += include/tesseract/capi.h
    +pkginclude_HEADERS += include/tesseract/genericvector.h
    +pkginclude_HEADERS += include/tesseract/helpers.h
    +pkginclude_HEADERS += include/tesseract/ltrresultiterator.h
    +pkginclude_HEADERS += include/tesseract/ocrclass.h
    +pkginclude_HEADERS += include/tesseract/osdetect.h
    +pkginclude_HEADERS += include/tesseract/pageiterator.h
    +pkginclude_HEADERS += include/tesseract/platform.h
    +pkginclude_HEADERS += include/tesseract/publictypes.h
    +pkginclude_HEADERS += include/tesseract/renderer.h
    +pkginclude_HEADERS += include/tesseract/resultiterator.h
    +pkginclude_HEADERS += include/tesseract/serialis.h
    +pkginclude_HEADERS += include/tesseract/strngs.h
    +pkginclude_HEADERS += include/tesseract/thresholder.h
    +pkginclude_HEADERS += include/tesseract/unichar.h
    +
     # fuzzer-api is used for fuzzing tests.
     # They are run by OSS-Fuzz https://oss-fuzz.com/, but can also be run locally.
     # Note: -fsanitize=fuzzer currently requires the clang++ compiler.

amitdo:
>unittest also needs fixed include statements.

https://github.com/tesseract-ocr/tesseract/pull/2735/commits/e1bae15547aa7dd40d3d175bfa259db34b2f3226#diff-ee6e88ad125acd6b96556b3373ed48d7

amitdo:
This pull request fixes 1 alert when merging 2f8884a into cede5b3 - view on LGTM.com

fixed alerts:

 *  1 for Comparison result is always the same

I have no clue how this PR fixed that alert :-)

stweil:
> > unittest also needs fixed include statements.
> 
> [e1bae15#diff-ee6e88ad125acd6b96556b3373ed48d7](https://github.com/tesseract-ocr/tesseract/commit/e1bae15547aa7dd40d3d175bfa259db34b2f3226#diff-ee6e88ad125acd6b96556b3373ed48d7)

What about `unittest/applybox_test.cc` (and some more)?

amitdo:
>The include statements in include/tesseract don't need the added tesseract/.

At least some Google projects does what I did. I'm not sure what's the right thing to do here. 
https://github.com/google/leveldb/blob/master/include/leveldb/cache.h#L23

amitdo:
Another example:
https://github.com/apache/mesos/blob/master/include/mesos/executor.hpp#L24

amitdo:
>What about unittest/applybox_test.cc (and some more)?

My hacky python script search for files with *.cpp or *.h.

That's why it missed *.cc ...

include/tesseract/Makefile.am
amitdo: Yes, I agree.

zdenop:
I am fine with this change. 
Also there is a need to move [tesseractmain.cpp](https://github.com/tesseract-ocr/tesseract/blob/master/src/api/tesseractmain.cpp) from `api` directory as it is not part of API (to `progs`?)...

stweil:
>  Also there is a need to move tesseractmain.cpp from api directory as it is not part of API (to progs?)...

I would not mix this with the current pull request. One could also argue that `tesseractmain.cpp` is an example of using the API, so it can stay there.

stweil:
> > The include statements in include/tesseract don't need the added tesseract/.
> 
> At least some Google projects do what I did. I'm not sure what's the right thing to do here.
> https://github.com/google/leveldb/blob/master/include/leveldb/cache.h#L23

That are more examples of recursive Makefiles. They have several disadvantages. Parallel builds using more than one CPU are much slower with recursions. Additional recursive make processes need more RAM.

That's why I would like to move the `autoconf` builds away from recursive builds. The `cmake` and `sw` builds already use a flat build model.

zdenop:
I agree not to include it to this PR, but when we are "moving sources", it would be nice to make all known changes "together", so we could have stable source tree for some time ;-). IMO using `examples` instead of `progs` for `tesseractmain.cpp` would be misleading because tesseract executable is needed for training.

Maybe it would make sense to move tesseract library to separate repository and include it to tesseract repository as submodule. This should make easier to include tesseact library to other projects (e.g. wrappers)

src/ccutil/Makefile.am
stweil: At least all others use `+=`, too.

stweil:
I fixed Travis CI and the unittest code now, but Appveyor CI still fails. @egorpugin, @zdenop, can we merge the pull request nevertheless and fix cppan.yml and sw.cpp later? Other changes to be done later:

- Rename tesseract/tess_version.h to tesseract/version.h
- Use flat autoconf make for include/tesseract instead of recursive make
- Maybe replace `#include "tesseract/*.h"` by `#include <tesseract/*.h>`

amitdo:
https://github.com/tesseract-ocr/tesseract/issues/594
Egor suggested to move the programs to 'tools', 'tools/training'.

About a non-recursive make build:
https://github.com/tesseract-ocr/tesseract/pull/974

egorpugin:
+1 for Makefile.am removal from include/tesseract

PR #2743, opened 2019-11-01
Exposed the text2image option --ptsize to tesstrain.sh.

DavidMaung:
Text2image has the option --ptsize which defaults to 12.  This option is not exposed through
tesstrain.sh; thus, you cannot use tesstrain.sh to explore training with
different font sizes.  I made a small modification to expose the --ptsize
option to tesstrain.sh.  It defaults to 12 if not specified.


stweil:
@DavidMaung, the shell script is deprecated in favour of the Python script. Could you send a pull request for that, too?

DavidMaung:
@stweil Will address Monday.

PR #2785, opened 2019-11-26
automake: Use non-recursive build for tesseract and libtesseract

stweil:


stweil:
See also discussions for #974 and #2735.

stweil:
Note that the new Makefile builds tesseract and libtesseract in the build root directory, not in src/api.

Makefile.am
```diff
@@ -84,6 +82,795 @@ pkginclude_HEADERS += include/tesseract/strngs.h
 pkginclude_HEADERS += include/tesseract/thresholder.h
 pkginclude_HEADERS += include/tesseract/unichar.h
 
+# Rules for all subdirectories.
+
+noinst_HEADERS =
+noinst_LTLIBRARIES =
+
+AM_CPPFLAGS += -DLOCALEDIR=\"$(localedir)\"
```
amitdo: This flag has no effect.

https://github.com/tesseract-ocr/tesseract/search?q=LOCALEDIR&unscoped_q=LOCALEDIR

amitdo:
Do you really need all these internal static la libs?

Can't you write it more like how CMakeLists.txt is written?
I mean, put everything in libtesseract and only make internal libs related to `arch`.

Makefile.am
stweil: You are right, thanks. I removed that old code.

stweil:
> Do you really need all these internal static la libs?

They are only needed when those source files need special compiler flags, so restricting those libs to `arch` is possible if we use the same list of include directories for all source files.

I thought about that, too, but decided to migrate the old Makefiles as closely as possible in the first step (as you noticed even with flags without effect). Maybe the lengthy list of CPPFLAGS can be replaced by `-I$(top_srcdir)/src`. That requires include statements which use a directory prefix.

amitdo:
OK :-)

Did you test it?

stweil:
I have run a default build (basically `mkdir build && cd build && ../configure && make training`) and a debug build and compared the resulting file sizes. In addition I have run `make check`.

zdenop:
Thanks!

amitdo:
You didn't keep 
```
if DISABLED_LEGACY_ENGINE
AM_CPPFLAGS += -DDISABLED_LEGACY_ENGINE
endif
```

It is needed, for example, here:
https://github.com/tesseract-ocr/tesseract/blob/master/src/ccmain/tesseractclass.h#L34-L36

stweil:
Yes, I moved it to `config_auto.h` in the follow-up commit 39cc7b58080b72c9d6edfecf428c9600d6afab23. Maybe an include statement has to be added to `tesseractclass.h` now.

amitdo:
That was just one example.

Here is another:

https://github.com/tesseract-ocr/tesseract/blob/cb0c024a6f92fd04c951add6cf7ff497625cfda0/src/wordrec/wordrec.h

If you'll go with the 'include statement way', you will have to edit many files.

IMO, it will be easier to go with the 'Makefile.am way'.

stweil:
@amitdo, it should be fixed now by commit a06d0d84497d7b0ad905c298bf5bacf8f7063c69.

amitdo:
Thanks!

PR #2807, opened 2019-12-05
Remove Emacs C modeline in comment line 1

stweil:
Those files are C++, and the wrong modeline is not needed at all.


zdenop:
thanks.

PR #2826, opened 2019-12-18
make BlockPolygon usable

bertsky:
Layout analysis delivers segments that often need to be non-rectangular to accomodate close neighbours, image near text, ragged/unaligned lines etc. The CLI builds on the internal, polygon-based data structures, which allow for that without conflicts/overlaps. But API users up until now could only use bounding boxes, because `BlockPolygon` did not deliver correct coordinates:

1. unlike `BoundingBox`, they were not "unrotated" (i.e. the internal deskewing has not been compensated for, giving coordinates widely out of scope)
2. unlike `BoundingBox`, they were not clipped (i.e. forced into the interval of the image/rectangle)

Both issues have been fixed here. 

But there is a third problem that I have not tackled yet: Polygon paths are often self-intersecting or otherwise unusable. This even confuses Tesseract itself when it tries to provide `GetImage()` on the raw image (i.e. when passed non-null `original_img`). It's not even enough to take the convex hull of these invalid paths. (My guess is that there is a bug in the incremental aggregation of the polyblock segments, in how it orders them, depending on their orientation.)

Anyway, this is already much more usable than before. I can provide example images if necessary.

### DIFF
--- a/src/ccmain/pageiterator.cpp
+++ b/src/ccmain/pageiterator.cpp
@@ -370,7 +370,13 @@ Pta* PageIterator::BlockPolygon() const {
     return nullptr;  // Already at the end!
   if (it_->block()->block->pdblk.poly_block() == nullptr)
     return nullptr;  // No layout analysis used - no polygon.
-  ICOORDELT_IT it(it_->block()->block->pdblk.poly_block()->points());
+  // Copy polygon, so we can unrotate it to image coordinates.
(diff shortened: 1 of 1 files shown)


bertsky:
Partial CI failure (on macos) is unrelated (it cannot find cmake)...

zdenop:
thanks!

PR #2861, opened 2020-01-18
Remove redundant method from class GenericVector

stweil:
length() is not needed: it can be replaced by size().

### DIFF
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -721,9 +721,9 @@ void Tesseract::blamer_pass(PAGE_RES* page_res) {
         static_cast<IncorrectResultReason>(bl)),
         page_res->blame_reasons[bl]);
   }
-  if (page_res->misadaption_log.length() > 0) {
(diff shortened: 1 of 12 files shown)


zdenop:
thanks!

src/ccmain/control.cpp
```diff
@@ -721,9 +721,9 @@ void Tesseract::blamer_pass(PAGE_RES* page_res) {
         static_cast<IncorrectResultReason>(bl)),
         page_res->blame_reasons[bl]);
   }
-  if (page_res->misadaption_log.length() > 0) {
+  if (page_res->misadaption_log.size() > 0) {
```
amitdo: Maybe you can drop the `> 0` part.

egorpugin: `.empty()` is better.

stweil: Yes. In this case `! .empy()`, of course. `git grep "length().*0"` and `git grep "size().*0"` shows more code locations which could use `empty`.

Issue #2859, opened 2020-01-17
tesseract.exe not flushing stdout 


jkang-eng:
### Environment
* **Tesseract Version**: latest master
* **Platform**: Windows 32/64 bit

### Current Behavior:
In my application, I spawn tesseract.exe from java and continuously send request (path to image) / receive response  (tokens) through standard pipe-stdio.
For example, tesseract is launched with following args.
```
tesseract.exe stdin stdout -c stream_filelist=true
```

Problem with the existing code, **tokens are never flushed from  tesseract.exe** when OCR is finished for the image (token is fully-buffered).  In other words, tokens will not be flushed until tesseract.exe is killed, because they are only buffered.

Based on research, this is due to the way buffering behaves differently for following streams:
![buffering](https://user-images.githubusercontent.com/55756666/72619873-3af8dc00-390c-11ea-8e22-d0f1fb21de69.png)
Since Java communicates with the child process over a pipe, it falls into the “stdout (not a TTY)” category.
More ref: https://eklitzke.org/stdout-buffering

I was able to resolve by adding flush command in source code. (see below)

### Expected Behavior:
tesseract.exe flushes tokens to stdout when finished OCR for the given image.

### Suggested Fix:
1.  Add ``fflush(fout_);`` at the end of ``TessResultRenderer::AppendData(const char* s, int len)`` (_src/api/renderer.cpp_)
2.  Add ``fflush(stderr);`` after line 1017 (_src/api/renderer.cpp_) - https://github.com/tesseract-ocr/tesseract/blob/master/src/api/baseapi.cpp#L1017

stweil:
Do you want to send a pull request which adds the missing `fflush` in `TessResultRenderer::AppendData`?

Is flushing for `stderr` really needed? Do you have a practical example which fails without that?

jkang-eng:
> Do you want to send a pull request which adds the missing `fflush` in `TessResultRenderer::AppendData`?
> 
> Is flushing for `stderr` really needed? Do you have a practical example which fails without that?

Sure, I can do that.

Regarding ``stderr``, I can't recall top of my head, but there was an issue with specific image if without flush command.  (I think it hanged..)

jkang-eng:
> Do you want to send a pull request which adds the missing `fflush` in `TessResultRenderer::AppendData`?

Hi @stweil,
I'm trying to push changes to remote branch, but getting following messages.
```
[...]

PR #2865, opened 2020-01-21
Fix "tesseract.exe not flushing stdout/stderr" (Issue #2859)

jkang-eng:
In my application, I spawn tesseract.exe from java and continuously send request (path to image) / receive response  (tokens) through standard pipe-stdio.

For example, tesseract is launched with following args.
```
tesseract.exe stdin stdout -c stream_filelist=true
```

Problem with the existing code, **tokens are never flushed from  tesseract.exe** when OCR is finished for the image (token is fully-buffered).  In other words, tokens will not be flushed until tesseract.exe is killed, because they are only buffered.

Based on research, this is due to the way buffering behaves differently for following streams:
buffering
Since Java communicates with the child process over a pipe, it falls into the “stdout (not a TTY)” category.
More ref: https://eklitzke.org/stdout-buffering

I was able to resolve by adding flush command in source code.

Reference: https://github.com/tesseract-ocr/tesseract/issues/2859

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1015,7 +1015,6 @@ bool TessBaseAPI::ProcessPagesFileList(FILE *flist,
       return false;
     }
     tprintf("Page %d : %s\n", page, pagename);
-    fflush(stderr);
     bool r = ProcessPage(pix, page, pagename, retry_config,
                          timeout_millisec, renderer);
     pixDestroy(&pix);
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1015,6 +1015,7 @@ bool TessBaseAPI::ProcessPagesFileList(FILE *flist,
       return false;
     }
     tprintf("Page %d : %s\n", page, pagename);
+    fflush(stderr);
     bool r = ProcessPage(pix, page, pagename, retry_config,
                          timeout_millisec, renderer);
     pixDestroy(&pix);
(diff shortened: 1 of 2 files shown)


jkang-eng:
Hi @stweil, this is my first PR to tesseract repository. :-)
Who should I include as reviewers?  Also, is there any PR standard that I have to conform to (PR title, commit messages, etc)

src/api/baseapi.cpp
```diff
@@ -1015,6 +1015,7 @@ bool TessBaseAPI::ProcessPagesFileList(FILE *flist,
       return false;
     }
     tprintf("Page %d : %s\n", page, pagename);
+    fflush(stderr);
```
stweil: I suggest to keep that code unchanged unless it's clear that the new code is really needed. As `stderr` is normally unbuffered, flushing should not be necessary.

jkang-eng: reverted change.

stweil:
> Who should I include as reviewers?

This is optional, so normally it is not necessary to suggest a reviewer, but of course you are free to do that.

> Also, is there any PR standard that I have to conform to (PR title, commit messages, etc)

There exist no strict rules for Tesseract pull requests. Just have a look on previous commit messages. When referring to issues, it is not necessary to repeat the information from there.

dmigowski:
Hello,

I still have problems with flushing, and I saw that the change provided in this patch tried to modify baseapi.cpp.

      return false;
    }
    tprintf("Page %d : %s\n", page, pagename);
    fflush(stderr);  <-- This was proposed

But the problem here is that it wasn't seen that tprintf writes to stdout, and I need exactly this line flushed in streaming mode to start providing now work to the tesseract (I pool 8 instances for more speed and processor effeciency). 

Could someone please provide a patch that flushes stdout here? Thanks a lot, I am sadly not very into git.

PR #2871, opened 2020-01-24
ChoiceIterator: skip symbol without choices

bertsky:
This fixes a segfault when running with `lstm_choice_mode=2` and querying the choice iterator.


bertsky:
> This would also fix it:
> 
> ```
>  if (word_res_->CTC_symbol_choices[0].empty() ||
>      strcmp(word_res_->CTC_symbol_choices[0][0].first, " ")) {
> ```

Indeed it would.

> Are you sure that your fix is the right one? And I wonder whether symbols without a choice may exist at all or whether there is a bug somewhere else.

Probably just a follow-up, yes.

PR #2894, opened 2020-02-23
Report errors from curl_easy functions

stweil:

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1149,13 +1149,33 @@ bool TessBaseAPI::ProcessPagesInternal(const char* filename,
     } else {
       CURLcode curlcode;
       curlcode = curl_easy_setopt(curl, CURLOPT_URL, filename);
-      ASSERT_HOST(curlcode == CURLE_OK);
+      if (curlcode != CURLE_OK) {
+        fprintf(stderr, "Error, curl_easy_setopt failed with error %s\n",
+                curl_easy_strerror(curlcode));
+        curl_easy_cleanup(curl);
+        return false;
+      }
       curlcode = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, WriteMemoryCallback);
-      ASSERT_HOST(curlcode == CURLE_OK);
+      if (curlcode != CURLE_OK) {
+        fprintf(stderr, "Error, curl_easy_setopt failed with error %s\n",
+                curl_easy_strerror(curlcode));
+        curl_easy_cleanup(curl);
+        return false;
+      }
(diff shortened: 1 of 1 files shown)


egorpugin:
Wow, this definitely must be a lambda.

stweil:
> Wow, this definitely must be a lambda.

I tried that now, but I'm not sure whether it can be simplified further. Suggestions are welcome.

src/api/baseapi.cpp
```diff
@@ -1148,14 +1148,28 @@ bool TessBaseAPI::ProcessPagesInternal(const char* filename,
       return false;
     } else {
       CURLcode curlcode;
+      auto error = [curl, curlcode](const char* function) {
```
egorpugin: Capture `curlcode` by ref? Otherwise it won't be changed.

egorpugin:
Looks better now.

src/api/baseapi.cpp
stweil: It isn't changed, so I see no reason to capture it by reference.

egorpugin: I mean it is used in lambda. And you capture some trash stack value.
`curl_easy_strerror(curlcode));`

zdenop:
thanks

egorpugin:
Why? Issue was not fixed.
```CURLcode curlcode; -> you create integer on stack with track in it
auto error = [curl, curlcode](const char* function) -> now you copy that trash into function
```
and what you expect `curl_easy_strerror(curlcode aka TRASH)` will return?

zdenop:
ups. I am sorry - misunderstood your comments...

stweil:
> I mean it is used in lambda. And you capture some trash stack value.

I don't think there is a trash stack value, and the code works as expected.

egorpugin:
Ok, why?

stweil:
The lambda function uses `curl` and `curlcode` from the surrounding function.

egorpugin:
curlcode in the line before lambda is used. Do you agree?

stweil:
That's right, but the lambda function is only used after both local variables have valid values.

egorpugin:
But lambda captures values on its creation right after variable is declared.

stweil:
The compiler agrees:

    baseapi.cpp:1151:27: warning: variable 'curlcode' is uninitialized when used here [-Wuninitialized]

I fixed that now, see commit 9f5a3f6ac790212a124fcbf94413b98a2418a6fd. Thanks for correcting me.

egorpugin:
?????????
It is still wrong, just add & in front of curlcode capture.

Now you will capture only first result and others won't update it again.

stweil:
Thanks again. See commit 4a00b68c6363cb2126be360ff97c1410634108b6.

PR #2952, opened 2020-04-19
[trie.h] pattern definition: fix documentation

jannick0:
The fix makes the definition of `\n` consistent with the examples given below the definition.  Please note that I did not check this against how it is implemented in the code.


amitdo:
https://github.com/tesseract-ocr/tesseract/blob/912c9978dacf9f7a7cc97ba9a906b40e3df8fc09/src/dict/trie.cpp#L380

Issue #2926, opened 2020-03-18
The buffer unicode_repr in tesseract-4.1/src/api/baseapi.cpp at line 2486 does not have a null terminator


agarwlanurag:
Got this feedback when tried to use this source code.

The buffer unicode_repr in tesseract-4.1/src/api/baseapi.cpp at line 2486 does not have a null terminator

Can you please let me know how can i remediate it.

agarwlanurag:
![image](https://user-images.githubusercontent.com/14285773/76973126-e5d64800-6905-11ea-8cec-e3e0f41a2cab.png)

amitdo:
Why do you ignore the instructions given to everyone when posting issues?

amitdo:
Tesseract itself (the command line tool) does not use this code since at least 2011.

`TesseractExtractResult` that uses this code is also not used by tesseract. It is obsolete. Don't use it. 

Use this instead:

`ResultIterator *it = GetIterator();`

amitdo:
https://groups.google.com/forum/?hl=en#!searchin/tesseract-dev/TesseractExtractResult%7Csort:date/tesseract-dev/l2ZFrpgYkSc/GFkYhng64yEJ



>Ray 3/4/12
>
>PageIterator and ResultIterator should *definitely* be added to the list of exported/visible classes.
These classes provide the main API to get at detailed recognition results.
Conversely, forget TesseractExtractResult and everything else below the OCROpus cut-line.

amitdo:
@zdenop, @stweil,

Can we remove "everything else below the OCROpus cut-line"?

zdenop:
AFAIK OCROpus  does not use tesseract  anymore, so from this point it could be removed.

stweil:
I'll prepare a pull request.

stweil:
See PR #2957.

PR #2957, opened 2020-04-27
Remove old code which was used for Ocropus

stweil:


stweil:
This should fix issue #2926.

zdenop:
thanks

amitdo:
SetFillLatticeFunc(FillLatticeFunc() was not part of the ocropus stuff.

amitdo:
https://github.com/tesseract-ocr/tesseract/search?q=fill_lattice_&unscoped_q=fill_lattice_

stweil:
@amitdo, maybe the code which refers to `fill_lattice_` could be removed, too. Or is there someone who uses that?

amitdo:
Maybe that code is/was used by Google internal version.

amitdo:
Here's another SetSomethingFunc

https://github.com/tesseract-ocr/tesseract/blob/23be532f7d91c88fd3d3c32199526438b3990092/src/api/baseapi.cpp#L1993

[probability_in_context_](https://github.com/tesseract-ocr/tesseract/search?q=probability_in_context_)


https://groups.google.com/forum/?hl=en#!topic/tesseract-dev/YsJogVVFbGk


>Ray 9/9/14
>
>There is no code in tesseract to implement or train an n-gram model, only to make use of one.
>
>There is a "hook" function and code to set it to an external function.
TessBaseAPI::SetProbabilityInContextFunc is used to set it.
>
>We experimented with it in Google with some existing Google-specific n-gram library. It didn't help very much in most languages, although it does in cases where the dictionary is particularly weak for some reason, like CJK.
>
>A single function is required that returns the probability of the current "character" (in general a utf-8 string) given the context of a previous utf-8 string.
It is fairly simple to write and train, given appropriate training data.
>
>The difficulty is with balancing the strength of the n-gram model and the shape classifier confidences. For speech recognition, where typical system accuracies are 80-90%, it is easy to use language models to combine with a poor accuracy phoneme classifier, but in OCR, where typical word accuracies exceed 95% (for alphabetic, word-segmented languages) language models of any kind can easily start replacing infrequent, but correct results with more frequent, but wrong results, and nobody has properly solved the problem in a way that is mathematically and statistically rigorous. Looking at OCR output, it is easy to spot cases where language models could help, but not so easy to see the already-correct results that would be made worse.

amitdo:
@stweil,

GetPageRes() is used in unittest/baseapi_test.cc

stweil:
Thank you. Sorry that I missed that. Pull request #2962 should fix it.

PR #2962, opened 2020-05-04
Add TessBaseAPI::GetPageRes again

stweil:
It is now added unconditionally, so it is always available for the unittest.


zdenop:
thanks.

Issue #2333, opened 2019-03-16
Define minimal tesseract dependancies


zdenop:
I tried to build (cmake&clang) tesseract with minimal dependencies, so I build leptonica without any dependencies (not even zlib). Build finished fine, ` tesseract -v` produced:
```
tesseract 4.1.0-rc1-97-g681e
 leptonica-1.78.0 (Mar 13 2019, 19:12:40) [MSC v.1915 LIB Release x64]
  (null)
 Found AVX2
 Found AVX
 Found SSE
```

But when I run `tesseract test.pnm -`  it produced following error messages and than it crashed:
```
Error in pixReadMemTiff: function not present
Error in pixReadMem: tiff: no pix returned
Error in pixaGenerateFontFromString: pix not made
Error in bmfCreate: font pixa not made
Warning: Invalid resolution 0 dpi. Using 70 instead.
Estimating resolution as 333
Error in pixWriteMemPng: function not present
```
These indicates that there are several **must dependencies** for running tesseract (tiff, png and therefore zlib...) and they have to be checked (bmfCreate and pixReadMem)  during configuring build.

More advance approach would to analyze if relevant code could be skipped - but than we should somehow distribute information about not available API (e.g. [`ProcessPagesMultipageTiff`](https://github.com/tesseract-ocr/tesseract/blob/ea3b806357f5e23097f4ae9ac8c5b7d756b02e3e/src/api/baseapi.cpp#L1036)).

amitdo:
Looks like a bug in baseapi.cpp.

stweil:
Leptonica is normally used as a shared library. So it is possible to provide a minimal `liblept.so` which obviously causes a crash or to use a full `liblept.so` which will work fine.

Therefore I think it is not a build problem, but a problem of the error handling at runtime (which should also be fixed, of course).

amitdo:
Leponica can handle pnm format without any dependency.

Shreeshrii:
what about adding minimum versions of python (3.6) and bash (4.4) for training tools?

zdenop:
@Shreeshrii: we can put  info about minimum dependencies to README,md. 
Maybe I found some time to rewrite python script to 3.5 compatible version (it should not be a big issue) or at least to check python version at the begin...

zdenop:
[...]

PR #2965, opened 2020-05-04
Avoid tesseract writing Pix out/reading them back.

robinwatts:
By default, when we ImageData::SetPix, we write the data out as a
PNG, just to read it back in to get a compressed buffer of data.
We then use this to generate a new Pix.

In builds of Tesseract on systems where we don't have temp files,
writing files out is problematic.

Not only that, but compressing/uncompressing is slow, and on minimal
builds of leptonica, where we've disabled the format writers to reduce
memory footprint, we get no compression anyway.

In such cases, it'd be far nicer just to keep the original Pix as
the internal data.

Also, when recovering the pixmap from the ImageData, if we know we're
only going to read from the data, we can avoid duplicating it and
just use the original. This is exactly the case when GRAPHICS_DISABLED
is set.

So, introduce a TESSERACT_IMAGEDATA_AS_PIX predefine that we can use
to cause the internal data to be a Pix rather than a compressed
buffer.

Given we don't do compression, and they were writing to memory,
this was all just more effort than we needed.

Also, if we're using GRAPHICS_DISABLED, we might as well just
pixCopy rather than pixClone as only the scaler uses this.

### DIFF
--- a/src/ccstruct/imagedata.cpp
+++ b/src/ccstruct/imagedata.cpp
@@ -207,12 +207,28 @@ bool ImageData::SkipDeSerialize(TFile* fp) {
 // In case of missing PNG support in Leptonica use PNM format,
 // which requires more memory.
 void ImageData::SetPix(Pix* pix) {
+#ifdef TESSERACT_IMAGEDATA_AS_PIX
+  internal_pix_ = pix;
+#else
   SetPixInternal(pix, &image_data_);
+#endif
 }
 
 // Returns the Pix image for *this. Must be pixDestroyed after use.
 Pix* ImageData::GetPix() const {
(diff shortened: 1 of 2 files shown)


robinwatts:
I am looking into using Tesseract with Ghostscript. These are some changes that I made locally to smooth operation. For systems without fmemopen (such as windows), tesseract drives leptonica to save images out, just so it can then read the compressed version back in again.

This commit enables a build define that avoids that.

zdenop:
I like the concept to build minimal tesseract (I tried to do it [some time ago](https://github.com/tesseract-ocr/tesseract/issues/2333) AFAIK only zlib was needed).

PR #2966, opened 2020-05-04
Add a mechanism to avoid creating debug fonts.

robinwatts:
If TESSERACT_DISABLE_DEBUG_FONTS is defined, tesseract doesn't
atetmpt to create any debug fonts. This not only saves memory,
but it (combined with the change to optionally use Pix as
internal storage for the ImageData) allows us to use an
embedded Leptonica library with no format handlers at all.

### DIFF
--- a/src/ccstruct/debugpixa.h
+++ b/src/ccstruct/debugpixa.h
@@ -12,7 +12,11 @@ class DebugPixa {
   // TODO(rays) add another constructor with size control.
   DebugPixa() {
     pixa_ = pixaCreate(0);
+#ifdef TESSERACT_DISABLE_DEBUG_FONTS
+    fonts_ = NULL;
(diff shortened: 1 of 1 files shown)


robinwatts:
Again, this is a change from my integration with Ghostscript. For embedded systems, we can minimise the size of the leptonica library used by avoiding the creation of fonts to be used for debugging. This removes the need for TIF support in the leptonica lib.

zdenop:
Does not Ghostscript use TIF?

robinwatts:
"Does not Ghostscript use TIF?"

It does, optionally. It's possible to build gs (particularly for embedded systems) where TIF input/output isn't used.

Also, MuPDF, somewhere else we may be using this, doesn't use libtiff at all.

PR #2968, opened 2020-05-06
Ensure PACKAGE_VERSION is defined, even without config.h

robinwatts:
If we aren't using config.h, then PACKAGE_VERSION is not
defined. Spot this, and include tess_version.h to get
TESSERACT_VERSION_STR instead.

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -57,7 +57,6 @@
 #include "blobclass.h"         // for ExtractFontName
 #endif
 #include "boxword.h"           // for BoxWord
-#include "config_auto.h"       // for PACKAGE_VERSION
 #include "coutln.h"            // for C_OUTLINE_IT, C_OUTLINE_LIST
 #include "dawg_cache.h"        // for DawgCache
 #include "dict.h"              // for Dict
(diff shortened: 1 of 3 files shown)


stweil:
Tesseract uses `config_auto.h`, not `config.h`. Why do you want to compile without `config_auto.h`?

robinwatts:
"Tesseract uses config_auto.h, not config.h."

Yes, indeed, that's a typo. If you set the HAVE_CONFIG_H define, then tesseract does indeed look for a config_auto.h file.

Given that tesseract requires you to set HAVE_CONFIG_H to use config_auto.h, it seems reasonable that it should compile without HAVE_CONFIG_H being set (otherwise, why check that define at all?)

"Why do you want to compile without config_auto.h?"

Because for various reasons I'm not using the autoconf based build.

Now, it's entirely reasonable that this fix isn't right. Possibly we should ALWAYS be including tess_version.h and always using TESSERACT_VERSION_STR. That'd seem neater to me. I'm happy to prepare a patch like that if that's what it takes to get a fix committed.

stweil:
I suggest to drop this pull request and add a `config_auto.h` to your code.

robinwatts:
Actually, there is a nicer way to do this. Let me have one last go at getting your approval :)

robinwatts:
There you go, this version seems much nicer.

PR #2963, opened 2020-05-04
Add build script for oss-fuzz fuzzers (improved copy from google/oss-fuzz)

stweil:

PR #2969, opened 2020-05-06
Moved oss-fuzz build script to Tesseracts own repository

AdamKorcz:
This PR moves the [oss-fuzz build](https://github.com/google/oss-fuzz/blob/master/projects/tesseract-ocr/build.sh) script over to Tesseracts own repository. No changes have been made to the build-script besides removing the license header that won't be necessary.

If the PR is merged, I will adjust the build script in oss-fuzz's repository accordingly to run in a similar fashion as [libreoffice's build script](https://github.com/google/oss-fuzz/blob/master/projects/libreoffice/build.sh).

Having the build script in the projects own repository is recommended for maintainers and contributors to easier maintain the oss-fuzz integration. Build scripts become outdated, and fixing a broken build is easier and faster with the current setup.

Other examples of projects on oss-fuzz having this setup are: [curl](https://github.com/google/oss-fuzz/blob/master/projects/curl/build.sh), [leptonica](https://github.com/google/oss-fuzz/blob/master/projects/leptonica/build.sh) & [lz4](https://github.com/google/oss-fuzz/blob/master/projects/lz4/build.sh)

### DIFF
--- a/unittest/fuzzers/oss-fuzz-build.sh
+++ b/unittest/fuzzers/oss-fuzz-build.sh
@@ -1,0 +1,41 @@
+cd $SRC/leptonica
+./autogen.sh
+./configure
+make -j$(nproc)
+make install
+ldconfig
+
+cd $SRC/tesseract
+./autogen.sh
+CXXFLAGS="$CXXFLAGS -D_GLIBCXX_DEBUG" ./configure --disable-graphics --disable-shared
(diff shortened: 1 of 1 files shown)


stweil:
See PR #2963 which does nearly the same.

AdamKorcz:
Ah right - makes sense. Closing this PR.

PR #2974, opened 2020-05-11
Simplify intsimdmatrixavx2

robinwatts:
intsimdmatrixavx2 has some strange code in the
PartialMatrixDocVectorXX functions, whereby the final
ExtractResults can be run more than kNumOutputsPerRegister
times.

There *are* only kNumOutputsPerRegister result values to
extract, so all this ends up doing is pulling extra 0's out
and writing them to memory.

That memory is then overwritten by subsequent calls to
later PartialMatrixDocVectorXX calls, so it actually serves
no purpose.

Remove this special code. This should not only speed up the
runtimes as fewer needless operationations are happening, but
may also help due to improved compile time optimisation (as
the loop sizes are fixed).

### DIFF
--- a/src/arch/intsimdmatrixavx2.cpp
+++ b/src/arch/intsimdmatrixavx2.cpp
@@ -29,7 +29,7 @@
 #include <algorithm>
 #include <vector>
 
-#include <string.h>
+#include <cstring> // memcpy
 
 namespace tesseract {
 
--- a/src/arch/intsimdmatrixavx2.cpp
+++ b/src/arch/intsimdmatrixavx2.cpp
@@ -29,6 +29,8 @@
 #include <algorithm>
 #include <vector>
 
+#include <string.h>
+
 namespace tesseract {
 
 // Number of outputs held in each register. 8 x 32 bit ints.
@@ -108,7 +110,7 @@ static inline void ExtractResults(__m256i& result, __m256i& shift_id,
 // u must be padded out with zeros to
 // kNumInputsPerGroup*ceil(num_in/kNumInputsPerGroup) elements.
 static void PartialMatrixDotVector64(const int8_t* wi, const double* scales,
-                                     const int8_t* u, int num_in, int num_out,
+                                     const int8_t* u, int num_in,
                                      double* v) {
   // Register containing 16-bit ones for horizontal add with 16->32 bit
   // conversion.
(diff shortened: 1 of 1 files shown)


robinwatts:
Rebased to current master.

stweil:
I am afraid that this is the wrong kind of optimization. The NEON code is similar and causes a heap buffer overflow which is detected when building with sanitizers:

    SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../../../src/ccstruct/coutln.cpp:1016:19 in 
    =================================================================
    ==20250==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xaed03bf8 at pc 0x00dc23b8 bp 0xbed353c8 sp 0xbed353c4
    READ of size 8 at 0xaed03bf8 thread T0
        #0 0xdc23b4 in tesseract::PartialMatrixDotVector8(signed char const*, double const*, signed char const*, int, double*) /tesseract/src/arch/intsimdmatrixneon.cpp:118:77
        #1 0xdc23b4 in tesseract::matrixDotVector(int, int, signed char const*, double const*, signed char const*, double*) /tesseract/src/arch/intsimdmatrixneon.cpp:137:5
        #2 0xdb1b90 in tesseract::WeightMatrix::MatrixDotVector(signed char const*, double*) const /tesseract/src/lstm/weightmatrix.cpp:254:5

robinwatts:
I convinced myself this couldn't happen :(
I tried running with asserts enabled/inserted and couldn't trip it up.
Could I trouble you for the exact build/test steps you used to see the problem, and I'll look into it? Thanks.

stweil:
My testcase is based on files which are normally at hand, namely eng.traineddata (fast) and the largest image in the test set provided with Tesseract:

    ./tesseract test/testing/8087_054.3G.tif -

Initially I just added a `printf` statement to see which values for `num_out` were fed into `ExtractResults`. Most times the value was 8, but sometimes it was called with 7.

If you compile with `-O0`, you can also use gdb and a conditional breakpoint to see the cases with `num_out == 7`.

The build with sanitizers was configured on Debian (amd64) like this:

    ./configure --disable-openmp --disable-shared --enable-debug CXX=clang++-11 CXXFLAGS='-g -O0 -Wall -Wextra -Wpedantic -D_GLIBCXX_DEBUG -fsanitize=fuzzer-no-link,address,undefined'

It shows a heap buffer overflow (and two other runtime errors which already exist for a long time):

    ==4324==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6180000053f8 at pc 0x000001983d37 bp 0x7fff13354390 sp 0x7fff13354388
    READ of size 8 at 0x6180000053f8 thread T0
        #0 0x1983d36 in tesseract::ExtractResults(long long vector[4]&, long long vector[4]&, signed char const*&, double const*&, int, double*&) /tesseract/src/arch/intsimdmatrixavx2.cpp:92:60

stweil:
@robinwatts, I used the same test case for NEON and also a similar configuration, but an older CLANG version on Ubuntu bionic. If you need details, I can send them, too.

robinwatts:
I could reproduce the problem on windows and I've fixed it.

Here is a fixed version. I can't get clang-11 here on the Ubuntu 18.04 box I'm using, so I did the same test with clang-10. That appears to run clean now.

If this still doesn't work, then let me know and I'll have another go. Thanks.

robinwatts:
Rebuilt, and I'm seeing a problem with the sanitizers build, so making draft.

robinwatts:
Fixed, I believe.

src/arch/intsimdmatrixavx2.cpp
```diff
@@ -29,6 +29,8 @@
 #include <algorithm>
 #include <vector>
 
+#include <string.h>
```
stweil, line 32:
```suggestion
#include <cstring> // memcpy
```

stweil:
@robinwatts, are you sure that the commit message describes the old code correctly?

And did you compare old and new performance?

robinwatts:
> @robinwatts, are you sure that the commit message describes the old code correctly?

No, probably not. Let me squash the two commits and fix the commit message.

robinwatts:
Damn, this code is still wrong. I need to rework this. Sorry.

robinwatts:
This was intended to be the first of several optimisations building to something. This first step was broken, but overall the optimisations it built to was valid. I've opened https://github.com/tesseract-ocr/tesseract/pull/2994 with the reworked series of commits giving a net speedup.

PR #2975, opened 2020-05-11
Tweak architecture specific SIMD files for ease of compilation.

robinwatts:
This won't affect anything using the supplied build system. For
other projects that include tesseract within them, however, this
may make their life easier.

For example, I have an integration of Tesseract with Ghostscript,
in which tesseract is built as part of the Ghostscript build,
without using the tesseract build system.

The Ghostscript build system is makefile based, and has to work
on a range of make systems, including unix make, gnu make and
nmake. As such we have to avoid conditionals in the common
makefiles. It therefore becomes hard to build one set of files on
x86 systems, and another on (say) ARM systems.

Accordingly, this commit makes small tweaks to the architecture
specific files, so that they compile on EVERY platform; just they
only compile to anything useful on the appropriate platform.

Thus the makefiles can build all the files on all the systems, and
the preprocessor flags mean that the correct functions are actually
built.

### DIFF
--- a/src/arch/dotproductavx.cpp
+++ b/src/arch/dotproductavx.cpp
@@ -16,8 +16,10 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__AVX__)
-#error Implementation only for AVX capable architectures
-#endif
+ #if defined(__i686__) || defined(__x86_64__)
+  #error Implementation only for AVX capable architectures
+ #endif
+#else
 
 #include <immintrin.h>
 #include <cstdint>
@@ -57,3 +59,5 @@ double DotProductAVX(const double* u, const double* v, int n) {
 }
 
 }  // namespace tesseract.
+
+#endif
--- a/src/arch/dotproductfma.cpp
+++ b/src/arch/dotproductfma.cpp
@@ -16,8 +16,10 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__FMA__)
-#error Implementation only for FMA capable architectures
-#endif
+ #if defined(__i686__) || defined(__x86_64__)
+  #error Implementation only for FMA capable architectures
+ #endif
+#else
 
 #include <immintrin.h>
 #include <cstdint>
@@ -55,3 +57,5 @@ double DotProductFMA(const double* u, const double* v, int n) {
 }
 
 }  // namespace tesseract.
+
+#endif
--- a/src/arch/dotproductsse.cpp
+++ b/src/arch/dotproductsse.cpp
@@ -16,8 +16,10 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__SSE4_1__)
-#error Implementation only for SSE 4.1 capable architectures
-#endif
+ #if defined(__i686__) || defined(__x86_64__)
+  #error Implementation only for SSE 4.1 capable architectures
+ #endif
+#else
 
 #include <emmintrin.h>
 #include <smmintrin.h>
@@ -79,3 +81,5 @@ double DotProductSSE(const double* u, const double* v, int n) {
 }
 
 }  // namespace tesseract.
+
+#endif
--- a/src/arch/intsimdmatrixavx2.cpp
+++ b/src/arch/intsimdmatrixavx2.cpp
@@ -17,8 +17,10 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__AVX2__)
-#error Implementation only for AVX2 capable architectures
-#endif
+ #if defined(__i686__) || defined(__x86_64__)
+  #error Implementation only for AVX2 capable architectures
+ #endif
+#else
 
 #include "intsimdmatrix.h"
 
@@ -340,3 +342,5 @@ const IntSimdMatrix IntSimdMatrix::intSimdMatrixAVX2 = {
 };
 
 }  // namespace tesseract.
+
+#endif
--- a/src/arch/intsimdmatrixsse.cpp
+++ b/src/arch/intsimdmatrixsse.cpp
@@ -16,8 +16,10 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__SSE4_1__)
-#error Implementation only for SSE 4.1 capable architectures
-#endif
+ #if defined(__i686__) || defined(__x86_64__)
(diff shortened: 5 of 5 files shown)


src/arch/dotproductavx.cpp
```diff
@@ -16,8 +16,9 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__AVX__)
-#error Implementation only for AVX capable architectures
-#endif
+  //#error Implementation only for AVX capable architectures
+  int dotproductavx_empty_compilation_unit;
```
stweil, line 20: Please don't define unused variables. They will trigger compiler warnings and be reported when searching for unused code. It is better to make an empty compilation unit.

src/arch/dotproductavx.cpp
```diff
@@ -16,8 +16,9 @@
 ///////////////////////////////////////////////////////////////////////
 
 #if !defined(__AVX__)
```
stweil, line 18: How do you make sure that this code is compiled with AVX compiler options for x86 targets?

stweil:
Similar comments apply for all modified files.

src/arch/dotproductavx.cpp
robinwatts: I have used systems in the past where empty compilation units give errors. I've seen such dummy variables used in other projects in the past, so I'm sure I'm not alone in this.

The dummy variable is a global, and therefore cannot be detected as unused by static analysis tools. In order for tools to spot it as unused, they'd actually have to do the compile and link, and even then they'd have to be very confident that it wasn't going to be referenced by a caller of Tesseract.

I will happily remove the variables if you want though, and we can put them back in if we hit such a system in future.

src/arch/dotproductavx.cpp
robinwatts: __AVX__ is defined by the compiler (gcc) if -mavx is passed to it. On *nix the gs configure script sets flags which pass down from the top-level Makefile (which contains conditionals etc) through the platform-independent makefiles to our tesseract.mak file.

On Windows, our top-level msvc.mak file predefines the appropriate things in the CFLAGS.

Getting flags through the make system is relatively simple. Selecting between different sets of files to build is trickier.

If you're really interested, the whole nightmarish complexity of the changes to the Ghostscript build can be seen within the commit here: https://git.ghostscript.com/?p=user/robin/ghostpdl.git;a=commitdiff;h=e8fb4eb5f7e7d6514d7969e94a909654fe690fdf

(at least until I revise the commit and git garbage collects it)

src/arch/dotproductavx.cpp
stweil: The compiler will complain because each global symbol is expected to be declared in a header file.

src/arch/dotproductavx.cpp
stweil: Those files are special because they require compiler flags which must not be used for other files. Example: if you compile all code with `-mavx2`, the resulting binary will no longer work on a host without AVX2.

You are using a normal `autoconf` build as far as I see. Then I suggest to build Leptonica and Tesseract as subprojects with their own `autoconf` build process and simply use the resulting libraries for your build.

What is the use of duplicating build rules?

src/arch/dotproductavx.cpp
egorpugin: You can try to write `static void dummy(){}`.

robinwatts: stweil: Good point. I'll do a revised commit that removes them, and we can worry about it again if a platform pops up where it's an issue.

egotpugin: static void dummy(void) {} would definitely get picked up by static analysis tools as being unused, though?

src/arch/dotproductavx.cpp
robinwatts: On *some* platforms we use an autoconf based build, but not on all. Windows is the classic example, but there are others.

Yes, I know about -mavx2 only being used for the avx2 files etc. My code does that correctly.

src/arch/dotproductavx.cpp
egorpugin: Don't know. Maybe add some attribute. But, it's better than variable and seen in other projects.

robinwatts: I am happy to remove the ints and either to add "static void dummy(void) {}" or not as you guys prefer.

If you could agree on the preferred approach, I'll gladly implement it. Thanks :)

stweil: I clearly prefer the empty compilation unit.

egorpugin: I'm fine with empty/function. But some compielrs warn about empty files.

stweil:
Would it be sufficient to replace `#error` by `#warning`? Or would that cause problems with the MS compilers?

It would also be acceptable to suppress the `#error` when `__x86_64__` is undefined. That could be the simplest solution.

src/arch/dotproductavx.cpp
stweil: As long as they only warn, that's not a problem. It looks like we cannot avoid warnings from some tools, unless the build process skips those source files.

src/arch/dotproductavx.cpp
```diff
@@ -16,8 +16,8 @@
...
-#error Implementation only for AVX capable architectures
-#endif
+  //#error Implementation only for AVX capable architectures
```
stweil, line 19:
```suggestion
#if defined(__x86_64__)
#error Implementation only for AVX capable architectures
#endif
```

robinwatts: Yup. Working on that now. As I understand it __x86_64__ is only defined when building for 64bit systems. SSE, AVX2 etc work just fine on 32bit systems too - should we be using:

#if defined(__x86__64) || defined(something_else)

to capture both cases?

stweil: That's correct, but i386 systems are getting rarer, and the normal Tesseract build handles them correctly.  Nevertheless I updated my suggestion above to handle `__i686__`, too.

robinwatts: Updated. Please let me know if you'd like any further changes. Thanks.

amitdo:
Regarding empty-translation-unit:

https://github.com/llvm-mirror/clang/blob/master/test/Parser/empty-translation-unit.c

PR #2978, opened 2020-05-12
NEON SIMD code.

robinwatts:
In tests on my pi3b+, a release build of my ghostscript integration
takes 2 minutes 27 seconds to render a PDF and OCR it with the
vanilla sources. With this NEON coded added the time drops to 37
seconds.

I have not tested the configure/Makefile changes as I'm not using
them.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -148,6 +148,12 @@ if $sse41; then
   AC_DEFINE([HAVE_SSE4_1], [1], [Enable SSE 4.1 instructions])
 fi
 
+AX_CHECK_COMPILE_FLAG([-mfpu=neon], [neon=true], [neon=false], [$WERROR])
+AM_CONDITIONAL([HAVE_NEON], $neon)
+if $neon; then
+  AC_DEFINE([HAVE_NEON], [1], [Enable NEON instructions])
+fi
+
 AX_CHECK_COMPILE_FLAG([-march=native], [arch_native=true], [arch_native=false], [$WERROR])
 AM_CONDITIONAL([MARCH_NATIVE_OPT], $arch_native)
 
--- a/src/arch/intsimdmatrixneon.cpp
+++ b/src/arch/intsimdmatrixneon.cpp
@@ -1,0 +1,161 @@
+///////////////////////////////////////////////////////////////////////
+// File:        intsimdmatrixneon.cpp
+// Description: matrix-vector product for 8-bit data on neon.
+// Author:      Robin Watts (from the AVX2 original by Ray Smith)
+//
+// (C) Copyright 2017, Google Inc.
+// (C) Copyright 2020, Artifex Software Inc.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////
+
+#if !defined(__ARM_NEON)
+//#error Implementation only for NEON capable architectures
+#else
+
+#include "intsimdmatrix.h"
+
+#include "arm_neon.h"
+#include <cstdint>
+#include <algorithm>
+#include <vector>
+
+namespace tesseract {
+
+// Number of outputs held in each register. (Actually, we use a
+// pair of 4x32 registers, so 8 x 32 bit ints).
+constexpr int kNumOutputsPerRegister = 8;
+// Maximum number of registers that we will use.
+constexpr int kMaxOutputRegisters = 1;
+// Number of inputs in the inputs register.
+constexpr int kNumInputsPerRegister = 8;
+// Number of inputs in each weight group.
+constexpr int kNumInputsPerGroup = 8;
+
+// Function to compute part of a matrix.vector multiplication. The weights
+// are in a very specific order (see above) in w, which is multiplied by
+// u of length num_in, to produce output v after scaling the integer results
+// by the corresponding member of scales.
+// The amount of w and scales consumed is fixed and not available to the
+// caller.
+
+// Computes part of matrix.vector v = Wu. Computes N=8 results.
+// The weights *must* be arranged so that consecutive reads from wi
+// provides (num_in/kNumInputsPerGroup groups of (N output dim groups of
+// (kNumInputsPerGroup inputs))). After that there must be N consecutive
+// bias weights, before continuing with any more weights.
+// u must be padded out with zeros to
+// kNumInputsPerGroup*ceil(num_in/kNumInputsPerGroup) elements.
+static void PartialMatrixDotVector8(const int8_t* __restrict wi, const double* __restrict scales,
+                                     const int8_t* __restrict u, int num_in,
+                                     double* __restrict v) {
+  // Initialize all the results to 0.
+  int32x4_t result0123 = { 0, 0, 0, 0 };
+  int32x4_t result4567 = { 0, 0, 0, 0 };
+  // Iterate over the input (u), one registerful at a time.
+  for (int j = 0; j < num_in; j += 8) {
+    int8x8_t  vu   = vld1_s8(u);       // vu     = u0  u1  u2  u3  u4  u5  u6  u7
+    int8x16_t vw01 = vld1q_s8(wi);     // vw0    = w00 w01 w02 w03 w04 w05 w06 w07 w10 w11 w12 w13 w14 w15 w16 w17
+    int8x16_t vw23 = vld1q_s8(wi+8*2); // vw2    = w20 w21 w22 w23 w24 w25 w26 w27 w30 w31 w32 w33 w34 w35 w36 w37
+    int8x16_t vw45 = vld1q_s8(wi+8*4); // vw4    = w40 w41 w42 w43 w44 w45 w46 w47 w50 w51 w52 w53 w54 w55 w56 w57
+    int8x16_t vw67 = vld1q_s8(wi+8*6); // vw6    = w60 w61 w62 w63 w64 w65 w66 w67 w70 w71 w72 w73 w74 w75 w76 w77
+
+    int16x8_t vrow0q = vmull_s8(vget_low_s8(vw01),  vu); // vrow0q = vw00.u0 w01.u1 w02.u2 w03.u3 vw04.u4 w05.u5 w06.u6 w07.u7
+    int16x8_t vrow1q = vmull_s8(vget_high_s8(vw01), vu); // vrow1q = vw10.u0 w11.u1 w12.u2 w13.u3 vw14.u4 w15.u5 w16.u6 w17.u7
+    int16x8_t vrow2q = vmull_s8(vget_low_s8(vw23),  vu); // vrow2q = vw20.u0 w21.u1 w22.u2 w23.u3 vw24.u4 w25.u5 w26.u6 w27.u7
+    int16x8_t vrow3q = vmull_s8(vget_high_s8(vw23), vu); // vrow3q = vw30.u0 w31.u1 w32.u2 w33.u3 vw34.u4 w35.u5 w36.u6 w37.u7
+    int16x8_t vrow4q = vmull_s8(vget_low_s8(vw45),  vu); // vrow4q = vw40.u0 w41.u1 w42.u2 w43.u3 vw44.u4 w45.u5 w46.u6 w47.u7
+    int16x8_t vrow5q = vmull_s8(vget_high_s8(vw45), vu); // vrow5q = vw50.u0 w51.u1 w52.u2 w53.u3 vw54.u4 w55.u5 w56.u6 w57.u7
+    int16x8_t vrow6q = vmull_s8(vget_low_s8(vw67),  vu); // vrow6q = vw60.u0 w61.u1 w62.u2 w63.u3 vw64.u4 w65.u5 w66.u6 w67.u7
+    int16x8_t vrow7q = vmull_s8(vget_high_s8(vw67), vu); // vrow7q = vw70.u0 w71.u1 w72.u2 w73.u3 vw74.u4 w75.u5 w76.u6 w77.u7
+
+    int32x4_t vrow0q2 = vpaddlq_s16(vrow0q); // vrow0q2 = vw00.u0+w01.u1 w02.u2+w03.u3 vw04.u4+w05.u5 w06.u6+w07.u7
+    int32x4_t vrow1q2 = vpaddlq_s16(vrow1q); // vrow1q2 = vw10.u0+w11.u1 w12.u2+w13.u3 vw14.u4+w15.u5 w16.u6+w17.u7
+    int32x4_t vrow2q2 = vpaddlq_s16(vrow2q); // vrow2q2 = vw20.u0+w21.u1 w22.u2+w23.u3 vw24.u4+w25.u5 w26.u6+w27.u7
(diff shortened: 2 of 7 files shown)


stweil:
In tests which I had run some time ago (for different architectures, also neon), the compiler generated code was similar fast (or even faster) than hand optimized code. Could you please try `-c dotproduct=native`? That requires the right compiler flags for `src/arch/dotproduct.cpp`.

src/arch/intsimdmatrixneon.cpp
```diff
@@ -0,0 +1,166 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////
+
+#if !defined(__ARM_NEON)
+//#error Implementation only for NEON capable architectures
+#else
```
stweil, line 22:
```suggestion
#if defined(__ARM_NEON)
```

src/arch/simddetect.cpp
```diff
@@ -160,6 +162,11 @@ SIMDDetect::SIMDDetect() {
 #endif
 #endif
 
+#if defined(NEON)
+  /* No runtime detection for NEON yet */
+  neon_available_ = 1;
```
stweil, line 167: So neon will be enabled for all ARM systems, and those without neon will crash at runtime?

amitdo:
>In tests which I had run some time ago (for different architectures, also neon), the compiler generated code was similar fast (or even faster) than hand optimized code.

With autotools, the native code uses -O3, while the the optimized uses -O2. Not a 
fair comparison.

src/arch/simddetect.cpp
robinwatts: Yes. I can't find an easy way to detect it at runtime. Possibly we need a configure option to enable/disable it.

robinwatts:
The 'native' code extends to the "DotProduct" function. I haven't coded that, because it's not called in my runs.

I've coded the "IntSimdMatrix" function, which has no "native" equivalent.

src/arch/simddetect.cpp
amitdo: https://stackoverflow.com/questions/26701262/how-to-check-the-existence-of-neon-on-arm

Shreeshrii:
Also see @robinwatts msg at  https://groups.google.com/d/msgid/tesseract-ocr/0e4552a1-a0d3-4a38-8881-0972bdf3185b%40googlegroups.com.

amitdo:
>Presumably, there is a good reason why the existing code uses doubles rather than floats?

https://github.com/tesseract-ocr/tesseract/issues/943#issuecomment-303239929


Read the section that start with
>DotProduct:
>Float vs double:

@noahmetzger and @stweil have some code that uses float instead of double for x86.

robinwatts:
It looks to me like the discussion in that thread predates the existence of intsimdmatrix.

The argument given for using doubles there was that it's bad to do SIGMA(float*float) as the inaccuracies mount up. When we are using intsimdmatrix (matrixdotvector), we are doing all the multiplication and additions in the integer domain. We only convert to double at the end when we divide/bias/scale.

Thus, while that discussion is interesting, it seems to an ignorant reader such as myself, that circumstances may have moved on. A significant part of the stated reason for using doubles has now been removed, so maybe the decision could/should be revisited?

It's entirely possible that there are still good reasons for using doubles in preference to floats, but it's not clear to me what they are.

Possibly, it might be worth using doubles on some architectures, and floats on others?

robinwatts:
Updated with NEON detection for Android (untested) and Linux (tested on my pi).

stweil:
I tested the patch on ODROID-HC1.

Build:

    ./configure '--disable-openmp' '--disable-shared' 'CXXFLAGS=-Wall -Wextra -Wpedantic -g -O2 -fno-math-errno -march=native -mtune=native -mcpu=native'
    make

Test:

    time -p tesseract test/testing/8087_054.3G.tif 8087_054.3G.tif-neon

The time was reduced from 30 s to 17 s by using NEON, but the OCR result was slightly different:

    --- 8087_054.3G.tif-native.txt	2020-05-16 09:22:40.560168947 +0200
    +++ 8087_054.3G.tif-neon.txt	2020-05-16 09:29:03.007051399 +0200
    @@ -1,10 +1,10 @@
      
     
    -aN
    +oe
     
      
     
    -datecline
    +dateline
     
     “This is Greater Serbia,” yet they didn’t take the city.
     On returning to the region in March I visited the
    @@ -64,7 +64,7 @@
     of Dubrovnik’s inhabitants, he depends on Red Cross
     food parcels to survive. Once a month, any person in
     
    -545 TRAVEL&LEISURE SEPTEMBER 1993
    +545 TRAVELG&LEISURE SEPTEMBER 1993

Maybe this comes from small differences in the dot product calculation. Ideally the OCR should not depend on the hardware used.

robinwatts:
You're right, it shouldn't change the result. I've spotted a silly mistake. I'll retest and get a new version up.

robinwatts:
Updated version.

In the original version, due to a misplacement of a bracket, I was doing the division by INT8_MAX before the cast to double. Fixed here.

I also tweaked the code so that the loads happen as loads of q registers rather than d registers, and added some __restricts.

This tests out for me - I hope it solves the problems you were seeing. If not, I will look again. Thanks!

robinwatts:
Rebased to current master.

stweil:
@robinwatts, I am currently rather busy, but will have a look on the latest version and test it later this week.

robinwatts:
No problem, no hurry. I appreciate all your time and efforts! Thanks.

src/arch/simddetect.cpp
```diff
@@ -65,6 +75,8 @@ bool SIMDDetect::avx512BW_available_;
 bool SIMDDetect::fma_available_;
 // If true, then SSe4.1 has been detected.
 bool SIMDDetect::sse_available_;
+// If true, then Neon has been detected.
+bool SIMDDetect::neon_available_;
```
stweil, line 79:
```suggestion
#if defined(HAVE_NEON)
// If true, then Neon has been detected.
bool SIMDDetect::neon_available_;
#endif
```

src/arch/simddetect.cpp
```diff
@@ -160,6 +172,29 @@ SIMDDetect::SIMDDetect() {
...
+                         ANDROID_CPU_ARM_FEATURE_NEON);
+#if 0
+    /* There is no NEON flag to test on ARM64. A wild guess would have
+     * it being the ASIMD flag, but I'll leave this disabled for now. */
+    else if (family == ANDROID_CPU_FAMILY_ARM64)
+      neon_available_ = (android_getCpuFeatures() &
+                         ANDROID_CPU_ARM64_FEATURE_ASIMD);
+#endif
+    else
+      neon_available_ = 0;
```
stweil, line 192:
```suggestion
      neon_available_ = false;
```

src/arch/simddetect.h
```diff
@@ -74,6 +78,8 @@ class SIMDDetect {
   static TESS_API bool fma_available_;
   // If true, then SSe4.1 has been detected.
   static TESS_API bool sse_available_;
+  // If true, then NEON has been detected.
+  static TESS_API bool neon_available_;
```
stweil, line 82:
```suggestion
#if defined(HAVE_NEON)
  // If true, then NEON has been detected.
  static TESS_API bool neon_available_;
#endif
```

src/arch/simddetect.h
```diff
@@ -54,6 +54,10 @@ class SIMDDetect {
   static inline bool IsSSEAvailable() {
     return detector.sse_available_;
   }
+  // Returns true if NEON is available on this system.
+  static inline bool IsNEONAvailable() {
+    return detector.neon_available_;
+  }
```
stweil, line 60:
```suggestion
#if defined(HAVE_NEON)
  // Returns true if NEON is available on this system.
  static inline bool IsNEONAvailable() {
    return detector.neon_available_;
  }
#endif
```

src/arch/intsimdmatrixneon.cpp
```diff
@@ -0,0 +1,162 @@
+///////////////////////////////////////////////////////////////////////
+// File:        intsimdmatrixneon.cpp
+// Description: matrix-vector product for 8-bit data on neon.
+// Author:      Robin Watts (from the AVX2 original by Ray Smith)
+// Created:     Mon May 12 2020
```
stweil, line 5:
```suggestion
```

amitdo:
>@noahmetzger and @stweil have some code that uses float instead of double for x86.

https://github.com/noahmetzger/tesseract/commits/floatModeRecognition
https://github.com/noahmetzger/tesseract/commits/trainingFloat

https://github.com/stweil/tesseract/commits/dp-float
https://github.com/stweil/tesseract/commits/floatModeRecognition

src/arch/intsimdmatrixneon.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+// Computes part of matrix.vector v = Wu. Computes N=8 results.
+// The weights *must* be arranged so that consecutive reads from wi
+// provides (num_in/kNumInputsPerGroup groups of (N output dim groups of
+// (kNumInputsPerGroup inputs))). After that there must be N consecutive
+// bias weights, before continuing with any more weights.
+// u must be padded out with zeros to
+// kNumInputsPerGroup*ceil(num_in/kNumInputsPerGroup) elements.
+static void PartialMatrixDotVector8(const int8_t* __restrict wi, const double* __restrict scales,
+                                     const int8_t* __restrict u, int num_in, int num_out,
```
stweil, line 58: @robinwatts, you don't use `num_out` here. Can this argument be removed?

src/arch/intsimdmatrixneon.cpp
```diff
@@ -0,0 +1,162 @@
...
+// pair of 4x32 registers, so 8 x 32 bit ints).
+constexpr int kNumOutputsPerRegister = 8;
+// Maximum number of registers that we will use.
+constexpr int kMaxOutputRegisters = 1;
+// Number of inputs in the inputs register.
+constexpr int kNumInputsPerRegister = 8;//32;
+// Number of inputs in each weight group.
+constexpr int kNumInputsPerGroup = 8;
+
+// Functions to compute part of a matrix.vector multiplication. The weights
```
stweil, line 43:
```suggestion
// Function to compute part of a matrix.vector multiplication. The weights
```

src/arch/simddetect.cpp
robinwatts: I don't think I should do that.

None of the other things (avx_available_ etc) are guarded in that way. Also, that would stop IsNEONAvailable() being available in all builds, which would appear to be against the way things are done?

stweil: NEON is only available for ARM, never for Intel, right? So it does not seam useful to add a NEON related variable to Intel code, and it also does not make sense to add AVX or SSE related variables to ARM code.

The way things are done is not constant, but changes when necessary or reasonable.

robinwatts:
I've done all the suggested changes, with the exception of adding the #ifdef ARM_NEON guards, because they are not in keeping with the rest of the code.

If they are really necessary, I'll add them, but it seems slightly odd to treat NEON as different to the x86 SIMD variants.

src/arch/simddetect.cpp
robinwatts: done. thanks.

stweil:
@robinwatts, I tested with commit https://github.com/stweil/tesseract/commit/c5b96260bc025c8e1ffa43ef7e07b427e150d964 on top of your pull request. Would it be okay for you to merge with that commit?

robinwatts:
As I said on the other PR, it wouldn't be the way I'd do it personally, but it's more your code than mine. I'll go along with whatever you think.

stweil:
Unfortunately the new NEON code causes a heap buffer overflow which has to be fixed of course. See PR #2974 for details.

robinwatts:
I have a fix up for review here:

https://github.com/tesseract-ocr/tesseract/pull/2991

configure.ac
```diff
@@ -148,6 +148,12 @@ if $sse41; then
   AC_DEFINE([HAVE_SSE4_1], [1], [Enable SSE 4.1 instructions])
 fi
 
+AX_CHECK_COMPILE_FLAG([-mfpu=neon], [neon=true], [neon=false], [$WERROR])
+AM_CONDITIONAL([HAVE_NEON], $neon)
+if $neon; then
+  AC_DEFINE([HAVE_NEON], [1], [Enable NEON instructions])
```
stweil, line 154: Surprisingly this breaks OSS-Fuzz: https://oss-fuzz-build-logs.storage.googleapis.com/log-2f0c6bde-a12f-49cc-964d-fed11818127e.txt. It uses a build with a clang++ compiler which obviously accepts `-mfpu=neon` on Intel, and the build files because it does not find a header file:

    Step #3: checking whether C++ compiler accepts -mavx... yes
    Step #3: checking whether C++ compiler accepts -mavx2... yes
    Step #3: checking whether C++ compiler accepts -mfma... yes
    Step #3: checking whether C++ compiler accepts -msse4.1... yes
    Step #3: checking whether C++ compiler accepts -mfpu=neon... yes
    [...]
    Step #3: src/arch/simddetect.cpp:46:10: fatal error: 'asm/hwcap.h' file not found

amitdo:
IMO, the term `HAVE_NEON` and the other `HAVE_*` are misleading.

A possible replacement: `COMPILER_SUPPORTS_NEON`.

configure.ac
amitdo: >As a workaround we could skip the check for NEON if SSE or AVX is supported.

At configure time or at run-time?

In general, we should support cross-compiling with either gcc or clang.

robinwatts: AIUI, the configure check checks to see whether the compiler being used for compiling supports the flag. So any ARM targetted compiler (whether a cross-compiler or a native one) should pass the check.

I think I need another configure check in there for whether that header is available. If the header isn't available, I'll disable neon.

I'll have a crack at a PR for that tomorrow.

stweil: @amitdo, it's a check in `configure`, so it is a configure time check. :-)

The problem here is that a compilation for Intel architecture uses a compiler which claims to support NEON. I hope that there are no ARM compilers which support AVX. 

@robinwatts, checking for the header is not a save indicator for ARM, because the same header is also available for MIPS and RISCV64. `configure` also has so many checks that it takes a significant time to run. That's why I prefer a solution which only executes the relevant checks: no AVX ... checks for non-Intel, no NEON check for non-ARM.

stweil:
The `HAVE_` prefix is quite common for autoconf macros, therefore I'd stick to that. But maybe we could use defines like `HAVE_CFLAG_NEON` or `HAVE_NEON_CODE`.

amitdo:
If we still use `HAVE_`, I prefer the `HAVE_CFLAG_NEON` option (or `HAVE_NEON_CFLAG`).

stweil:
Both variants are fine for me. @egorpugin, @zdenop, @Shreeshrii, any preferences?

We would update the other similar macros (HAVE_AVX, ...) then, too.

egorpugin:
I'm fine for any consistent variant.

robinwatts:
How about we:

 a) Fix the current NEON code. (See https://github.com/tesseract-ocr/tesseract/pull/2991)
 b) Fix the neon runtime detection mechanism so oss-fuzz works again. (See https://github.com/tesseract-ocr/tesseract/pull/2992)
 c) Then do a followup commit to change all the flags as people like them?

amitdo:
robinwatts, agreed.

robinwatts:
After we get that lot in, plus the AVX2 acceleration (See #2974) I have a series of changes to accelerate all the different intsimdmatrix implementations by allowing the biasing/scaling to be done in SIMD too.

zdenop:
I have no preference.

PR #2979, opened 2020-05-13
Trivial code documentation fix: move comment about swap meaning for DeSerialize to correct function

juliangilbey:
The patch (which doesn't alter any code) is hopefully self-explanatory!  It had me a little confused for a bit.


stweil:
Thank you.

PR #2981, opened 2020-05-14
Fix OEM_DEFAULT in DISABLED_LEGACY_ENGINE builds.

robinwatts:
If api->Init is called with OEM_DEFAULT in DISABLED_LEGACY_ENGINE
build modes, the engine mode is never set, resulting in no
words being found.

### DIFF
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -100,7 +100,9 @@ bool Tesseract::init_tesseract_lang_data(
         " to your \"tessdata\" directory.\n");
     return false;
   }
-#ifndef DISABLED_LEGACY_ENGINE
+#ifdef DISABLED_LEGACY_ENGINE
+  tessedit_ocr_engine_mode.set_value(OEM_LSTM_ONLY);
+#else
   if (oem == OEM_DEFAULT) {
     // Set the engine mode from availability, which can then be overridden by
     // the config file when we read it below.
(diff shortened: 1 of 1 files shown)


amitdo:
I don't like it. Sorry.

instead of your change, IMO you should remove all values from the `OcrEngineMode` enum exept one - `OEM_LSTM_ONLY`.

amitdo:
An alternative:

```
#ifdef DISABLED_LEGACY_ENGINE
   tessedit_ocr_engine_mode.set_value(OEM_LSTM_ONLY);
#else
  if (oem == OEM_DEFAULT) {
  ...
#endif
```

amitdo:
Another alternative is to remove all values from the OcrEngineMode enum exept one - OEM_LSTM_ONLY.

robinwatts:
I am certainly happy to change the patch to your first suggestion.

The second one seems less ideal to me. a) I fear it'd have much larger knock on changes throughout the code, and b) it'd change the public API of the code. I thought the point of OEM_DEFAULT was that you could say "use whatever method is default for the installed library" - you'd be losing that, as the caller would have to know in advance what methods were available.

Let me know what method you'd prefer and I'll make a patch.

amitdo:
You can go for the first option.

robinwatts:
Done. Thanks, Amit.

robinwatts:
I've added another #ifdef/#endif pair to avoid *non* OEM_DEFAULT settings overwriting the OEM_LSTM_ONLY one that we want.

The build failure is nothing to do with me, I think...

amitdo:
Thanks!

PR #2982, opened 2020-05-15
Guard #include "config_auto.h" with HAVE_CONFIG_H.

robinwatts:
Every other file already does this.

### DIFF
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -15,7 +15,9 @@
 // limitations under the License.
 ///////////////////////////////////////////////////////////////////////
 
+#ifdef HAVE_CONFIG_H
 #include "config_auto.h"     // for HAVE_AVX, ...
+#endif
 #include <numeric>           // for std::inner_product
 #include "simddetect.h"
 #include "dotproduct.h"


stweil:
I am afraid that the guards are there for historic reasons. Builds without config_auto.h are not reasonable because important compile time information would be missing. Therefore I suggest to close this pull request.

Instead we could consider removing all guards.

zdenop:
`config_auto.h` is autotools way how to handle configuration. 
Maybe I miss something, but:
1. This patch does not hurt anything and give more freedom how to handle configuration
2. Removing guards just push other building system to create this file. I do not see any benefit from this step.

robinwatts:
I don't understand the objection here.

Every .cpp file in tesseract except this one properly uses #ifdef HAVE_CONFIG_H. This file is the exception. I am just proposing we bring it into line.

It makes no difference to autotools builds (indeed, with this patch applied, we'd be doing what we're supposed to do).

It just gives the freedom to people to NOT use autotools if they so desire.

Yes, you *could* remove all the guards from the files, but a) this would make you a non-standard autotools build, and b) would screw other users of your library who don't/can't use autotools. All of that for no benefit other than being able to take a dogmatic "everyone must use autotools!" position.

stweil:
> config_auto.h is autotools way how to handle configuration.

That's right, but not the whole story for Tesseract. We currently also create `config_auto.h` for cmake builds. Without putting the required defines into a file any build system would have to pass them as compiler flags. Personally I think that a file is more elegant.

To make thinks easier for non autotool build systems, we could add a template for `config_auto.h` with suggested defines.

robinwatts:
How exactly does a config_auto.h template help people that would rather do it by passing defines in? Certainly, it wouldn't help me.

To turn this around, what exactly is the downside of this patch?

All it does is fix a tiny aberrance in a single file. A single file that isn't following the standard mechanism that all the other files already follow.

The fact that you also create a config_auto.h file for CMake builds is neither here nor there. By simply adding in the missing guards, you give people the freedom to work any way they want, with no disadvantages to anyone.

stweil:
I want to simplify things, remove unneeded lines of code, reduce conditional compilation.

You are free to pass compiler flags, just have to know what you are doing. Things already work for your case - all you have to do is run `touch include/tesseract/config_auto.h` or `echo "/* dummy */" > include/tesseract/config_auto.h` in your build.

robinwatts:
You are right. I can do that, and I will if I have to. (Though, FWIW, portably "echoing" to a file from a makefile is not completely trivial if you have to allow for windows builds too).

It seems to me that what you propose won't be meaningfully simpler, will be more non-standard and certainly will be more awkward in some cases. But I've made my case, and if you still disagree, so be it.

zdenop:
I approve this PR: it is in line how we handle configuration at the moment.

stweil:
So let us get rid of unnecessary conditional compilations later. :-)

zdenop:
Maybe I would suggest to check autotools checks at first stage, because they are replicated in other building tools.

E.g. there is check for presence of ["limits.h" header file](https://github.com/tesseract-ocr/tesseract/blob/5d9b181d67c1da9a317abf0d63a8f0fd72905dc9/configure.ac#L415). If it is presents, it setup variable `HAVE_LIMITS_H` (to `config_auto.h` ;-) ) which according `git grep HAVE_LIMITS_H` is never used. Further interesting fact is that limit.h is used only on one place:

https://github.com/tesseract-ocr/tesseract/blob/5d9b181d67c1da9a317abf0d63a8f0fd72905dc9/include/tesseract/platform.h#L23-L27

According comments it should be used only when it used by c compiler (maybe in tesseract C wrappers?), but tesseract requires c++ compiler with minimum c++11 support, so I can not find reason for that checks for tesseract build... 
I guess most of "historic" part of configure.ac should be reviewed.

PR #2987, opened 2020-05-20
update

leeamen:


zdenop:
We do not accept this PR:  Training tools has be be ON by default, so they can be checked by travis and appveyor. If you do not need it, you can turn it OFF for your build.

PR #2992, opened 2020-05-26
Add configure based detection for NEON runtime detection.

robinwatts:
Previously, we assumed that if the compiler accepted
-mfpu=neon that we'd either be on ANDROID, or asm/hwcap.h
would be available to us.

There will be non-Android systems where asm/hwcap.h is not
available (indeed, the OS-FUZZ target seems to be one).

So, extend the checks to see whether asm/hwcap.h etc is
available. If it is not, then we don't attempt to use it,
and instead look at NEON_ALWAYS_AVAILABLE to decide whether
to hardwire support on/off.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -188,6 +188,9 @@ libtesseract_neon_la_CXXFLAGS = -mfpu=neon
 libtesseract_neon_la_SOURCES = src/arch/intsimdmatrixneon.cpp
 libtesseract_la_LIBADD += libtesseract_neon.la
 noinst_LTLIBRARIES += libtesseract_neon.la
+if HAVE_HWCAP_BASED_NEON_RUNTIME_DETECTION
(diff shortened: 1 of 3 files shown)


stweil:
See pull request #2993 for an alternative solution.

robinwatts:
I use configure to detect the presence of the required header and set a flag. Yes, that flag may be triggered on other architectures than ARM, but it doesn't matter. We only act on that flag in the HAVE_NEON case. Thus it's very simple and does exactly what we need.

IMHO, testing for the host cpu and doing different things based upon that is the wrong thing to do. If a compiler says it supports a given flag then that should be good enough for us. Second guessing the compiler is not going to lead us anywhere good.

But, as always, it's more your code than mine, and I'm not actually using the automake stuff, so... :)

stweil:
The problem here is that we know from OSS Fuzz that there exist compilers which claim to support NEON although not compiling for ARM. So there might also be compilers for PowerPC which do the same, find the header files and fail later during the build.

I close this pull request and take my own one. Thank you for your effort nevertheless.

PR #2993, opened 2020-05-26
Check compiler options depending on host cpu

stweil:


stweil:
This should fix the build for OSS-Fuzz.

PR #3022, opened 2020-06-16
Fix undefined shift with negative value (oss-fuzz issue 14658)

stweil:
This fixes a bug reported by OSS Fuzz:
https://oss-fuzz.com/issue/5697280134348800

The old code passed a negative value (-1) as argument to step_dir
when destindex was 0.


zdenop:
thanks

PR #3023, opened 2020-06-16
Make several functions inline for improved performance

stweil:
OSS Fuzz reports a timeout in issue 19900
(https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=19900).

Im my test, making some functions which are called most often in that test inline
reduces the execution time from 44 s to 39 s.


zdenop:
thanks

PR #3026, opened 2020-06-17
Fix CMake build with DISABLED_LEGACY_ENGINE

knopp:
With this applied I am able to build tesseract using
`cmake -DGRAPHICS_DISABLED=ON -DDISABLED_LEGACY_ENGINE=ON`


stweil:
Thank you!

PR #3031, opened 2020-06-20
Remove unused checks and code from build files

stweil:


amitdo:
What about `AC_HEADER_TIME`?
https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Particular-Headers.html#AC_005fHEADER_005fTIME

stweil:
Thanks. See commits 27b254476f5dce43117688e3c00058387fb6f18f (AC_HEADER_TIME) and bf39019d95d60d6d6c948fd56685f390be337c04 (AC_HEADER_STDC).

PR #3034, opened 2020-06-21
ScrollView: Remove C API callcpp.{cpp,h}

stweil:
Use C++ class ScrollView directly instead of using an intermediate C API.


stweil:
Build fails because pull request #3033 has to be applied first.

Issue #444, opened 2016-10-20
Buffer overflow in proto_evidence_


vidiecan:
It seems that the corner case described below is not handled at all; therefore, I kindly ask someone familiar with the oldest tesseract code (@theraysmith) to give his opinion about it.

We encountered the following issue that results in tesseract crashing (OS and compiler dependent) because of a buffer overflow (input picture cannot be shared). 

TL;DR
Overflow occurs here in intmatcher.cpp at:

```
UINT8Pointer =
            &(tables->proto_evidence_[ActualProtoNum + proto_offset][0]);
          for (ProtoIndex =
            ClassTemplate->ProtoLengths[ActualProtoNum + proto_offset];
          ProtoIndex > 0; ProtoIndex--, UINT8Pointer++) {
            if (Evidence > *UINT8Pointer) {
              Temp = *UINT8Pointer;
              *UINT8Pointer = Evidence;  <<---- **HERE**
              Evidence = Temp;
            }
            else if (Evidence == 0)
              break;
          }
```

 proto_evidence_ is initialised like this 

```
uinT8 proto_evidence_[MAX_NUM_PROTOS][MAX_PROTO_INDEX];
```

where

```
#define MAX_NUM_PROTOS    512
#define MAX_PROTO_INDEX   24
```

The problem is that ProtoLengths[511] has size of > 24 and (un)luckily the Evidence condition holds so an invalid write is performed (that in our case overwrote heap info that somewhere later in a delete caused the problem...)
## Details

The stacktrace is classify_word_pass1->LearnWord->LearnPieces->AdaptToChar->MakeNewTemporaryConfig
where `NumBadFeatures` is computed (im_.FindBadFeatures) and gets exactly 100 (also NumFeatures=100). Then `MakeNewTempProtos` is called where `SegmentLength` is computed as `2.2999` and in `ConvertProto` we have

```
Param = (Proto->Length / GetPicoFeatureLength()) + 0.5;
Class->ProtoLengths[ProtoId] = TruncateParam(Param, 1, 255, NULL);
```

where 

```
Param = 46
ProtoId = 511
```

with these values, the `UINT8Pointer` from above makes first 24 valid read/writes but then goes beyond  `proto_evidence_`.

vidiecan:
for the record, one hack/fix that solves it is `tesseract/classify/intproto.cpp`
```
-  Class->ProtoLengths[ProtoId] = TruncateParam(Param, 1, 255, NULL);
+
[...]

PR #3035, opened 2020-06-22
Avoid buffer overflow (issue #444)

stweil:


stweil:
See issue #444.

nironater:
@stweil, That's great news, good job!
I understand  that this fix will be released with v5.0.0?
Can you say when it's scheduled to be released?
Won't hold you to it don't worry :)

stweil:
There is currently no plan when v5.0.0 will be released, and I don't expect it will be this year.

PR #3041, opened 2020-06-25
Add MT option for MSVC compilation

prabindh:
Adding MT option for MSVC compiler option using CMAKE_MSVC_RUNTIME_LIBRARY option.

Applicable only to WIN32


zdenop:
thanks

Issue #3044, opened 2020-06-26
5.0 crashes on SSE architecture for builds with MS Visual Studio


tdhintz:
### Environment

* **Tesseract Version**: tesseract.exe --version
tesseract 5.0.0-alpha-704-g38046
leptonica-1.80.0 (Jun 17 2020, 02:02:24) [MSC v.1926 LIB Debug x64]
  libgif 5.2.1 : libjpeg 9d : libpng 1.6.37 : libtiff 4.1.0 : zlib 1.2.11 : libwebp 1.0.3 : libopenjp2 2.3.1
Found SSE

* **Commit Number**: June 17th
* **Platform**: Windows Server 2019 Standard virtual Intel Xeon X5650 2.67GHz (2 CPUs) cpuinfo[2]: 82B82203, Visual Studio 2019 Win64, Microsoft Visual C++ 2015-2019 Redistributable (x64).  Processor has AVX2 but it is disabled in the virtual environment.

### Current Behavior:
Crash occurs in the routine, IntDotProductSSE, where it first invokes _mm_loadl_epi64.  I have indicated the line below.
```
// Computes and returns the dot product of the n-vectors u and v.
// Uses Intel SSE intrinsics to access the SIMD instruction set.
static int32_t IntDotProductSSE(const int8_t* u, const int8_t* v, int n) {
  int max_offset = n - 8;
  int offset = 0;
  // Accumulate a set of 4 32-bit sums in sum, by loading 8 pairs of 8-bit
  // values, extending to 16 bit, multiplying to make 32 bit results.
  int32_t result = 0;
  if (offset <= max_offset) {
    offset = 8;
    __m128i packed1 = _mm_loadl_epi64(reinterpret_cast<const __m128i*>(u));  // Crash here

```
Tesseract.exe produces no output.  Windows event log shows a crash.  Using generic (disable SSE) works but is slow. 

Windows event log entries are attached.
[Error2.txt](https://github.com/tesseract-ocr/tesseract/files/4839306/Error2.txt)
[Error1.txt](https://github.com/tesseract-ocr/tesseract/files/4839307/Error1.txt)

Using libraries from Charlesw C# Tesseract wrapper results in an SEHException.

Works fine on a machine with cpuinfo[2] = 7FFAFBFF regardless whether Tesseract is asked to use only SSE or not.  Version report from working machine:

tesseract.exe --version
tesseract 5.0.0-alpha-692-g62ea
[...]

PR #3050, opened 2020-06-30
Cmake avx

zdenop:


CMakeLists.txt
```diff
@@ -154,30 +157,101 @@ else()
...
+CHECK_CXX_COMPILER_FLAG("-mavx" HAVE_AVX)
+if(HAVE_AVX)
+    set(AVX_COMPILE_FLAGS "-mavx")
+    add_definitions("-DHAVE_AVX")
+endif(HAVE_AVX)
+
+CHECK_CXX_COMPILER_FLAG("-mavx2" HAVE_AVX2)
+if(HAVE_AVX2)
+    set(AVX2_COMPILE_FLAGS "-mavx2")
+    add_definitions("-DHAVE_AVX")
```
stweil, line 175:
```suggestion
    add_definitions("-DHAVE_AVX2")
```

CMakeLists.txt
```diff
@@ -154,30 +157,101 @@ else()
...
+    add_definitions("-DHAVE_FMA")
+endif()
+
+CHECK_CXX_COMPILER_FLAG("-msse4.1" HAVE_SSE4_1)
+if(HAVE_SSE4_1)
+    set(SSE4_1_COMPILE_FLAGS "-msse4.1")
+    add_definitions("-DHAVE_SSE4_1")
 endif()
 
+if(NOT APPLE)
```
stweil, line 190: NEON is only available for ARM. As Apple is going to replace Intel by ARM, this would be wrong soon. We should check for ARM instead of checking for `NOT APPLE`.

stweil: Likewise the checks for Intel features (AVX, ...) could be run only for Intel architecture.

zdenop: to make it short: see [log from travis](https://travis-ci.org/github/tesseract-ocr/tesseract/jobs/703646220) - neon is tested positively in apple build, but our code for neon is not expecting it.
IMO iphones and ipad are already ARM devices, so we can remove this restriction when tesseract support for neon will be adjusted to this.

zdenop:
These changes should fix [travis builds](https://travis-ci.org/github/tesseract-ocr/tesseract/builds/703646218) and maybe fix issues on windows like #3044

PR #3052, opened 2020-07-01
Fix cmake build for MSVC

stweil:
MSVC does not support /arch:FMA or /arch:SSE4.1.
For /arch:AVX and /arch:AVX2 no check is needed because they are supported since a long time.


amitdo:
Just a note:
>AVX2 extends most integer operations to 256-bit vectors, and **enables use of Fused Multiply-Add (FMA) instructions**.

@stweil's patch already take this into account.

zdenop:
just quick note: this should work for VS2017 and higher with 64bit build (which should automatically use SSE). I am not sure about 32bit build and older VS.

PR #3055, opened 2020-07-07
Use const char* for filename parameters

stweil:
This replaces the proprietary STRING data type
(801 instead of 838 lines remaining).

It also removes STRING from osdetect.h and serialis.h.


zdenop:
thanks!

PR #3061, opened 2020-07-10
Always use NEON by default for ARMv8

stweil:
Signed-off-by: Stefan Weil <stefan.weil@bib.uni-mannheim.de>


stweil:
This reduces the time for `tesseract test/testing/8087_054.3G.tif -` from 18.3 s to 10.9 s on NVIDIA Jetson AGX.

zdenop:
thanks

PR #3064, opened 2020-07-15
Fix Memory leak when using TESSERACT_IMAGEDATA_AS_PIX

robinwatts:
If building with TESSERACT_IMAGEDATA_AS_PIX, then tesseract
doesn't compress/decompress images, but rather holds the
data as internal Pix structures. Unfortunately, I forgot to
make the ImageData destructor free these, so memory leaked
during use. Fixed here.


zdenop:
thanks

PR #3065, opened 2020-07-16
Squash some warnings in MSVC build.

robinwatts:
In particular, "defined but not used" (caused by GRAPHICS_DISABLED),
double constants being truncated to floats, and implicit casts.

### DIFF
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -74,9 +74,6 @@ void Tesseract::recog_pseudo_word(PAGE_RES* page_res,
  * @param pr_it the page results iterator
  */
 bool Tesseract::recog_interactive(PAGE_RES_IT* pr_it) {
-  int16_t char_qual;
-  int16_t good_char_qual;
-
   WordData word_data(*pr_it);
   SetupWordPassN(2, &word_data);
   // LSTM doesn't run on pass2, but we want to run pass2 for tesseract.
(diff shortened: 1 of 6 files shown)


zdenop:
thanks

stweil:
This PR contains several type casts of old type which either should be replaced by C++ type casts or be avoided by other means.

zdenop:
@stweil: I fixed this, but AFAIK there are more code to be changes...

robinwatts:
I build Tesseract/Leptonica as part of MuPDF. So they are minimal builds for just what I need. I don't build the command-line tools etc (except on Linux for testing).

The commits haven't quite made it to our main repo yet (https://git.ghostscript.com/?p=mupdf.git;a=summary), but they can be found on my personal repo, here: https://git.ghostscript.com/?p=user/robin/mupdf.git;a=summary

If you:

  git clone git://git.ghostscript.com/user/robin/mupdf.git --recursive

you should get everything checked out, with thirdparty/leptonica and thirdparty/tesseract being git repos for leptonica and tesseract respectively.

Open platform/win32/mupdf.sln in VS2019, and there should be libtesseract and libleptonica projects. You'll need to use the {Debug,Release,Memento}Tesseract configurations for them to actually be built.

Any problems, just let me know! If you want to talk to me direct, I am generally on the #mupdf irc channel, that you can reach here: https://webchat.freenode.net/?channels=mupdf

PR #3066, opened 2020-07-16
Remove unused char constant that causes a warning.

robinwatts:
The kDictWildcard is never actually used, so removing it makes
no difference. It causes warnings in MSVC builds as MSVC doesn't
know how to pack a unicode value into chars.

### DIFF
--- a/src/dict/dict.h
+++ b/src/dict/dict.h
@@ -60,7 +60,6 @@ using DawgVector = GenericVector<Dawg *>;
 // Constants
 //
 static const int kRatingPad = 4;
-static const char kDictWildcard[] = "\u2606";   // WHITE STAR
 static const int kDictMaxWildcards = 2;  // max wildcards for a word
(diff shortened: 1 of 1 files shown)


zdenop:
It is commenting and not removing...

robinwatts:
Would you rather I removed it? Or just changed the commit message to say "commented"?

stweil:
Please remove that line.

src/dict/dict.h
```diff
@@ -60,7 +60,7 @@ using DawgVector = GenericVector<Dawg *>;
 // Constants
 //
 static const int kRatingPad = 4;
-static const char kDictWildcard[] = "\u2606";   // WHITE STAR
+//static const char kDictWildcard[] = "\u2606";   // WHITE STAR
```
stweil, line 63:
```suggestion
```

robinwatts:
Done (sorry for the delay).

stweil:
Thanks.

PR #3073, opened 2020-08-11
Converted 3 echo commands to one

uddeshyatyagi:


stweil:
I am not sure that this is an improvement. It saves some lines of script code, but does not improve readability. And there might be a risk that not all implementations of `echo` support the `-e` option.

uddeshyatyagi:
The ‘-e‘ option in Linux acts as interpretation of escaped characters that are backslashed like \b,\t,\n,\c,\r etc.
So it saves our time by writing the command in one go,instead of writing echo command again and again.
But its okay if it is not mergable,i will forward to contribute to this project in some other form.

stweil:
>  But its okay if it is not mergable,i will forward to contribute to this project in some other form.

I close this pull request because it might be a risk for the portability if we used `echo -e`.

Thank you nevertheless for your contribution.

PR #3082, opened 2020-08-24
Fix separator line detector

bertsky:
This fixes 2 minor issues with foreground line separator detection.

Here is an example (using the ALTO renderer and PageViewer for display):
- before (misses small vertical and large horizontal line):
fb2_-_007_-_srgb tesseract-orig
- after (fixed):
fb2_-_007_-_srgb tesseract-fixed-lineseg

### DIFF
--- a/src/textord/alignedblob.cpp
+++ b/src/textord/alignedblob.cpp
@@ -53,7 +53,7 @@ const int kMinRaggedTabs = 5;
 // Min number of points to accept for an aligned tab stop.
 const int kMinAlignedTabs = 4;
 // Constant number of pixels minimum height of a vertical line.
-const int kVLineMinLength = 500;
+const int kVLineMinLength = 300;
 // Minimum gradient for a vertical tab vector. Used to prune away junk
 // tab vectors with what would be a ridiculously large skew angle.
 // Value corresponds to tan(90 - max allowed skew angle)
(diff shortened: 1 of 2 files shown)


Shreeshrii:
@bertsky Since you are fixing line detection, can you also look into incorrect sgmentation when the script extends below the baseline. See example:

3_1L_pageview

Zip file with sample images and hocr and alto output enclosed.
[grantha.zip](https://github.com/tesseract-ocr/tesseract/files/5129085/grantha.zip)

bertsky:
@Shreeshrii sorry, but this does _not_ address textline detection directly. It's about the v/h-line (foreground separator) detection that is a preliminary to block/column detection. (I have renamed the title to make this difference more clear.)

But your case looks like merely an artefact of the layout extraction API or ALTO representation: bboxes are too coarse here. We should try to allow retrieving polygons. There was an (slightly off-topic) discussion about this in #2971. No dedicated issue or PR yet.

Does the recognition result also degrade where line bboxes overlap? (This is currently the best indicator of the internal row structure IIUC.)

Shreeshrii:
Yes, the recognition result degrades in such cases, which has been reported earlier in other issues. I checked with Pageviewer today and noticed the overlaps.

Should I open this as a separate issue?

bertsky:
> Should I open this as a separate issue?

Please do!

stweil:
Thank you, @bertsky.

stweil:
These commits break the `resultiterator_test`. I think that test requires an update. See issue #3122.

bertsky:
> These commits break the `resultiterator_test`. I think that test requires an update. See issue #3122.

As I have said _somewhere_ (cannot find it) already, sorry and I'll have a look at that unit test and make another PR – as soon as I found some time.

PR #3090, opened 2020-09-06
Correct "NoImages" in debug pdf file

nam-leduc:
- Issues:
  - Debug information for "NoImages" just be binary image,
  it don't show up the result of photo_mask_pix to developer

- Fix:
  - Substract binary image to photo_mask_pix, the result
  are "NoImages" binary pix


zdenop:
thanks

PR #3112, opened 2020-10-01
Fix CMake install configs

Shatur95:
Currently the CMake files are being installed in the `cmake` folder in the root of the install prefix. This is a bit odd. On Linux installation through CMake will create `cmake` in the `/usr` (common installation prefix on most distros), which is not desirable.

Usually CMake files are installed to the `lib/cmake` folder and export files are installed to the `lib/library-name`. That's what I changed in this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -372,10 +372,10 @@ configure_file(
     ${CMAKE_CURRENT_BINARY_DIR}/vs2010/tesseract/libtesseract.rc @ONLY)
 configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/templates/TesseractConfig-version.cmake.in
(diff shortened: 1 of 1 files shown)


Shatur95:
[CMake Search Procedure](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure).

egorpugin:
Thank you!

yuriisk:
The path should be updated also in `cmake/templates/TesseractConfig.cmake.in`.

Shatur95:
@yuriisk, sorry, don't get it. Could you explain where in this file the path should be updated?

yuriisk:
This command `include(${CMAKE_CURRENT_LIST_DIR}/TesseractTargets.cmake)` can't find `TesseractTargets.cmake` because it isn't located in the same directory with `TesseractConfig.cmake` anymore.

Shatur95:
Thanks, will send a PR.

PR #3116, opened 2020-10-05
Use DESTINATION instead of TYPE

Shatur95:
For compatibility with older CMake, references #3112, fixes [this](https://travis-ci.org/github/tesseract-ocr/tesseract/jobs/732859406) CMake issue on Travis.


zdenop:
thanks

PR #3121, opened 2020-10-12
update the sw.yml

Anshulverma-prg:
change the name and change the os

### DIFF
--- a/.github/workflows/sw.yml
+++ b/.github/workflows/sw.yml
@@ -1,4 +1,4 @@
-name: sw
+name: as
 
 on: [push, pull_request]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/sw.yml
```diff
@@ -1,4 +1,4 @@
-name: sw
+name: as
```
stweil, line 1: What is the reason for this change?

Anshulverma-prg:
Because it is bulid the new program name and it will be necessary for the
program.

zdenop:
@egorpugin : any idea about this?

egorpugin:
Maybe @Anshulverma-prg tried to do changes only in his fork?
1. Name is fine.
2. We'll use Ubuntu LTS, no need to update.

PR #3123, opened 2020-10-13
Summation of matrix diagonally

taukirzcoder:
first add the right diagonal's element and then left diagonal's element


zdenop:
this is not tesseract relevant

PR #3127, opened 2020-10-17
backup the file instead removing them

dr298:
sometime you need the old file, but accidentally deleting them.
so, here the way out. simply move them to backup folder, and after you sure that u dont need that file anymore, just do rm -rf ~/backup

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -37,12 +37,14 @@
 #
 # All the rest is auto-generated.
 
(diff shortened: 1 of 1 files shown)


zdenop:
You did not explain why EVERYBODY has to backup generated  files (from configuration/system). 
I do not see any reason for this PR.

stweil:
@dr298, you can get those files again by running `./autogen.sh`, so there is no need for a backup.

PR #3128, opened 2020-10-17
Create .gitattributes for cross os contributors

acoder77:
With this set, Windows users will have text files converted from Windows style line endings (\r\n) to Unix style line endings (\n) when they’re added to the repository.
https://www.edwardthomson.com/blog/git_for_windows_line_endings.html

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto


amitdo:
Here it recommended not to use `* text=auto`:
 
https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file

@zdenop, should we use the suggested alternative, which is to explicitly list text files' extensions?

zdenop:
thanks

PR #3129, opened 2020-10-18
Update INSTALL

davreign:


stweil (commented):
You did not explain why those modifications are desired. Or is it just to get a hacktoberfest t-shirt?

PR #3130, opened 2020-10-19
Tweak SIMDDetect for ANDROID Neon.

robinwatts:
cpufeatures.h should be cpu-features.h, with the latest NDK
at least. The #if 0'd section is not required because armv8
always includes NEON.

### DIFF
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -39,7 +39,7 @@
 
 #if defined(HAVE_NEON) && !defined(__aarch64__)
 #ifdef ANDROID
-#include <cpufeatures.h>
+#include <cpu-features.h>
 #else
 /* Assume linux */
 #include <sys/auxv.h>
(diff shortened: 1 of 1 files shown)


zdenop:
Is the name of header file NDK version depend?

robinwatts:
I *believe* it was a typo from when I adapted the neon detection stuff from an assortment of blog posts. Certainly, in the reports I've been getting back from the Android devs here, it should have been cpu-features.h.

zdenop:
thanks

robinwatts:
Thanks for the review! That's all my changes into master now.

amitdo:
For some reason (speed?), Ray disabled the lstm engine in Android, so this code is unreachable.

robinwatts:
> For some reason (speed?), Ray disabled the lstm engine in Android, so this code is unreachable.

Really? <some feverish greping later> Ah. He undefines areas of code based upon ANDROID_BUILD. I don't set that in the build system I'm using, so it works just fine for me :)

amitdo:
Yes, you're right. It seem;s that our build tools does not disable lstm on Android. 

Still, I wonder about this `ANDROID_BUILD`.

stweil:
> Still, I wonder about this ANDROID_BUILD.

Maybe it was part of internal builds at Google? Should we remove it? It seems to be confusing.

robinwatts:
+1 for removal, for what that's worth.

zdenop:
IMO most prominent usage of tesseract on android is [textfairy](https://github.com/renard314/textfairy). So if would be nice to hear from @renard314  what is needed for android...

robinwatts:
Having a #define DISABLE_LSTM_ENGINE or similar would make sense to me. Far nicer than having to know that ANDROID_BUILD actually means "disable the lstm engine". #defines should be named for what they do. IMHO etc.

renard314:
For Textfairy,  I'm building tesseract using custom make files for android and I'm only using the LSTM mode. So ANDROID_BUILD is not set by my make files or the NDK.

Issue #2879, opened 2020-02-06
Invisible glyph bounds at wrong positions in PDF


THausherr:
### Environment

* **Tesseract Version**: Tesseract Open Source OCR Engine v5.0.0.20190623 with Leptonica
(downloaded from https://digi.bib.uni-mannheim.de/tesseract/ )
* **Platform**: W7 64 bit 6.1.7601

### Call:

"C:\Program Files\Tesseract-OCR\tesseract" scan.tif scan-ocr pdf

### Current Behavior:

text bounds are not identical to visible glyphs in Adobe Reader. Example:

![grafik](https://user-images.githubusercontent.com/6665575/73934890-a61c4980-48df-11ea-8eae-351a695d1c9f.png)

### Expected Behavior:

text bounds should be identical to visible glyphs in Adobe Reader. In the graphic, the blue color should cover the "n".

### Suggested Fix:

I suspect that the /W array is missing in the font dictionary:
![grafik](https://user-images.githubusercontent.com/6665575/73935323-905b5400-48e0-11ea-8d7f-b0237dbd04aa.png)
So Adobe will use the /DW 500 entry (screenshot from PDF 32000 specification):
![grafik](https://user-images.githubusercontent.com/6665575/73935364-a406ba80-48e0-11ea-9b76-f5481696c09c.png)

[scan-ocr.pdf](https://github.com/tesseract-ocr/tesseract/files/4164943/scan-ocr.pdf)
[scan.tif.zip](https://github.com/tesseract-ocr/tesseract/files/4164945/scan.tif.zip)

jbreiden:
Interesting suggestion. If correct, why would it show up as an n - 1
problem in highlighting?

THausherr:
Sorry, I don't understand what you mean. My argument is that the highlights widths don't match. Adobe gets these from the font data, and widths are different in a proportional font. And it isn't just the "n". When trying to highlight the "I" it looks like this:
![grafik](https://user-images.githubusercontent.com/6665575/73942244-01a20380-48ef-11ea-9db4-c01798a3900a.png)

jbreiden:
The glyphless font deliberately uses equal width for every character. I
stretch the the word using Tz in the PDF to make it fit. So I expect word
highlighting to look correct, but not character highlighting within a word.
This design was chosen to maximize compatibility across all the scripts
supported by Tesseract while minimizing complexity.

THausherr:
[...]

PR #3139, opened 2020-10-23
Fix word box boundaries in rendered PDF

sbjorn:
This addresses the dimension of the text boxes in the generated PDFs (essentially a one-off bug) that are slightly too small to accommodate the boundary of the words. Here is an example of some text selected in a sandwiched PDF:

pre_000

Sandwiched document regenerated after fix:

post_000

### DIFF
--- a/src/api/pdfrenderer.cpp
+++ b/src/api/pdfrenderer.cpp
@@ -468,7 +468,9 @@ char* TessPDFRenderer::GetPDFTextObjects(TessBaseAPI* api,
     } while (!res_it->Empty(RIL_BLOCK) && !res_it->IsAtBeginningOf(RIL_WORD));
     if (res_it->IsAtBeginningOf(RIL_WORD)) {
       pdf_word += "0020";
-      pdf_word_len++;
+      // We don't increment `pdf_word_len` here because it repesents the number
+      // of characters of the word itself - the added space is not part of the
+      // word!
     }
     if (word_length > 0 && pdf_word_len > 0) {
       double h_stretch =


zdenop:
Please provide testing case for problem and other details (e.g, which pdf viewer you use...)

stweil:
I tried the patch on `test/testing/8087_054.3G.tif`. In Firefox the result looks pretty good. With macOS preview it has several problems: For some words the new boxes look better, but many words are no longer separate, so it is for example no longer possible to select a word by double clicking. Some lines even change the height (they are now too high). Therefore I think this needs more changes to be really good.

[PDF without patch](https://github.com/tesseract-ocr/tesseract/files/5433660/old.pdf)
[PDF with patch](https://github.com/tesseract-ocr/tesseract/files/5433661/new.pdf)

amitdo:
See #2879.

sbjorn:
@stweil thanks for submitting an example. I have tested exclusively under Linux, and only seen improved results with the following PDF viewers:

- evince version 3.34.2 (poppler)
- chromium version 85 (pdfium)

I just got hold of a windows machine to test with Adobe DC Reader DC 2020.012.20048 (English) with improved results as well (the words are completely covered on selection). No problems selecting words by double-clicking during any of my tests. The height of the selected text stays the same from what I can see by visual inspection. I find it strange that the macOS viewer shows a different height of the text box, as this code change only tweaks the horizontal stretch. Do you see this behavior in all readers on macOS? (I don't have a mac myself, so I can't check).

MerlijnWajer:
(I made a Python port of pdfrenderer.cpp that reads hOCR files rather than Tesseract results, but am seeing the same problem. Code: https://git.archive.org/merlijn/archive-pdf-tools/-/blob/master/pdfrenderer.py)

This image shows the 'original' code, with the `pdf_word_len++` increase in place. The center PDF is generated with `pdf_word_len++` removed. The right PDF is just provided for reference, since the bounding boxes make much of the PDF not visible.

In the both left and center PDF, I pressed Ctrl+A in evince to 'highlight' all text. The reason that you actually see glyphs is because I inserted a second font that references a system font to show glyphs, which doesn't break text copy / paste. (I plan to submit a PR for this later this month)

pdfspaces


While the new code does seem to make some bounding boxes better, it also looks like the titles overlap more (but that might be an artifact from my port, or just from the hOCR file(s))

amitdo:
MerlijnWajer,

First, you pasted an incorrect link. Here is the script location:  
https://git.archive.org/merlijn/archive-pdf-tools/-/blob/show-text-on-selection/internetarchivepdf/pdfrenderer.py

I don't see a license informantion in your repo. 

> I inserted a second font that references a system font to show glyphs, which doesn't break text copy / paste. (I plan to submit a PR for this later this month)

See this comment: https://github.com/tesseract-ocr/tesseract/issues/2879#issuecomment-583897489

You should try to test with [this patch](https://github.com/tesseract-ocr/tesseract/issues/2879#issuecomment-583896719), which was provided by the person that wrote the pdf output code in Tesseract.

MerlijnWajer:
> @MerlijnWajer,
> 
> First, you pasted an incorrect link. Here is the script location:
> https://git.archive.org/merlijn/archive-pdf-tools/-/blob/show-text-on-selection/internetarchivepdf/pdfrenderer.py
> 
> I don't see a license informantion in your repo.

Don't want to derail, but just wanted to say that the license will be the same as the pdfrenderer.cpp in Tesseract, of course. There is a comment in the file that already hints at that. I'm planning to finish that this week - sorry for the noise.

> > I inserted a second font that references a system font to show glyphs, which doesn't break text copy / paste. (I plan to submit a PR for this later this month)
> 
> See this comment: [#2879 (comment)](https://github.com/tesseract-ocr/tesseract/issues/2879#issuecomment-583897489)

I will open a separate issue for this later, not to derail discussion here.

> You should try to test with [this patch](https://github.com/tesseract-ocr/tesseract/issues/2879#issuecomment-583896719), which was provided by the person that wrote the pdf output code in Tesseract.

Ok, thanks, I will test that this week and let you know if it improves the situation.

amitdo:
I think @jbreiden's patch should be used instead of this PR.

https://github.com/tesseract-ocr/tesseract/issues/2879#issuecomment-583896719

amitdo:
Actually, this patch has a good impact on Firefox's pdf.js, unlike the patch from https://github.com/tesseract-ocr/tesseract/issues/2879#issuecomment-583896719.

Both have a bad impact on macOS's Preview.

MerlijnWajer:
I ran into this the other day: https://github.com/nickjwhite/gofpdf/commit/3e3f1fbc561c4ae28e33d92959d58f2aa74428c8

mentioned in https://blog.rescribe.xyz/posts/pdfs/ - maybe it has something useful there.

PR #3141, opened 2020-10-26
Modify OCR for inverted text

stweil:
The old code tries a 2nd OCR on lines when a word confidence is below 50 %, so a single word with low confidence triggers it. This happens rather often, mostly for lines without inverted text, and costs performance.

The new code checks the mean confidence instead of the minimum. This typically improves the performance significantly and still works for lines with much inverted text.

In addition, the OCR result of the inverted image is now accepted if its mean confidence is better than the original one. In my test this improves the OCR result, especially for lines which only have some words with inverted text.

The algorithm still does not handle lines with both normal and inverted text optimally. OCR for such lines could be improved by inverting word wise or maybe by using Leptonica function `pixAutoPhotoinvert`.

### DIFF
--- a/src/lstm/lstmrecognizer.cpp
+++ b/src/lstm/lstmrecognizer.cpp
@@ -302,7 +302,7 @@ bool LSTMRecognizer::RecognizeLine(const ImageData& image_data, bool invert,
                       &inv_outputs);
     float inv_min, inv_mean, inv_sd;
     OutputStats(inv_outputs, &inv_min, &inv_mean, &inv_sd);
-    if (inv_min > pos_min && inv_mean > pos_mean && inv_sd < pos_sd) {
+    if (inv_mean > pos_mean) {
       // Inverted did better. Use inverted data.
       if (debug) {
         tprintf("Inverting image: old min=%g, mean=%g, sd=%g, inv %g,%g,%g\n",
--- a/src/lstm/lstmrecognizer.cpp
+++ b/src/lstm/lstmrecognizer.cpp
@@ -2,7 +2,6 @@
 // File:        lstmrecognizer.cpp
 // Description: Top-level line recognizer class for LSTM-based networks.
 // Author:      Ray Smith
-// Created:     Thu May 02 10:59:06 PST 2013
 //
 // (C) Copyright 2013, Google Inc.
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -291,7 +290,7 @@ bool LSTMRecognizer::RecognizeLine(const ImageData& image_data, bool invert,
   // Check for auto inversion.
   float pos_min, pos_mean, pos_sd;
   OutputStats(*outputs, &pos_min, &pos_mean, &pos_sd);
-  if (invert && pos_min < 0.5) {
+  if (invert && pos_mean < 0.5) {
     // Run again inverted and see if it is any better.
     NetworkIO inv_inputs, inv_outputs;
     inv_inputs.set_int_mode(IsIntMode());


stweil:
The time for OCR of a double page from a historical newspaper was reduced from 185 s to 136 s by these modifications.

zdenop:
thanks.

Shreeshrii:
@zdenop Are you also applying these improvements to the 4.1 branch?

stweil:
> Are you also applying these improvements to the 4.1 branch?

I am not sure whether that would be correct for 4.1, because it is not a bug fix, but a modification which changes OCR results. So it would require at least a new 4.2.

Shreeshrii:
Thanks, @stweil. Yes, 4.2 will be appropriate.

@zdenop You could include other recent improvements also which are suitable for 4.x branches.

stweil:
I got some [test images](https://archive.org/~merlijn/tesseract-images/) from the Internet Archive. The new code gives better results for many lines in those images, but some lines which were inverted with the old code are now no longer inverted and are not recognized.

Changing the threshold from 0.5 to 0.7 still got all improvements, but fixed the lines with regressions. If that is confirmed in more tests, it should be changed in the code. Maybe the threshold should be a parameter.

Old and new results differ here:

    diff --rec -u old/sim_japanese-economy_fall-1989_18_1_0099.txt new/sim_japanese-economy_fall-1989_18_1_0099.txt
    --- old/sim_japanese-economy_fall-1989_18_1_0099.txt	2020-10-27 09:49:56.105843034 +0100
    +++ new/sim_japanese-economy_fall-1989_18_1_0099.txt	2020-10-27 14:35:58.858059925 +0100
    @@ -2,7 +2,7 @@
     agement material from Japanese sources, primarily scholarly journals and
     books. The selections are intended to reflect developments in the Japanese
     economy and to be of interest to those professionally concerned with this
    -i H
    +field.
     
     Editor: Kazuo Sato, Rutgers University
     Assistant Editor: Anita M. O’Brien, M. E. Sharpe, Inc.
    diff --rec -u old/sim_journal-of-burn-care-research_1983-02_4_1_0000.txt new/sim_journal-of-burn-care-research_1983-02_4_1_0000.txt
    --- old/sim_journal-of-burn-care-research_1983-02_4_1_0000.txt	2020-10-27 09:50:07.041917640 +0100
    +++ new/sim_journal-of-burn-care-research_1983-02_4_1_0000.txt	2020-10-27 14:36:01.022075792 +0100
    @@ -10,10 +10,10 @@
     \dult Learning Principles—Basis for Educaling the Burn Nurse
     Opportunities and Expanded Educational Roles in Burn Nursing
     ducational Approaches to Burn Nuising Orientation
    -Development of Competencv-Based STt R IR Tt \m"sin:
    +Development of Competencv-Based Orientation for Burn \m"sin:
     Developing Outreach Programs in Community Hospitals
     
    -The |’|'i|n;u'.\ A TR Svstemn
    +The |’|'i|n;u'.\ Nursing Svstemn
     
     Characteristics of Burn Orientation Programs
     
    diff --rec -u old/sim_journal-of-burn-care-research_1983-02_4_1_0006.txt new/sim_journal-of-burn-care-research_1983-02_4_1_0006.txt
    --- old/sim_journal-of-burn-care-research_1983-02_4_1_0006.txt	2020-10-27 09:50:09.589935013 +0100
    +++ new/sim_journal-of-burn-care-research_1983-02_4_1_0006.txt	2020-10-27 14:36:02.006083007 +0100
    @@ -5,11 +5,11 @@
     
     LR
     
    -bk
    +112
     
     =y
     
      
     
    -£
    +I
    diff --rec -u old/sim_journal-of-burn-care-research_2007-04_28_2_0000.txt new/sim_journal-of-burn-care-research_2007-04_28_2_0000.txt
    --- old/sim_journal-of-burn-care-research_2007-04_28_2_0000.txt	2020-10-27 09:50:39.058135718 +0100
    +++ new/sim_journal-of-burn-care-research_2007-04_28_2_0000.txt	2020-10-27 14:36:06.314114591 +0100
    @@ -10,9 +10,9 @@
     
     2
     e
    -5
    -2
    -O
    +)
    +b
    +(4]
     =z
     
     jiation
    diff --rec -u old/sim_television-broadcast-tvb_2003-04_26_4_0000.txt new/sim_television-broadcast-tvb_2003-04_26_4_0000.txt
    --- old/sim_television-broadcast-tvb_2003-04_26_4_0000.txt	2020-10-27 09:50:44.414172153 +0100
    +++ new/sim_television-broadcast-tvb_2003-04_26_4_0000.txt	2020-10-27 14:36:07.990126878 +0100
    @@ -1,4 +1,4 @@
    -A T D R T OF TELEVlSlON
    +& THE SCIENCE, ART & BUSINESS OF TELEVlSlON
     
     NABZUUS
     
    @@ -11,18 +11,18 @@
     1553
     4 el sl L, 1. n A
     
    -Il Congess ik
    +Will Congess Put The
     
     2
    -- £ 4
    +h £ 4
     L] . i »
    -- { bl
    +. { bl
     ¢ § ias P
     ' i
     A L I | b+ Pe. .:v
     5  je
     v. = !
    -- b 5 52
    +. b : 52
     R 4
     ~ R
     
    diff --rec -u old/sim_television-broadcast-tvb_2003-04_26_4_0003.txt new/sim_television-broadcast-tvb_2003-04_26_4_0003.txt
    --- old/sim_television-broadcast-tvb_2003-04_26_4_0003.txt	2020-10-27 09:50:58.062264937 +0100
    +++ new/sim_television-broadcast-tvb_2003-04_26_4_0003.txt	2020-10-27 14:36:11.038149223 +0100
    @@ -7,7 +7,7 @@
      
     
     | ‘ DM-3000
    -Single or Multi-Channel Composite; Component, TRy a0
    +Single or Multi-Channel Composite; Component, Svnchronous AES/EBU
     SDI Master Control Switcher SDI, ASI, HDTV Routers ar\:gcAr:glnooguZudio/Routers
     
     System Configuration, Operation,
    diff --rec -u old/sim_television-broadcast-tvb_2003-04_26_4_0006.txt new/sim_television-broadcast-tvb_2003-04_26_4_0006.txt
    --- old/sim_television-broadcast-tvb_2003-04_26_4_0006.txt	2020-10-27 09:51:04.978311923 +0100
    +++ new/sim_television-broadcast-tvb_2003-04_26_4_0006.txt	2020-10-27 14:36:12.558160366 +0100
    @@ -5,13 +5,13 @@
           
       
     
    -L/
    +/,
     
    -See Us AT NAB-Booth #C-2660 TRE ol a1 = Ao aa!
    +See Us AT NAB-Booth #C-2660 www.sachtler.com
     
    -YT ]
    +sachtler,
     corporation of america
    -F 55. North Main Stréet, Freeport, N.Y. 11520
    +& 55. North Main Stréet, Freeport, N.Y. 11520
     f"‘&w Phone: 516 867 4900  Fax: 516 623 6844
     Saalag W email: sachtlerUS@aol.com
     S— sachtler
    diff --rec -u old/sim_television-broadcast-tvb_2003-04_26_4_0048.txt new/sim_television-broadcast-tvb_2003-04_26_4_0048.txt
    --- old/sim_television-broadcast-tvb_2003-04_26_4_0048.txt	2020-10-27 09:51:47.066597411 +0100
    +++ new/sim_television-broadcast-tvb_2003-04_26_4_0048.txt	2020-10-27 14:36:16.834191712 +0100
    @@ -1,8 +1,8 @@
    -e
    +"
     
      
     
    -|
    +3
     "
     
      
    @@ -63,15 +63,15 @@
     
      
     
    -e g
    +o
     
    -A\ ol B AY
    +"I\ ™ o AY
     
             
     
     THE PROFESSIONAL OPTICAL
     DISC SYSTEM WILL ENHANCE
    -QY= 115 HO 27520
    +EVERY STEP OF VIDEO
     WORKFLOW, FROM ACQUISITION
     TO POSFPRODUCTION
     
    @@ -84,14 +84,14 @@
     duction, will be greatly improved, says
     Theresa Alesso, director of marketing
     for the optical and network products
    -1118
    +group.
     
     “Our professional optical disc sys-
     tem will offer dramatically faster edit-
     ing, faster transfer from field to stu-
     dio, far easier identification of record-
     ed assets, and lower operating costs,”
    -LE R E TN
    +says Alesso.
     
     Sony'’s two professional optical disc
     camcorders will enable users to mark
    diff --rec -u old/sim_television-broadcast-tvb_2004-05_27_5_0014.txt new/sim_television-broadcast-tvb_2004-05_27_5_0014.txt
    --- old/sim_television-broadcast-tvb_2004-05_27_5_0014.txt	2020-10-27 14:30:45.231753729 +0100
    +++ new/sim_television-broadcast-tvb_2004-05_27_5_0014.txt	2020-10-27 14:36:19.318209920 +0100
    @@ -3,7 +3,7 @@
      Register Today
     
     | FOR THE EAST COAST'S
    -Y\ [ 33 AT DI W YD
    +' LARGEST VIDEO AND AUDIO
     | CONFERENCE AND EXPOSITION!
     
      
    @@ -41,7 +41,7 @@
     
     GovernmentVideo  videography
     
    -FIALH
    +FEATURING

MerlijnWajer:
Thanks for this!

I'm happy to run a set of tests for some Internet Archive items. It would be easier for me if the threshold is a parameter, but can also build Tesseract locally. Did the higher threshold have a significant impact on the processing speed? I assume it probably doesn't have a significant impact, since only a small portion will be inverted & processed again?

zdenop:
@Shreeshrii : I do not plan 4.x release at the moment:
- it is quite difficult to make release with breaking API/ABI... ;-)
- changes in master are quite significant (code modernisation, source reorganisation), so backporting interesting commits start to be time expensive
- personally I do not have time a lot of free time

stweil:
I also see no urgent need for a new Tesseract 4 release.

We might consider tagging new releases from Git master as the last one was `5.0.0-alpha`. I suggest to name them `5.0.0-alpha-YYYYMMDD`, for example `5.0.0-alpha-20201027`, or a little bit longer ``5.0.0-alpha-2020-10-27`.

stweil:
> Did the higher threshold have a significant impact on the processing speed? I assume it probably doesn't have a significant impact, since only a small portion will be inverted & processed again?

That's right. The most significant impact on the performance comes from testing the mean confidence value instead of the minimum confidence value. A lot of lines contain some part with low word confidence, and all those lines were processed twice with the old code.

stweil:
@zdenop, @egorpugin, @Shreeshrii, @amitdo: Would it be okay if I tag a new release like I described it above? Or are there other suggestions?

egorpugin:
If you set pre-release tag, I think it's ok.

stweil:
>  I think it would be cool if we switch to std::{string,vector etc.} types before it.

Yes, that's why I think it is worth waiting with 5.0.0.

amitdo:
`5.0.0-alpha-YYYYMMDD`

+1

egorpugin:
According to https://semver.org/ we can do `5.0.0-alpha.YYYYMMDD` (dot instead of second dash).
But I'm ok with both.

Shreeshrii:
Thanks @stweil. I like the idea of having the date as part of the tag. 

What will the version strings for later commits look like?

M3ssman:
What about switching from semantic versioning to calendar versioning?

The next official version be `2021.03` (example).

This scheme (used by Ubuntu, Eclipse, IntelliJ, Windows, Docker ... and OCR-D/all) signals an ongoing development and advances regular releases. Usually, it expresses versions on monthly or even daily schedules. And most important: it decouples release management from semantically sophistries, to decide what sort of changes are included in the next version and how they might affect application users. Because, these decisions are quite often not so straight forward:  

What are the reasons for tagging a new major version? 

Semantically, this is widely used to communicate changes to the external API that client applications *might* break (https://en.wikipedia.org/wiki/Software_versioning#Sequence-based_identifiers). If there are *only* changes in behavior, even clear improvements (like with this PR), this is something desired, but this will not affect client applications (like https://github.com/sirfz/tesserocr). 
Changes of the underlying code base should be noted as refactoring, which means, that they don't even change existing behavior, because the new implementation should produce the same results as the old one did. 
From a developers point of view I can comprehend the psychological demand for a new major version, just to underline that internally so many things were changed and so much efforts have had been taken to improve or refactor, but from an interface perspective I cannot obey.

Mixing versioning schemes is not considered a good practise (https://mitchdenny.com/dates-in-version-numbers/)

Shreeshrii:
@M3ssman FYI, see https://abi-laboratory.pro/?view=timeline&l=tesseract

M3ssman:
@Shreeshrii Thanks, this is indeed very informative. I didn't knew this service before. Even extracts a change log!

stweil:
> According to https://semver.org/ we can do 5.0.0-alpha.YYYYMMDD (dot instead of second dash).

We switched to semver versioning some time ago, so of course the new release tag must match the semver rules.

`5.0.0-alpha-YYYYMMDD` is compatible with semver: the patch version is `5.0.0`, and the pre-release identifier is `alpha-YYYYMMDD`. `.0.0-alpha.YYYYMMDD`would be possible, too, and has two pre-release identifiers `alpha` and `YYYYMMDD`.

egorpugin:
Check grammar https://semver.org/

> <version core> "-" <pre-release> "+" <build>

As I understand there are no any other "-" dash symbols except for splitting core and pre-release.

stweil:
> What will the version strings for later commits look like?

We would use the pattern `5.0.0-alpha-YYYYMMDD` until we have release candidates like `5.0.0-rc1` or the final `5.0.0`. A Tesseract version string would look like `5.0.0-alpha-20201217-2-g58aa` for example.

Shreeshrii:
https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

stweil:
> As I understand there are no any other "-" dash symbols except for splitting core and pre-release.

Dash can be part of the pre-release identifier:

    <non-digit> ::= <letter>
                  | "-"

egorpugin:
Ah, I see, yes.

Shreeshrii:
Heads up for tag for proposed new 5.0.0-alpha-YYYYMMDD release
@lvc for updating abi-tracker
@AlexanderP for PPA etc

linuxhw:
ABI-3

See https://abi-laboratory.pro/index.php?view=timeline&l=tesseract

M3ssman:
Well then, if you like to tag a release with mixing semantic versioning with date plus commit hash, where the latter two are implicitly provided by the tag itself already. I'm just afraid that all the pre-beta-gamma-whatever stuff will lead to a situation where the next stable release transforms Tesseract into Tesseract Nukem Forever (or Perl 6 or ... ). 

I'm looking straight from a user's point of view in this case. Especially when we start our succeeding historical newspaper project next spring, I'd like to have something that is a) considerably stable and b) something more recent than 4.1.1 to reflect enhancements in open source. If the release is tagged candidate, I cannot communicate it's stable, whereas a calendar version represents just some state of time.

Other points to concern: 
1) There are still some issues tied to milestone 4.1.1 (https://github.com/tesseract-ocr/tesseract/milestone/5). 
What happens with these? 
2) If there's a hard cut between version 4 and 5, must both major releases be maintained?

stweil:
Thanks for the reminder. As 4.1.1 already exists and there is no urgent need to fix those issues in 4.1.x, I moved the milestones to 5.0.0 now. One of the issues could also be closed.

As long as there are only few people working on Tesseract, I see no chance for full support of more than a single major release. That means only critical bugs will be fixed in 4 by backporting of changes from 5. But this is open source development, so anybody can send pull requests to backport compatible changes to 4.

Regarding the release tag, we don't mix. We use semantic versioning which allows pre-release identifiers with a date part. We also don't add a commit hash to the tag. Such commit hashs are added automatically to the Tesseract version string for builds which are not based on a tagged release.

amitdo:
It a bit complicated issue.

We have two kinds of users:
* tesseract command line users (regular users).
* libtesseract users (developers).

The reason we did not release 5.0.0 yet is that we still plan to break libtesseract ABI, and we don't want to release 5.0.0 and then to release 6.0.0 with a new ABI 6/12 months later.

But that leaves the regular users with the one year old 4.1.1, and we don't know when we will release 5.0.0.

When people report about an issue they have with 4.1.1. many time we tell them: "Try 5.0.0...".

There is also the LTS users with there old/very old versions of tesseract.

Many users won't use a alpha/beta version  on production. so they want a new stable release.

I don't know what can we do about this issue in a way that pleases most users. You should remember that we have a limited resources. Ignore the 'by Google" we still advertise. They don't participate in this project anymore.

stweil:
The new pre-release [5.0.0-alpha-20201224](https://github.com/tesseract-ocr/tesseract/releases/tag/5.0.0-alpha-20201224) is available now. Merry Christmas and thank you to everybody.

Shreeshrii:
Merry Christmas @stweil. Thank you and all the contributors.

MerlijnWajer:
JFYI as a follow up, this is what happened to archive.org's OCR speed once we switched from 4.1.1 to 5.0.0-alpha-202012-31.

ocr-perf

Seems like the 20-30% estimated speedup holds even at 3 million pages per day. Awesome. :-)

amitdo:
> I got some [test images](https://archive.org/~merlijn/tesseract-images/) from the Internet Archive. The new code gives better results for many lines in those images, but some lines which were inverted with the old code are now no longer inverted and are not recognized.
> 
> Changing the threshold from 0.5 to 0.7 still got all improvements, but fixed the lines with regressions. If that is confirmed in more tests, it should be changed in the code. Maybe the threshold could be a parameter.

@stweil,

Maybe we should change the threshold to `0.7` or add a `invert_threshold` parameter for 5.0.1?

stweil:
This should have be done for the 5.0.0 release, but I missed it because the issue was neither assigned to a project nor to a milestone.

amitdo:
Can this change be done in 5.2.0, or do you prefer to change it in 6.0 ?

stweil:
Thanks for the reminder. I think it can be done in 5.2.0.

stweil:
> Maybe we should change the threshold to `0.7` or add a `invert_threshold` parameter for 5.0.1?

I just try to implement such a parameter. It could make the existing parameter `tessedit_do_invert` redundant, because a threshold greater than 0.0 could be equivalent to `tessedit_do_invert == true`.

Should the new parameter be called `invert_threshold` as suggested or is there a better name (`tessedit_invert_threshold`)?

amitdo:
Choose your preferred option.

rmast:
I bumped into this discussion. I have a image with both normal and inverted text that doesn't quite recognize well on the left bottom with language nld. However, when I invert it, all text seems to be recognized right, inverted or not...

https://github.com/internetarchive/archive-pdf-tools/issues/55

stweil:
That image is recognized much better with model `script/Latin` instead of `nld`. But the results for `nld` also improve with parameter `-c invert_threshold=0.9` in my new code.

rmast:
Is the double content in the resulting hocr by design? Should the recieving party pick the best option?

Outlook voor Android downloaden<https://aka.ms/ghei36>
________________________________

stweil:
Is there double content? Then that's wrong.

MerlijnWajer:
@rmast - I would test with just Tesseract, not other tools like OCRMyPDF that use Tesseract, and then see if you still get double content. I have never seen this - the only issue I've seen is diplopia (for which is there is an open MR), but that is only per character.

rmast:
tesseract 175789293-f39ddfdb-6f3e-4598-8d16-80a1f4a88b36.jpg out -l nld -c invert_threshold=0.9 hocr

For example "wis - clear" twice at slightly the same spot, by the way both with sufficient but slightly different confidence.
```
   <div class='ocr_carea' id='block_1_27' title="bbox 2164 361 2399 410">
    <p class='ocr_par' id='par_1_37' lang='nld' title="bbox 2149 361 2399 410">
     <span class='ocr_line' id='line_1_76' title="bbox 2164 361 2399 410; baseline -0.008 -14.121; x_size 28.333334; x_descenders 5.3333335; x_ascenders 7">
      <span class='ocrx_word' id='word_1_308' title='bbox 2164 375 2175 396; x_wconf 49'>&gt;</span>
      <span class='ocrx_word' id='word_1_309' title='bbox 2194 361 2341 410; x_wconf 92'>wis-clear</span>
      <span class='ocrx_word' id='word_1_310' title='bbox 2373 361 2399 410; x_wconf 46'>|</span>
      <span class='ocrx_word' id='word_1_311' title='bbox 2194 372 2237 395; x_wconf 90'>wis</span>
      <span class='ocrx_word' id='word_1_312' title='bbox 2249 384 2257 388; x_wconf 85'>-</span>
      <span class='ocrx_word' id='word_1_313' title='bbox 2269 372 2335 395; x_wconf 96'>clear</span>
     </span>
    </p>
   </div>
```

```
tesseract 5.1.0-71-g7c0c
 leptonica-1.79.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 2.0.3) : libpng 1.6.37 : libtiff 4.1.0 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.1
 Found AVX512BW
 Found AVX512F
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
 Found OpenMP 201511
```

amitdo:
@rmast, mabye #3489 is related to your issue.

rmast:
Yes, I focused on that one instance by accident, the rest of the file doesn’t contain double content, so I agree.

stweil:
`git bisect` shows that before commit 2252936fc801aee6c65a0bae9daee0638f56febf that example of double content just returned "EEEN wis - clear".

amitdo:
@stweil, do you consider this issue a blocker for releasing 5.2.0?

rmast:
> @rmast, mabye #3489 is related to your issue.

I tried to understand #3489 to look whether my fix would apply, but I guess #3489 has to do with the difference between automatic segmentation and 'straight' -psm 4 processing of a line that really isn't that straight. My issue was about inversion which has implementations in Tesseract on different levels.

rmast:
I’m reverting to your last commit before that commit, as that first older commit doesn’t compile.

4366d811d44d9c2331dbb4405727e0251bdc01d7

I want to confirm your git bisect-finding. Never heard of it…

I’ll use the old tessedit_do_invert=True.

PR #3155, opened 2020-11-18
Fix CMake targets path

Shatur95:
Fix path issue mentioned in #3112 by @yuriisk.


Shatur95:
Fix path issue mentioned in #3112.

Issue #3099, opened 2020-09-17
[  FAILED  ] ResultIteratorTest.ComplexTest with current master


Shreeshrii:
[ RUN      ] ResultIteratorTest.ComplexTest
[INFO]  BBox: [L:76, T:4, R:144, B:15]
[INFO]  BBox: [L:219, T:3175, R:277, B:3191]
[INFO]  BBox: [L:1115, T:368, R:1439, B:376]
[INFO]  BBox: [L:1111, T:387, R:1434, B:424]
[INFO]  BBox: [L:281, T:600, R:1237, B:3137]
[INFO]  BBox: [L:217, T:3192, R:1053, B:3225]
[INFO]  BBox: [L:870, T:1030, R:1682, B:2278]
[INFO]  BBox: [L:1082, T:2319, R:1462, B:2612]
[INFO]  BBox: [L:1311, T:602, R:2267, B:3139]
[INFO]  outfile = ./tmp/failedxor.png
[INFO]  At level 0: pix diff = 2073
../../../unittest/resultiterator_test.cc:117: Failure
Expected: (pixcount) <= (max_diff), actual: 2073 vs 400
[INFO]  BBox: [L:76, T:4, R:144, B:15]

...

[INFO]  BBox: [L:2170, T:3112, R:2188, B:3139]
[INFO]  BBox: [L:2204, T:3113, R:2221, B:3131]
[INFO]  BBox: [L:2223, T:3112, R:2245, B:3131]
[INFO]  BBox: [L:2247, T:3100, R:2267, B:3131]
[INFO]  outfile = ./tmp/failedxor.png
[INFO]  At level 4: pix diff = 2090
../../../unittest/resultiterator_test.cc:117: Failure
Expected: (pixcount) <= (max_diff), actual: 2090 vs 650
[  FAILED  ] ResultIteratorTest.ComplexTest (1710 ms)

stweil:
That changed with @bertsky's commit 65a077d3e9cd3be1f2ef97b84bed9f5649722775.

It is either a regression or an improvement (which needs an update of the test case). That still has to be checked.

bertsky:
Thanks for alerting me – I'll have a look. (Not sure if I ran the regression tests on that one...)

Shreeshrii:
>stweil commented 14 days ago in resultiterator_test fails #3122

>Pull request #3082 improved the layout recognition. This changes the OCR results, so the resultiterator_test currently fails and needs an update.

[...]

PR #3158, opened 2020-11-23
fixes issue #3099

Shreeshrii:


amitdo:
To link to an issue you need to add the `#` symbol before the issue number.

Fix issue #3099.

amitdo:
Does the test pass with this patch?

Shreeshrii:
Yes, it does.

I used the actual results from the test run and put those as expected
results. See the original report which had the lines with the error.

amitdo:
Thanks, Shree.

PR #3162, opened 2020-11-28
Specify ADD_RT for android platform, which embeds librt in libc

xloem:
I needed this change to build tesseract 3.05.02 using the android ndk.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -109,6 +109,9 @@ case "${host_os}" in
     powerpc-*-darwin*)
         OPENCL_LIBS=""
         ;;
+    *android*)
+        AM_CONDITIONAL([ADD_RT], false)
+        ;;
     *)
         # default
         AM_CONDITIONAL([ADD_RT], true)


zdenop:
3.05 is nut supported and you push it to master branch (5.0)

stweil:
@Robyer, does this change help for Android builds?

egorpugin:
Assuming https://developer.android.com/ndk/guides/stable_apis
Note that on Android, unlike Linux, there are no separate libpthread or librt libraries. That functionality is included directly in libc, which does not need to be explicitly linked against.

Robyer:
@egorpugin It looks like this PR was too old and it's fix is already included in current code, so the merge should be reverted, See: https://github.com/xloem/tesseract/blob/fed6603a77dfb628bd4bcde43484c68f89532e92/configure.ac#L106-L108

egorpugin:
Uh-oh.

PR #3163, opened 2020-11-30
fix invalid unsigned arithmetic.

ihsinme:
your check is incorrect. since the variables are unsigned, it is i != j.
so I suggest a simple fix for the error.

### DIFF
--- a/src/textord/cjkpitch.cpp
+++ b/src/textord/cjkpitch.cpp
@@ -733,7 +733,7 @@ bool FPRow::Pass2Analyze() {
         if (!is_final(j)) intersecting = true;
       } else {
         not_intersecting = true;
-        if (i - j > 0) {
+        if (i > j) {
           // Merge character(j+1) ... character(i) because they fit
           // into the body nicely.
           if (i - j == 1) {
@@ -781,7 +781,7 @@ bool FPRow::Pass2Analyze() {
         if (!is_final(j)) intersecting = true;
       } else {
         not_intersecting = true;
-        if (j - i > 0) {
+        if (j > i) {
           if (j - i == 1) {
             if (!skipped_whitespaces) mark_good(i);
             if (box(i).width() <= estimated_pitch_ * 0.5) {


stweil:
`j` is signed. Why do you think this change is needed nevertheless?

ihsinme:
> `j` is signed. 
yes. but https://wiki.sei.cmu.edu/confluence/display/c/INT02-C.+Understand+integer+conversion+rules  

and  
```cpp
sizeArray = 60; (sizeArray - 10) = 50; true
sizeArray = 10; (sizeArray - 10) = 0; false
sizeArray = 1; (sizeArray - 10) = 18446744073709551607; true
sizeArray = 0; (sizeArray - 10) = 18446744073709551606; true
if (sizeArray - 10 > 0)
```

egorpugin:
Better question and a fix - why is `i` unsigned (size_t)?
1. size_t i = numchars() https://github.com/tesseract-ocr/tesseract/blob/2ba69ae888ca0bdfd9b431163509c9740599f8e8/src/textord/cjkpitch.cpp#L700
2. numchars() returns chars_.size() https://github.com/tesseract-ocr/tesseract/blob/2ba69ae888ca0bdfd9b431163509c9740599f8e8/src/textord/cjkpitch.cpp#L336
3. genericvector.size() returns int https://github.com/tesseract-ocr/tesseract/blob/2ba69ae888ca0bdfd9b431163509c9740599f8e8/include/tesseract/genericvector.h#L71

Having both i and j as ints should not create a problem in `i - j > 0` expr.

ihsinme:
I agree this is also a valid way to fix the problem.

ihsinme:
Having estimated the complexity, I again propose to accept my PR as an intermediate solution. and in the subsequent preparation of more global changes with type replacement.

stweil:
I don't think that using `int` again like it was used in earlier versions goes into the right direction. I'd like to use `std::vector` or other standard types instead of `GenericVector`, and standard types use `size_t` for `size()` methods.

src/textord/cjkpitch.cpp
```diff
@@ -733,7 +733,7 @@ bool FPRow::Pass2Analyze() {
         if (!is_final(j)) intersecting = true;
       } else {
         not_intersecting = true;
-        if (i - j > 0) {
```
stweil, line 736: The current code works correct, so there is no need to change it. `j` is initially equal to `i` (so `(i - j) > 0 is false) and then possibly decremented (in that case the test is always true).

ihsinme: if these conditions about i and j are true, then everything is really ok. I would appreciate a demonstration of this statement in code.

if it's not time consuming.

egorpugin:
> I don't think that using `int` again like it was used in earlier versions goes into the right direction. I'd like to use `std::vector` or other standard types instead of `GenericVector`, and standard types use `size_t` for `size()` methods.

Is it possible to remove generic vector completely?

stweil:
> Is it possible to remove generic vector completely?

Yes, I think so. `STRING` is the other type which I would like to replace completely.

src/textord/cjkpitch.cpp
stweil: > I would appreciate a demonstration of this statement in code.

What kind of demonstration are you thinking of?

PR #3172, opened 2020-12-04
tesseract.1.asc: sync with languages available in tessdata-fast

MerlijnWajer:
`cos`, `div`, `fao`, `fyr`, `gla`, `hye` are available in Ubuntu's 'tesseract-ocr-*' packages but not mentioned in the manpage.

I will look at updating the file as mentioned here as well https://tesseract-ocr.github.io/tessdoc/Data-Files-in-different-versions.html


Shreeshrii:
>I will look at updating the file as mentioned here as well https://tesseract-ocr.github.io/tessdoc/Data-Files-in-different-versions.html

Please also see https://github.com/tesseract-ocr/tessdoc/blob/master/Data-Files.md#data-files-for-version-400-november-29-2016

So, there are different sets of files for 4.00 and then for 4.0.0.  I think the languages that I have listed under the column for 4.0.0 are actually for 4.00. The table has not been updated later for 4.0.0 / tessdata_fast which also has the script traineddata files.

doc/tesseract.1.asc
```diff
@@ -203,11 +205,14 @@ following languages:
 *est* (Estonian),
 *eus* (Basque),
 *fas* (Persian),
+*fao* (Faroese),
 *fil* (Filipino),
 *fin* (Finnish),
 *fra* (French),
 *frk* (Frankish),
 *frm* (French, Middle, ca.1400-1600),
+*fry* (Western Frisian),
```
stweil, line 214:
```suggestion
*fry* (West Frisian),
```

stweil: See https://en.wikipedia.org/wiki/West_Frisian_language.

stweil (approved):

doc/tesseract.1.asc
MerlijnWajer: The reason I wrote that is the iso639 language pack on Python showing that as name:

```
$ python
Python 3.8.6 (default, Nov 26 2020, 23:33:31)
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from iso639 import languages
>>> languages.get(part3='fry').name
'Western Frisian'
```

stweil: Ah sorry, yes, obviously the English Wikipedia and ISO 639 disagree on the name. Would it help to use the ISO name, or is the Wikipedia name fine? I am not a native speaker, so I am not sure which of both variants is more common in daily life.

MerlijnWajer: I have no preference, and I also don't know what's common for the Tesseract documentation and man pages, just wanted to point out why I picked "Western". I speak Dutch but not Frysian. :-) 

I have seen the `iso639` python library return alternative names for some languages (`.names` attribute), but at least for `fry` it does not list `West Frisian` as an alternative.

PR #3179, opened 2020-12-11
Improve CMake configuration

Shatur95:
* Use `find_dependency` instead of `find_package`. It's recommended to use [find_dependency](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) for CMake config files ([official tutorial](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-a-package-configuration-file)).
* Now all CMake config files located in `lib/cmake/tesseract`. This is a more common practice. Previously `TesseractConfig.cmake` and `TesseractVersion.cmake` files was `lib/cmake/tesseract/tesseract` (additional subfolder).
* Fix typo in `Tesseract_INCLUDE_DIR` -> `Tesseract_INCLUDE_DIRS` (as noted in the description if this file).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -374,11 +374,11 @@ configure_file(
 include(CMakePackageConfigHelpers)
 configure_package_config_file(
     cmake/templates/TesseractConfig.cmake.in
-    ${CMAKE_CURRENT_BINARY_DIR}/cmake/TesseractConfig.cmake
-    INSTALL_DESTINATION lib/tesseract/cmake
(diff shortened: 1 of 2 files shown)


stweil:
This commit deleted the `test` directory which is needed for the `test` submodule. I'll add a commit to fix that.

Shatur95:
@stweil, omg, I am very sorry, this happened by mistake.

stweil:
It's now fixed again, so no reason to worry.

PR #3186, opened 2020-12-19
automake: Flat build for src/training

stweil:


stweil:
The flat build allows massive parallel builds, reducing the time for `make training -j64` to about 10 s.

stweil:
Note that the training tools are now built in the build root directory (not in `src/training`).

Shreeshrii:
Will they be built by running make alone or will it still be a two step
process of make and make training?

stweil:
`make training` also builds the `tesseract` executable – like before.
`make` or `make all` does not build the training tools – like before.

PR #3190, opened 2020-12-22
Force fontconfig pangocairo backend for pango_font_info_test

stweil:
This fixes the test on platforms like MacOS (maybe also Windows) which don't use the fontconfig backend by default.
Fix and improve also some other smaller issues in the pango_font_info.cpp code.


zdenop:
@stweil : thanks
@egorpugin : sw build for [mac is failing](https://github.com/tesseract-ocr/tesseract/pull/3190/checks?check_run_id=1597043697) - at dependency building level...

egorpugin:
Will check.

PR #3191, opened 2020-12-23
Fix more unittests for MacOS and add dependencies for unittest

stweil:


amitdo:
Fonts:

* Will this download the fonts every time the tests are invoked?
* Maybe we should have the open source fonts in one of our repos?

unittest/Makefile.am
```diff
@@ -495,8 +498,31 @@ fonts += Times_New_Roman.ttf
 fonts += UnBatang.ttf
 fonts += Verdana.ttf
 
+# These tests depend on installed model files and fonts:
+#
+# apiexample_test baseapi_test lang_model_test layout_test
+# ligature_table_test loadlang_test lstm_recode_test lstm_squashed_test
+# lstm_test lstmtrainer_test mastertrainer_test osd_test
+# pagesegmode_test pango_font_info_test progress_test#
```
amitdo, line 506: An unwanted '#' at EOL.

stweil:
>     * Will this download the fonts every time the tests are invoked?

No, it should only download it once. It also respects existing symbolic links to fonts which are installed from the Linux distribution.

>     * Maybe we should have the open source fonts in one of our repos?

That is not possible for the Microsoft fonts.

unittest/Makefile.am
stweil: Ah, thanks, I'll fix that.

amitdo:
The macOS build failed. Why?

stweil:
> The macOS build failed. Why?

It failed when compiling `gunixconnection.c`, so this is unrelated to Tesseract code.
I think @egorpugin will have a look as soon as his time permits.

egorpugin:
Yes, sorry for that.
I've updated pango package, it required glib.gio package what was not tested on macos yet.

PR #3194, opened 2020-12-25
embed pdf.ttf to tesseract library #2551

zdenop:


stweil:
We could also create `pdf_ttf.h` during the build process instead of adding it to the sources.

zdenop:
This would requires additional multi-platforms tool - just for build.  I think easier is to provide generated file (without removing pdf.ttf).

stweil:
Yes, `pdf.ttf` should not be removed. The build could be done using either Perl, Python3 or sed. All of them are available for the relevant platforms, and we already use Python3 for training. Do you think it would harm using it for the build, too?

zdenop:
Personally I do not see any benefit from this (maybe I miss something): instead of final and tested header file you propose to include script (that we do not have)?. I prefer possibility to build [tesseract with minimal extra dependencies](https://sk-spell.sk.cx/building-minimalistic-tesseract). To install python (perl?) just to generate one small file is IMO overkill.
 
I prefer to adjust "final product" instead of tweaking script - e.g. my first commit work for me on windows (clang, msvs) and linux (opensuses), but appveyor and travis were not happy with it...

amitdo:
Both ways have their advantages and disadvantages.

This specific file does not change very often, so I think it's fine to go with the manually generated way.

amitdo:
Thanks!  :-)

src/api/pdfrenderer.cpp
```diff
@@ -623,24 +625,21 @@ bool TessPDFRenderer::BeginDocumentHandler() {
 
   stream.str("");
   stream << datadir_.c_str() << "/pdf.ttf";
```
stweil, line 627: Wouldn't it be cleaner to remove the code for reading `pdf.ttf` and remove that file from installations, too?

zdenop: I leave it there if somebody would like to adapt and test pdf.ttf

stweil: Would it be sufficient to enable that feature only for `DEBUG` code? For most users (and that includes all Linux distributions) installing `pdf.ttf` is not necessary, and I'd prefer to remove it from the normal installation.

zdenop: I think using it in DEBUG code only it good idea.

PR #3195, opened 2020-12-26
Add const attribute for pdf_ttf

stweil:
That moves its data into the text segment and reduces the total size
slightly:

  39788	    693	      0	  40481	   9e21	old/libtesseract_la-pdfrenderer.o
  40360	     88	      0	  40448	   9e00	new/libtesseract_la-pdfrenderer.o

Signed-off-by: Stefan Weil <sw@weil.de>


egorpugin:
After relative "..." includes must go this project includes `<tess/...>`, then 3rd party - `<leptheaders>`, then std c++, std c (if any), system headers.
Please, reorder in this way, if you are touching headers.

PR #3196, opened 2020-12-27
Fix `tesseract --list-langs` and replace several GenericVector

stweil:


src/api/baseapi.cpp
```diff
@@ -1903,7 +1909,9 @@ void TessBaseAPI::End() {
   delete block_list_;
   block_list_ = nullptr;
   if (paragraph_models_ != nullptr) {
-    paragraph_models_->delete_data_pointers();
+    for (auto model : *paragraph_models_) {
+      delete model;
```
egorpugin, line 1913: Not good.
We should consider unique ptrs or something like this here.
I also tried to replace `delete_data_pointers()`, but noticed that it frees memory of elements and reverted changes locally.

There are several common patterns over the codebase. I think those places must be upgraded uniformely.

egorpugin:
At the end I think I'm against raw `delete`.
Vector changes are fine except for eliminating `delete_data_pointers()`

src/api/baseapi.cpp
```diff
@@ -278,6 +278,9 @@ void TessBaseAPI::SetInputName(const char* name) {
 
 /** Set the name of the output files. Needed only for debugging. */
 void TessBaseAPI::SetOutputName(const char* name) {
+  if (name == nullptr) {
+    name = "";
```
egorpugin, line 282: This will be solved if we allow std::string everywhere.

include/tesseract/baseapi.h
```diff
@@ -815,7 +815,7 @@ class TESS_API TessBaseAPI {
   EquationDetect* equ_detect_;     ///< The equation detector.
   FileReader reader_;              ///< Reads files from any filesystem.
   ImageThresholder* thresholder_;  ///< Image thresholding module.
-  GenericVector<ParagraphModel*>* paragraph_models_;
+  std::list<ParagraphModel*>* paragraph_models_;
```
egorpugin, line 818: Also suspicious.
If vector worked here, why change to list?
Vector merges are find using .insert(end(), v2.begin(), v2.end());

stweil: The code removes elements. That's expensive for `std::vector`, but cheap for `std::list`.

src/api/baseapi.cpp
stweil: My first priority is currently getting a fully working git master again. A lot of code can be simplified or rewritten, but that can be done later.

include/tesseract/baseapi.h
egorpugin: I see. Legit then.
But I propose to optimize things at the end.
We try to update 1-to-1 code to STL containers first. Optimization will be at the very end.

stweil:
https://github.com/stweil/tesseract/tree/new-api is based on this pull request and completely removes `genericvector.h` and `strngs.h` from the public Tesseract API.

egorpugin:
> completely removes genericvector.h and strngs.h

I see.
In this case I think you could merge it and we'll fix issues here in master together?
I do not want to deny or keep such changes in review for a long time.

egorpugin:
We have unittests, we have base line for my changes (that alpha tag you made), so we can always compare things and fix new errors.

stweil:
This pull request is part of the more complete pull request #3198, so I close it.

PR #3198, opened 2020-12-28
New Tesseract API without GenericVector and STRING

stweil:


include/tesseract/serialis.h
```diff
@@ -43,14 +41,17 @@ constexpr size_t countof(T const (&)[N]) noexcept {
   return N;
 }
 
-// Function to read a GenericVector<char> from a whole file.
+// Function to read a std::vector<char> from a whole file.
```
egorpugin, line 44: Probably we should replace `vector<char>` with `vector<uint8_t>` for binary data later.

egorpugin:
Let's wait for builders and merge if they are green.

src/api/tesseractmain.cpp
```diff
@@ -299,12 +299,11 @@ static void SetVariablesFromCLArgs(tesseract::TessBaseAPI* api, int argc,
 }
 
 static void PrintLangsList(tesseract::TessBaseAPI* api) {
-  GenericVector<STRING> languages;
+  std::vector<std::string> languages;
   api->GetAvailableLanguagesAsVector(&languages);
   printf("List of available languages (%d):\n", languages.size());
```
stweil, line 304:
```suggestion
  printf("List of available languages (%zu):\n", languages.size());
```

stweil: This should fix the issue reported by LGTM. I don't add that now to allow the continuous integration to finish.

egorpugin:
I wonder if we see similar changes but "without `const char *`". :)

amitdo:
Lots of presents in the last few days :-)

@egorpugin and @stweil, thank you!

stweil:
@amitdo, wait with thanks until it works again. The good news is that it can be built, so it is possible to run tests.

amitdo:
std::list vs. std::vector:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slcon2-prefer-using-stl-vector-by-default-unless-you-have-a-reason-to-use-a-different-container

If you still think list is better in this particular case, than it's fine.

egorpugin:
I also think that std::list place is not really a hot path (is it?) and does not process gigs of data in vector/list, so it can be just vector.

amitdo:
list->vector
reverted in deec8ef46fbf2f

stweil:
Yes, the code no longer uses `std::list`. The new implementation should also be faster than the old one based on `GenericVector` which might improve the performance for large images with a complex layout.

egorpugin:
I assure you, when we replace more places with modern C++, performance will increase. :)

amitdo:
It seems we moved to C++17 as minimum requirement.

amitdo:
@egorpugin

More places to modernize:
clst, elst, elst2.

Macros can be replaced with templates.

PR #3199, opened 2020-12-29
Remove serialis.h from public API

stweil:


stweil:
I did not update the order of include statements. That should be done later.

PR #3203, opened 2020-12-30
Replace more GenericVector by std::vector and clean include statements

stweil:


egorpugin:
Vectors of bools must be eliminated completely over time.

PR #3205, opened 2020-12-30
Api

stweil:


stweil:
The truth callback was not used in the Tesseract code, and I doubt that some external library uses it.

I suggest to review the public API and make suggestions what else should be removed / changed / cleaned.

egorpugin:
> I suggest to review the public API and make suggestions what else should be removed / changed / cleaned.

Very true.

My propositions:
1. Switch to std::string.
2. Remove any pointers from there. Use unique ptr or something else instead of raw data (`char *` etc.).

stweil:
Other suggested changes:

- add API function to release the text which was returned by `GetUTF8Text` (allocated by `new char[]`) and similar functions

egorpugin:
> add API function to release the text which was returned by GetUTF8Text (allocated by new char[]) and similar functions

But is it possible to return it in a string? This is exactly I mentioned above.

stweil:
Using `std::string` would not solve the problem of releasing memory which was allocated in a different shared library Especially under Windows that is a common problem. It would also require additional wrappers for the Tesseract C API (or not work at all with that API). I don't know how `std::unique_ptr` could be used with the C API. It is very useful, but I am afraid that we can only use it internally.

egorpugin:
C API is definitely a question. (Actually is not, just return pointers like before or add tessmalloc() and tessfree() to use them for releasing data allocated by tesseract.)

But we don't have this issue on C++ side of api completely.
Need stability over many different heaps? Use C api.
Otherwise build libtess dll in the same config as the software. Same compiler, same stl to use same STD heap.

egorpugin:
And api chain for us: Internal Tess APIs -> Public C++ Tess APIs -> Public C Tess APIs.
C must be a wrapper to C++ api, not to internal apis. If it is not like that, this must be fixed.

stweil:
The C API is never used internally (just search for `capi.h`), so that's fine.

egorpugin:
I mean C API may use "src/ccutil/" or other "src/X/" files, headers. This is wrong.
C API must use C++ API, so only `include/tesseract/` must be allowed.

PR #3206, opened 2020-12-30
Make tmp directory for all unit tests, remove RecognizeForChopTest from public API and replace some old-style type casts

stweil:


egorpugin:
Thank you.

stweil:
@egorpugin, ommit 4366d811d44d9c2331dbb4405727e0251bdc01d7 in git master fixes `lgtm_test` and two other tests on platforms which complain when the non-existing first element of an empty `std::vector` is accessed.

egorpugin:
I tested couple of hours ago and all lstm tests were ok for me.

PR #3207, opened 2020-12-30
[DO NOT MERGE] Use std::unordered_map as unicharmap.

egorpugin:
@stweil 

I'm not sure in `minmatch()`.
Maybe I understood it incorrectly or even it has bad implementation before.

Also I did not change api to std::strings intentionally.


egorpugin:
Closing this, it will be still available for reference.

PR #3208, opened 2020-12-30
Run more unittests without requiring tensorflow

stweil:


unittest/baseapi_thread_test.cc
```diff
@@ -23,7 +23,9 @@
 #include <functional>
 #include <memory>
 #include <string>
+#ifdef INCLUDE_TENSORFLOW
```
egorpugin, line 26: Is it used only for thread pool in this test?
If yes, we should rewrite it to c++11 or abseil.

egorpugin: https://github.com/abseil/abseil-cpp/blob/master/absl/synchronization/internal/thread_pool.h

stweil: Yes. I'd prefer c++11 and reduce the use of abseil (which causes lots of compiler warnings).

egorpugin:
New tests work fine for me (57/60).

egorpugin:
@stweil 
If you remember something to change/update, put a note here (in progress col):
https://github.com/tesseract-ocr/tesseract/projects/1

I've added some general items as well as this test upgrade.

PR #3219, opened 2021-01-08
Delete TESSDATA_BEST_DIR macro

Shreeshrii:
See related discussion at https://github.com/tesseract-ocr/tesseract/pull/3218#issuecomment-756782983


stweil:
Thank you.

PR #3221, opened 2021-01-10
Fix use of NEON on ARMv8

Robyer:
Flag neon_available_ is automatically set to true when __aarch64__ is defined,
but the actual check for neon_available_ required having also HAVE_NEON defined.

Now we check the flag also when only __aarch64__ is defined.

### DIFF
--- a/src/api/tesseractmain.cpp
+++ b/src/api/tesseractmain.cpp
@@ -137,7 +137,7 @@ static void PrintVersionInfo() {
     }
   }
 #endif
-#if defined(HAVE_NEON)
+#if defined(HAVE_NEON) || defined(__aarch64__)
   if (tesseract::SIMDDetect::IsNEONAvailable()) printf(" Found NEON\n");
 #else
   if (tesseract::SIMDDetect::IsAVX512BWAvailable()) printf(" Found AVX512BW\n");
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -212,7 +212,7 @@ SIMDDetect::SIMDDetect() {
     // SSE detected.
     SetDotProduct(DotProductSSE, &IntSimdMatrix::intSimdMatrixSSE);
 #endif
-#if defined(HAVE_NEON)
+#if defined(HAVE_NEON) || defined(__aarch64__)
   } else if (neon_available_) {
     // NEON detected.
     SetDotProduct(DotProduct, &IntSimdMatrix::intSimdMatrixNEON);


stweil (changes requested):
This change is not needed with the autotools build which sets `HAVE_NEON` for ARMv8.

Did you build with cmake? It looks like the setting is missing there. @egorpugin, this should be fixed.

stweil:
@Robyer, the correct fix is setting `HAVE_NEON` in the build system (where it is currently missing for cmake). That would enable all code for Neon, not only the code in `simdetect.cpp`.

egorpugin:
@Robyer 
Try build after https://github.com/tesseract-ocr/tesseract/commit/a6321940dc8c9b292c283cd824db9f9acf569929 commit.

stweil:
I'm afraid that won't work. `-mfpu=neon` is not supported for ARMv8 compilers because they always use Neon.

Robyer:
@stweil @egorpugin Thanks for looking into this, I still think there is *something* wrong regarding the NEON support.

I am maintaining a [Tesseract4Android](github.com/adaptech-cz/Tesseract4Android/) library, where I build Tesseract with CMake and my simple [CMakeLists.txt](https://github.com/adaptech-cz/Tesseract4Android/blob/tesseract-5.0.0/tesseract4android/src/main/cpp/tesseract/CMakeLists.txt).

When I was updating to Tesseract 5.0.0-alpha, I naturally tried the [block from your CMakeLists](https://github.com/tesseract-ocr/tesseract/blob/a52237799338580c761b13ccc2bfd8b8ffaf4edb/CMakeLists.txt#L191-L198) where you set HAVE_NEON, but with that I've got a compile error when building the x86/x86_64 target (as Android supports armeabi-v7a, arm64-v8a, x86 and x86_64 [ABIs](https://developer.android.com/ndk/guides/abis)). Error was: `undefined reference to 'tesseract::IntSimdMatrix::intSimdMatrixNEON'`. I will get to this later.

So I've been looking at the code where HAVE_NEON is used and it is used only in `simdetect.cpp` and `tesseractmain.cpp`, but since I am on Android, I don't use the latter.

Now, NEON is supported only on armeabi-v7a (but not on every device, so we need runtime checks for that) and on arm64-v8a (always supported, so we doesn't need runtime checks as existence of `__aarch64__` definition is enough). As I was looking into `simdetect.cpp` code, I understood it this way:
- In order to compile armeabi-v7a, we need runtime check for NEON support, that is guarded by HAVE_NEON check so we need to define HAVE_NEON.
- In order to compile arm64-v8a, we need only `__aarch64__` which automatically sets the `neon_available_` flag to true. **But we can't reach the code that actually checks the flag and sets the `IntSimdMatrixNEON`, so that feels like bug in code** -> that's why I proposed this PR.

Let's get back to my build error when compiling on x86/x86_64. In your CMakeLists.txt is used check `CHECK_CXX_COMPILER_FLAG("-mfpu=neon" HAVE_NEON)`. [Docs](https://cmake.org/cmake/help/latest/module/CheckCXXCompilerFlag.html) says that it `Check that the <flag> is accepted by the compiler without a diagnostic.` So it doesn't really check NEON support on target platform, but only that current compiler understands the flag. So in my case the HAVE_NEON has been set for every build target and it somehow set `neon_available_=true` so it tried to also set `SetDotProduct(DotProduct, &IntSimdMatrix::intSimdMatrixNEON);` in simdetect.cpp, **but in `intsimdmatrixneon.cpp` there is correct check for `#if defined(__ARM_NEON)`, it couldn't find reference to the intSimdMatrixNEON.

So there are 3 different definitions used with NEON support: `__aarch64__`, `HAVE_NEON` and `__ARM_NEON`. And based on the above I think use of HAVE_NEON is used wrongly somewhere.

Robyer:
I can solve the issue by specifying `HAVE_NEON` in my CMakeLists.txt by explicitly checking armeabm-v7a/arm64-v8a and ignoring it on x86/x86_64, but use of that flag feels redundant for arm64-v8a if code already uses just `__aarch64__` to determine NEON support in this case, and also when there is another `__ARM_NEON` for guarding use of the instructions themselves in `intsimdmatrixneon.cpp`.

And in the case of `tesseractmain.cpp` there is block `#if defined(HAVE_NEON)` which then does `if (tesseract::SIMDDetect::IsNEONAvailable()) printf(" Found NEON\n");`, where `IsNEONAvailable()` internally checks the `neon_available_` flag. That is same situation as in my PR, so I would also append the `|| defined(__aarch64__)` there. For some reason I can't (or I don't know how to) append that change to this PR.

I still think my proposed changes makes things better than they are now, but I would understand if you have different idea.

stweil:
You are right and I was wrong: the old code already used `__aarch64__`, so there are no good arguments against your two commits which added two more of them. I merged your two commits in git master now. Thank you!

Shreeshrii:
Issue building with cmake on Aarch64 is still there - please see https://github.com/tesseract-ocr/tesseract/issues/3254#issuecomment-770374431

stweil:
That was caused by the incomplete cmake configuration (see my [comment above](https://github.com/tesseract-ocr/tesseract/pull/3221#issuecomment-757683637)) and is fixed now in git master.

Shreeshrii:
Thanks!

Issue #3001, opened 2020-06-02
Image too small to scale!! (2x48 vs min width of 3) - vertical lstm training 


davidb1:
### Environment

Tesseract Version: 4.00
Commit Number:
Platform: Ububtu 18.04

### Current Behavior:
I am generating vertical lstm training files using tesstrain.sh but when I try to train on them I get (on all the training data):

> Image too small to scale!! (2x48 vs min width of 3)
> Line cannot be recognized!!
> Image not trainable
> Compute CTC targets failed!

I couldn't find much on the problem except https://github.com/tesseract-ocr/tesseract/issues/590 and I couldn't find a solution there. 

Where should I be looking in order to fix this?

stweil:
Usually you can simply ignore that message. It is typically caused by very narrow glyphs (like `i`, `I`, `l` or `|`) in your training images, or there are bad training images with a small width. Could you please check that? Tesseract expects that a character image should be at least 3 pixels wide. Maybe we should reduce that limit to 2 pixels if such images are reasonable.

amitdo:
Duplicate of #2890.

niki13010:
@amitdo

davidb1:
@stweil thanks, I do have a lot of small glyphs but I don't see any other output other than these errors. Is the setting controlling the minimum glyph size in the network specification?
Is the whole image ignored if one glyph is too small (since the output is "Image not trainable")? 
![image](https://user-images.githubusercontent.com/13868996/83576644-dfd22980-a53a-11ea-857c-eb4741e262d3.png)

Additionally, any way to make the font bigger when creating the pages - since most of the characters look too dark and squished to be good training data?

YuvalFeldman:
I'm having the same issue as @davidb1

filippesic:
I have the same error but when I'm actually trying to do OCR. Tesseract v4.1.1-rc2-25-g9707 and error is` 2x36 vs min width of 3 Line cannot be recognized!!`, without the CTC error. There are a lot of errors, not just a few. Images that I'm trying to do OCR on are upscaled from 231 native DPI to 800DPI. I'm gonna try 600DPI too and maybe lower, but I had the most success with 800, even tho it may be too large, the text in the images is very small, it has a lot of separate columns and it's a two page scan.

Shreeshrii:
I investigated this a bit today, since there were reports of successful training of jpn_vert in the forum in 2018. So it seemed to me that maybe something was broken after that. 

[...]

PR #3223, opened 2021-01-12
No --psm 6 for lstm.train

Shreeshrii:
Fixes https://github.com/tesseract-ocr/tesseract/issues/3001 and other similar issues

### DIFF
--- a/src/training/tesstrain.sh
+++ b/src/training/tesstrain.sh
@@ -82,7 +82,7 @@ initialize_fontconfig
 phase_I_generate_image 8
 phase_UP_generate_unicharset
 if $LINEDATA; then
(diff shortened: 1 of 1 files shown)


stweil:
That reverts a part of commit b34cf9d424e88cd09aaa193697127c90ff76e0ce.

stweil:
@Shreeshrii, I think a similar fix is needed for `tesstrain.py`. Are there still reasons for keeping both Bash and Python scripts?

Shreeshrii:
@stweil I use tesstrain.sh only and many users who follow the tutorial would look for it.

I think the plan was to move to python only version for training and move the tesstrain.py script to tesstrain repo - it would need a change to create single line images instead of multi page tiffs.

I will make the PR for tesstrain.py also.

PR #3230, opened 2021-01-13
Remove ANDROID_BUILD macro

Robyer:
Build fails when ANDROID_BUILD is defined, because it removes parts of the LSTM engine, but there are still some unguarded references. But removing LSTM engine is not needed as it works perfectly fine on Android.

This macro doesn't provide any benefit anymore and is not even used in current build config. If needed, ANDROID macro should be used instead (which is already used on few places).

### DIFF
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -1398,7 +1398,6 @@ void Tesseract::classify_word_pass1(const WordData& word_data,
   BLOCK* block = word_data.block;
   prev_word_best_choice_ = word_data.prev_word != nullptr
       ? word_data.prev_word->word->best_choice : nullptr;
-#ifndef ANDROID_BUILD
 #ifdef DISABLED_LEGACY_ENGINE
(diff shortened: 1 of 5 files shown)


src/ccmain/control.cpp
```diff
@@ -1423,9 +1422,8 @@ void Tesseract::classify_word_pass1(const WordData& word_data,
                                     classify_bln_numeric_mode,
                                     textord_use_cjk_fp_model,
                                     poly_allow_detailed_fx, row, block);
-#endif  // ndef DISABLED_LEGACY_ENGINE
+  #endif  // ndef DISABLED_LEGACY_ENGINE
```
stweil, line 1425:
```suggestion
#endif  // ndef DISABLED_LEGACY_ENGINE
```

Robyer: @stweil This is intentional change. Look at the indent of opening macro at line 1419/1418.

stweil (approved):
ok, thanks

Issue #3220, opened 2021-01-09
lstm_test failing


Shreeshrii:
```
uname -a
Linux tesseract-ocr-1 4.15.0-124-generic #127-Ubuntu SMP Fri Nov 6 10:58:37 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```
```
tesseract -v
tesseract 5.0.0-alpha-20201231-100-g3c717
 leptonica-1.78.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.2) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
 Found NEON
 Found OpenMP 201511
 Found libarchive 3.2.2 zlib/1.2.11 liblzma/5.2.2 bz2lib/1.0.6 liblz4/1.7.1
 Found libcurl/7.58.0 NSS/3.35 zlib/1.2.11 libidn2/2.0.4 libpsl/0.19.1 (+libidn2/2.0.4) nghttp2/1.30.0 librtmp/2.3
```
```
lstmtraining -v
5.0.0-alpha-20201231-100-g3c717
```

```
[==========] 11 tests from 1 test suite ran. (1,013,640 ms total)
[  PASSED  ] 6 tests.
[  FAILED  ] 5 tests, listed below:
[  FAILED  ] LSTMTrainerTest.Test2D
[  FAILED  ] LSTMTrainerTest.DeterminismTest
[  FAILED  ] LSTMTrainerTest.SoftmaxBaselineTest
[  FAILED  ] LSTMTrainerTest.SoftmaxTest
[  FAILED  ] LSTMTrainerTest.EncodedSoftmaxTest

 5 FAILED TESTS
```

The test was passing earlier.  

The build was done with following:
```
export CXXFLAGS="-O3"
../../configure --enable-openmp --disable-shared --with-tensorflow=no --disable-debug  --disable-opencl
```

The error messages are:

```
../../../unittest/lstm_test.cc:135: Failure
Value of: trainer_->ReadTrainingDump(trainer_a_data, trainer_.get())
  Actual: false
Expected: true
```

```
../../../unittest/lstm_test.cc:140: Failure
Expected equality of these values:
  act_error_a
    Which is: 5.7480001
  act_error_b
    Which is: 36.243
../../../unittest/lstm_test.cc:141: Failure
Expected equality of these values:
  char_error_a
[...]

PR #3231, opened 2021-01-14
Add Github Action for autotools based daily unittests on ubuntu

Shreeshrii:
Suggested by @egorpugin in https://github.com/tesseract-ocr/tesseract/issues/3220#issuecomment-758246368

Thanks @stweil, Github Actions work after using ssh as suggested by you in https://github.com/tesseract-ocr/tesseract/issues/3220#issuecomment-758778709

### DIFF
--- a/.github/workflows/dailyunittest.yml
+++ b/.github/workflows/dailyunittest.yml
@@ -1,0 +1,54 @@
+name: dailyunittest
+
+on:
+  schedule:
+    # every day at  11:15 AM UTC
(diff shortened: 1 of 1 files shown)


Shreeshrii:
[logs_20.zip](https://github.com/tesseract-ocr/tesseract/files/5813694/logs_20.zip)

The zip file has logs from a test run. The only thing I am not sure about is whether the configure options are applied correctly.

egorpugin:
Feel free to send further patches based on results.

PR #3234, opened 2021-01-14
Try to fix MSVC unresolved symbols

stweil:


egorpugin:
It's not that simple. :)
See 
https://github.com/egorpugin/primitives/blob/master/src/settings/include/primitives/settings.h#L594
https://github.com/egorpugin/primitives/blob/master/src/settings/src/settings.cpp#L396

stweil:
It seems to work now.

PR #3236, opened 2021-01-14
lstmtraining: Interpret negative value for --max_iterations as epochs

stweil:

### DIFF
--- a/doc/lstmtraining.1.asc
+++ b/doc/lstmtraining.1.asc
@@ -41,7 +41,7 @@ OPTIONS
   Index in continue_from Network at which to attach the new network defined by net_spec  (type:int default:-1)
 
 '--max_iterations  '::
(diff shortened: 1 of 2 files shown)


stweil:
This simplifies using `tesstrain` with epochs instead of iterations. See also https://github.com/tesseract-ocr/tesstrain/pull/223.

amitdo:
Maybe a separate flag` --max_epochs_iterations` should be used instead of using a negative parameter?

stweil:
Yes, that is something to consider. I ask this question already for `tesstrain`: https://github.com/tesseract-ocr/tesstrain/pull/223#issuecomment-760408433. As `lstmtraining` is usually called by some wrapper (`tesstrain` Makefile, `tesstrain.sh` or `tesstrain.py`), it might be sufficient to add the separate flag for the epochs there.

PR #3242, opened 2021-01-17
Add Macports build with training tools for daily ci test

Shreeshrii:


egorpugin:
@stweil @Shreeshrii 
Please pay attention to these new nightly builds and tests on those configs.
https://github.com/tesseract-ocr/tesseract/actions?query=workflow%3Adailyunittest
https://github.com/tesseract-ocr/tesseract/actions?query=workflow%3Adailycmake
@stweil 
Also please check if workflows are correct and we can consider them like as you run tests, sanitizers on your machine.

Shreeshrii:
@egorpugin Thanks for merging these.  Yes, these need corrections and review. 

I also need to check why the macos build failed here. It worked till make check on by test repo - see https://github.com/Shreeshrii/testactions/runs/1717479792?check_suite_focus=true

Shreeshrii:
@egorpugin Please see what needs to be added to windows, I am getting error

```
  Could not find a package configuration file provided by "SW" with any of
  the following names:

    SWConfig.cmake
    sw-config.cmake

  Add the installation prefix of "SW" to CMAKE_PREFIX_PATH or set "SW_DIR" to
  a directory containing one of the above files.  If "SW" provides a separate
  development package or SDK, be sure it has been installed.
```

egorpugin:
Windows cmake build requires sw at the moment.
If you want to build it without sw, it will require a lot of effort.

PR #3250, opened 2021-01-20
Reduce number of INFO messages from lstm_test

Shreeshrii:
Some more of these messages can be reduced. Attached log from run after this change.

[lstm_test.log](https://github.com/tesseract-ocr/tesseract/files/5840764/lstm_test.log)

### DIFF
--- a/unittest/lstm_test.h
+++ b/unittest/lstm_test.h
@@ -119,9 +119,9 @@ class LSTMTrainerTest : public testing::Test {
       trainer_->MaintainCheckpoints(nullptr, &log_str);
       iteration = trainer_->training_iteration();
       mean_error *= 100.0 / kBatchIterations;
-      LOG(INFO) << log_str.c_str();
-      LOG(INFO) << "Best error = " << best_error << "\n" ;
-      LOG(INFO) << "Mean error = " << mean_error << "\n" ;
+      //LOG(INFO) << log_str.c_str();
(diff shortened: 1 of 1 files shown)


unittest/lstm_test.h
```diff
@@ -119,9 +119,9 @@ class LSTMTrainerTest : public testing::Test {
       trainer_->MaintainCheckpoints(nullptr, &log_str);
       iteration = trainer_->training_iteration();
       mean_error *= 100.0 / kBatchIterations;
-      LOG(INFO) << log_str.c_str();
-      LOG(INFO) << "Best error = " << best_error << "\n" ;
-      LOG(INFO) << "Mean error = " << mean_error << "\n" ;
+      //LOG(INFO) << log_str.c_str();
```
stweil, line 122: Why do you want to keep those statements as a comment instead of removing them completely?

Shreeshrii: You are right. They can be removed. I will modify the PR.

egorpugin: Maybe make them as LOG(DEBUG) or LOG(TRACE)?

Shreeshrii:
I had tried changing it to
LOG(DEBUG), got some compile error.

unittest/lstm_test.h
Shreeshrii: Removed.

egorpugin:
Of course you should do it in terms of GLOG.

Issue #3215, opened 2021-01-06
Failing unittest


stweil:
These tests fail in latest git master with `-fsanitize=address,undefined`:

    ligature_table_test.log
    mastertrainer_test.log
    pango_font_info_test.log
    stringrenderer_test.log

egorpugin:
Did it pass before?
We may broke some vector freeing functions (GenericVector<T*> templates).
I tried to not touch them for some time. GenVec<T> is simpler to replace.

Anyways considering that there are no much smart pointers on the codebase, it is not very surprising.

stweil:
Yes, all tests passed. `pango_font_info_test` for example fails since commit db43bb43dcb673167014f4203468a6553527e668.

egorpugin:
> ligature_table_test.log
mastertrainer_test.log
pango_font_info_test.log
stringrenderer_test.log

Attach these logs.

stweil:
`ligature_table-test` shows errors like this one:

    Running main() from ../../../../unittest/../googletest/googletest/src/gtest_main.cc
    [==========] Running 7 tests from 1 test suite.
    [----------] Global test environment set-up.
    [----------] 7 tests from LigatureTableTest
    [ RUN      ] LigatureTableTest.DoesFillLigatureTables
    unknown file: Failure
    C++ exception with description "locale::facet::_S_create_c_locale name not valid" thrown in SetUp().
    [  FAILED  ] LigatureTableTest.DoesFillLigatureTables (1 ms)
    [ RUN      ] LigatureTableTest.DoesAddLigatures
    unknown file: Failure
    C++ exception with description "locale::facet::_S_create_c_locale name not valid" thrown in SetUp().
    [  FAILED  ] LigatureTableTest.DoesAddLigatures (0 ms)
    [...]

`pango_font_info_test` and `stringrenderer_test` also have the same kind of problem.

egorpugin:
Much better.
1.
Why all tests run?
```
    static std::locale system_locale("");
    std::locale::global(system_locale);
```
I think this does nothing and can be removed from all tests.
Try it locally to check if `ligature_table_test` passes without it.

2.
Just a leak on raw pointer (`mastertrainer.cpp:147:20`). We need to replace it with smart ptr. As well as other many places.

stweil:
[...]

PR #3256, opened 2021-01-23
Fix memory leak in mastertrainer_test (fixes issue #3215)

stweil:
The issue was introduced in commit 6e9456415.

Partially reverting this commit fixes it.


stweil:
@egorpugin, unless you have a better fix I'd like to merge this commit to get that memory leak issue closed.

egorpugin:
Where is the leak? Do you know it?

Issue #3226, opened 2021-01-12
Unitests failing with --disable-legacy


Shreeshrii:
```
============================================================================
Testsuite summary for tesseract 5.0.0-alpha-20201231-113-g5104
============================================================================
# TOTAL: 52
# PASS:  47
# SKIP:  0
# XFAIL: 0
# FAIL:  5
# XPASS: 0
# ERROR: 0
============================================================================
See unittest/test-suite.log
Please report to https://github.com/tesseract-ocr/tesseract/issues
============================================================================
```
```
	Line 51: FAIL lstm_squashed_test (exit status: 134)
	Line 369: FAIL lstm_test (exit status: 134)
	Line 395: FAIL pagesegmode_test (exit status: 1)
	Line 10690: FAIL resultiterator_test (exit status: 1)
	Line 11432: FAIL textlineprojection_test (exit status: 1)
```

Reported first by @stweil in https://github.com/tesseract-ocr/tesseract/issues/3211#issuecomment-753660884

Shreeshrii:
After PR https://github.com/tesseract-ocr/tesseract/commit/b0e46085f48d206797ad36ba643dbe9f12ad157a which fixes the lstm tests, the failing tests are same as original report:

```
============================================================================
Testsuite summary for tesseract 5.0.0-alpha-20201231-115-gb0e46
============================================================================
# TOTAL: 52
# PASS:  49
# SKIP:  0
# XFAIL: 0
# FAIL:  3
# XPASS: 0
# ERROR: 0
============================================================================
```
```
FAIL: pagesegmode_test
FAIL: resultiterator_test
FAIL: textlineprojection_test
```

Shreeshrii:
https://github.com/tesseract-ocr/tesseract/runs/1706154138?check_suite_focus=true
Report from the dailyunittest run.
```
commit 6d165528a265be9cbe17e539288df27a558125a8
Author: Egor Pugin <egor.pugin@gmail.com>
Date:   Fri Jan 15 01:08:15 2021 +0300

    Merge pull request #3236 from stweil/max_iterations
    
    lstmtraining: Interpret negative value for --max_iterations as epochs
=============================================================
   tesseract 5.0.0-alpha-20201231: unittest/test-suite.log
=============================================================

# TOTAL: 52
# PASS:  48
# SKIP:  0
# XFAIL: 0
# FAIL:  4
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2
```
config options were
[...]

PR #3261, opened 2021-01-24
Ignore Unittests subtests for DISABLED_LEGACY_ENGINE

Shreeshrii:


Shreeshrii:
Fixes https://github.com/tesseract-ocr/tesseract/issues/3226

Shreeshrii:
@amitdo Are these changes ok or should I just exclude them in the makefile?

unittest/resultiterator_test.cc
```diff
@@ -274,6 +274,10 @@ class ResultIteratorTest : public testing::Test {
 
 // Tests that Tesseract gets exactly the right answer on phototest.
 TEST_F(ResultIteratorTest, EasyTest) {
+#ifdef DISABLED_LEGACY_ENGINE
```
amitdo, line 277: This test is a layout analysis test which does not really need osd.

It should work with lstm, if the oem and psm are adapted.

amitdo:
I written a comment but removed it because I was not sure about it.

Do what you see as appropriate.

I don't have time to review it right now, sorry.

Shreeshrii:
Ok. I will test with an older version to see if the tests were passing earlier.

amitdo:
The changes look fine. In the iterator test we probably can re-enable some of the disabled tests with some additional work, but it's a good start.

amitdo:
#3626

@Shreeshrii, do you remember if you tested the modified resultiterator_test.cc with --disable-legacy?

amitdo:
@Shreeshrii, do you remember if you tested the modified resultiterator_test.cc with --disable-legacy?

Shreeshrii:
No, I haven't tested recently. The CI for this was on a monthly basis. You can change it to daily similar to the other Actions.

amitdo:
>No, I haven't tested recently.

But did you tested it when you originally submitted the PR? 

I already change it to run every Monday.

PR #3262, opened 2021-01-24
Fix Memory leak in ligature_table_test

Shreeshrii:
Fixes leak identified in 
https://github.com/Shreeshrii/tesseract/runs/1756021945?check_suite_focus=true#step:12:59


stweil (approved):
It does not fix the leak, but suppresses the error message.

stweil:
Merged, thanks!

Shreeshrii:
You are right, @stweil. I should have phrased it as 'suppresses Memory leak in ligature_table_test'

PR #3275, opened 2021-02-02
Improve cmake Github Actions workflow

Shreeshrii:
macOS builds training tools now after https://github.com/tesseract-ocr/tesseract/commit/4902e6868288b739bbbe8d9b864b0b19fb97256d by @stweil 
Thanks!


Shreeshrii:
See https://github.com/Shreeshrii/tesseract/actions/runs/527393559 
for test run in fork

.github/workflows/cmake.yml
```diff
@@ -50,6 +69,8 @@ jobs:
       - name: Configure (macOS)
         shell: bash
         run: |
+             set -e
+             export PKG_CONFIG_PATH=/usr/local/opt/icu4c/lib/pkgconfig:/usr/local/opt/libarchive/lib/pkgconfig:/usr/local/opt/libffi/lib/pkgconfig:$PKG_CONFIG_PATH
              mkdir build
```
stweil, line 74: A more general solution would use `$(brew --prefix)` instead of `/usr/local`: Homebrew for macOS on Arm64 suggests a different prefix `/opt/homebrew`, so `/usr/local` would not work there.

PR #3071, opened 2020-07-31
URW Bookman and Century Schoolbook are no longer valid

mozbugbox:
The fonts are not proper rendered by text2image anymore.

PR #3280, opened 2021-02-04
Change LATIN_FONTS to use replacement fonts from TeX Gyre collection

Shreeshrii:
Changed

URW Bookman L -> Bonum

Century Schoolbook L -> Schola

as per suggestion by @amitdo in https://github.com/tesseract-ocr/tesseract/pull/3071#issuecomment-772792181


stweil (approved):
Thanks.

Issue #3289, opened 2021-02-09
libarchive dependency wrong?


bertsky:
If I build Tesseract with autotools, it drags in `libarchive` (`libarchive13` in my case), for the shared library as well as the CLI – in contrast to the prebuilt version.

But this fails for tesstrain-generated models:

```
archive_read_open_filename(...,/usr/local/share/tessdata/GT4HistOCR_2000000.traineddata,...) failed, Invalid or incomplete multibyte or wide character
```

So I tried to remove that dependency when calling `configure`, but none of the documented methods work, it always ends up giving me `checking for libarchive... yes`. So far, I have tried:
- `./configure --without-libarchive ...`
- `./configure --disable-libarchive ...`
- `libarchive_CFLAGS= libarchive_LIBS= ./configure ...`

What is the reason for the above failure, and is there any known workaround at build time?

bertsky:
> But this fails for tesstrain-generated models:

Correction: it fails for all models.

amitdo:
There no way to disable it currently.

Same for issue with libcurl.

stweil:
Yes, there are currently no options to disable `libcurl` or `libarchive`.

The message listed above should only be shown for debug builds and is only an information. When opening a model file, Tesseract first tries whether it is a supported archive file (for example a zip or a compressed tar file). If that fails, it uses Tesseract's own proprietary format to access the components.

If the message is also shown for builds which were not configured with `--enable-debug`, that's an error.

In any case the message would disappear with model files using a standard archive format.

stweil:
We could offer model files in one of the new formats. That would make the handling easier (for example adding or removing the dictionary) because it no longer requires `combine_tessdata`. The model files would also be much smaller.

But it would break compatibility with existing Tesseract binaries without libarchive support. And git cannot compress deltas for compressed files, so managing such model files with git would create large git repositories.

The new formats can safely be used by Linux distributions, for communities like OCR-D with fixed guidelines or for local use.

bertsky:
Oh, I see. Thanks very much @stweil @amitdo for responding so quickly and explaining it!

Yes, it was a debug build in my case. Sorry, I should have checked myself whether the error actually does any harm.

IMHO backwards compatibility is extremely valuable in such cases...

amitdo:
IMO, since we don't have an official tessdata repo with the new format, by default tesseract should not link with libarchive (even when it found on the system).

There should be a '--with-libarchive' option and 'with-libcurl'.

[...]

PR #3292, opened 2021-02-11
Add new configure options --disable-doc, --with-archive and --with-curl

stweil:
The new options allow overriding the default behaviour which checks whether the documentation can be build and whether libarchive and libcurl are available.


stweil:
See discussion for issue #3289.

PR #3297, opened 2021-02-12
Add basicapitest to daily Github Actions Workflow

Shreeshrii:


.github/workflows/autotools.yml
```diff
@@ -60,41 +83,69 @@ jobs:
...
+           tesseract test/testing/raaj.tif - -l hin --oem 1   --tessdata-dir ../tessdata
+           tesseract test/testing/viet.tif - -l vie --oem 1   --tessdata-dir ../tessdata
+           tesseract test/testing/hebrew.png - -l heb --oem 1   --tessdata-dir ../tessdata
+           tesseract test/testing/eurotext.tif - -l fra --oem 1 --tessdata-dir ../tessdata_best
+           tesseract test/testing/arabic.tif - -l ara --oem 1 --psm 6  --tessdata-dir ../tessdata
+
+    - name: Run Tesseract basicapitest
+      run: |
+           export CXXFLAGS="-std=c++11"
+           cd basicapitest
```
stweil, line 105: This directory does not exist. Nor does basicapitest.cpp.

Shreeshrii: It is in my fork of tesseract. It is a copy of the basic api example from tessdoc.

Actually I wanted to check with you whether the cpp file should be uploaded to the test repo in testing directory and used from there rather than placed in a new directory.

Shreeshrii: Please see if this is ok
https://github.com/tesseract-ocr/test/pull/20

I will change the CI workflow accordingly.

PR #3305, opened 2021-02-20
Actions CI: Add workflow to build tesseract using vcpkg and cmake

Shreeshrii:
- no training tools
- also builds and runs basicapitest using cmake


amitdo:
@Shreeshrii,

Maybe you can also add MingW-w64 to CI?

PR #3309, opened 2021-02-23
Replace more STRING and GenericVector variables and other code maintenance

stweil:


src/ccutil/params.h
```diff
@@ -220,11 +219,11 @@ class StringParam : public Param {
   void set_value(const STRING& value) { value_ = value; }
   void ResetToDefault() { value_ = default_; }
   void ResetFrom(const ParamsVectors* vec) {
-    for (int i = 0; i < vec->string_params.size(); ++i) {
-      if (strcmp(vec->string_params[i]->name_str(), name_) == 0) {
+    for (auto* param : vec->string_params) {
+      if (strcmp(param->name_str(), name_) == 0) {
```
egorpugin, line 223: We should use std::string/str::string_view **later**.

egorpugin (approved):

PR #3313, opened 2021-02-27
Add new checks for floating point errors and fix a division by zero

stweil:
`lstmtraining --debug_float` now detects float errors like division by zero or overflow on systems with the GNU C library.

Such a division by zero occurs when running a training because of a `momentum == -1`.

That is fixed here by skipping the adaption of the learning rate for non-positive momentum values.

The float checks are unconditionally enabled by the new code for `tesseract` to see whether they occur there, too.

### DIFF
--- a/src/api/tesseractmain.cpp
+++ b/src/api/tesseractmain.cpp
@@ -22,6 +22,9 @@
 #endif
 
 #include <cerrno>               // for errno
+#if defined(__USE_GNU)
+#include <cfenv>                // for feenableexcept
+#endif
 #include <iostream>
 
 #include <allheaders.h>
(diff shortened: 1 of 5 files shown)


amitdo:
>The new code requires the function feenableexcept which is only
available with the GNU libc, so it is only used on Linux.

https://www.freebsd.org/cgi/man.cgi?query=feenableexcept&sektion=3&manpath=FreeBSD%209.2-RELEASE

https://man.openbsd.org/feenableexcept.3

https://android.googlesource.com/platform/bionic/+/donut-release2/libm/man/feenableexcept.3

stweil:
If desired and possible that code can be enabled for more platforms.

Raising floating point exceptions is also possible for Windows, but requires different code: https://stackoverflow.com/questions/14990565/catch-nan-and-inf-in-windows-c.

Enabling floating point exceptions does not add a new user feature to the Tesseract code, but helps finding problems, so it is useful for debugging. I expect that most debugging is done on Linux hosts, so the current solution might be sufficient.

PR #3315, opened 2021-03-01
Fix lstm_recode_test for big endian machines

stweil:


src/ccutil/serialis.cpp
```diff
@@ -73,6 +73,20 @@ TFile::~TFile() {
...
 
+bool TFile::DeSerializeSize(int32_t* pSize) {
+  uint32_t size;
+  if (FReadEndian(&size, sizeof(size), 1) != 1) {
+    return false;
+  }
+  if (size > data_->size() / 4) {
+    // Reverse endianness.
+    swap_ = !swap_;
+    ReverseN(&size, 4);
```
egorpugin, line 84: I do not understand a bit. Isn't FReadEndian reversed bytes already?

stweil: It reverses if `swap_ == true`. Initially swapping is disabled when reading an `lstmf` file. Then the number of entries is read. On big endian machines that value is much too large for `lstmf` files which were written on little endian machines.

Here is a heuristic check which simply expects that the number of entries must be less than the number of int values which fit into the size of the `TFile`. If that is not the case, the check assumes that `swap_` has the wrong value and swaps the size.

egorpugin: Ok, lstm files are written in which endianness?
If they are always l.e., we can pass __BIG_ENDIAN__ def right into swap argument.
If they are depends on host arch, not sure if this is right decision.

stweil: They are currently written in host endianness, like other binary files written by Tesseract code. So they can have any endianness. The files which are provided for the unit tests were written on little endian machines. That's why they failed on big endian machines.

For the future I'd prefer writing all new files in little endian order, no matter which endianness the host uses. That requires some smaller updates of the `Serialize` methods.

egorpugin: > For the future I'd prefer writing all new files in little endian order, no matter which endianness the host uses.

+1

egorpugin: Maybe 1) just declare that since 5.0. 2) Change to little endian always on write and 3) read with `__BIG_ENDIAN__` def in swap argument.

egorpugin:
Merge is blocked by the conflict.

some background:
I made serialis.h changes earlier. I send some my libtess program to other pc and it is crashes on reading eng trained data in `.Init()`. I thought that it has something with reading, like the issue I had with wrong locales. But no, the issue is still there, I'm investigating.
Also I see frequent crashes on lstm+default mode even on my machine.)

stweil:
> Merge is blocked by the conflict.

Okay, I can solve that conflict locally and commit the change to tesseract-ocr then.

PR #3322, opened 2021-03-11
master branch: missing changes that are needed too

GerHobbelt:
to make this codebase compile on Windows.

Just a minor glitch.

---

Your latest changes around `std::string` parameters missed this bit, which popped up as a compile error in my own (custom) tesseract build for Windows.

### DIFF
--- a/src/wordrec/wordrec.h
+++ b/src/wordrec/wordrec.h
@@ -49,7 +49,7 @@ class TESS_API Wordrec : public Classify {
   virtual ~Wordrec() = default;
 
   // tface.cpp
-  void program_editup(const char *textbase, TessdataManager *init_classifier,
+  void program_editup(const std::string &textbase, TessdataManager *init_classifier,
                       TessdataManager *init_dict);
   void program_editdown(int32_t elasped_time);
   int end_recog();
--- a/src/wordrec/wordrec.h
+++ b/src/wordrec/wordrec.h
@@ -49,7 +49,7 @@ class TESS_API Wordrec : public Classify {
   virtual ~Wordrec() = default;
 
   // tface.cpp
-  void program_editup(const std::string &textbase, TessdataManager *init_classifier,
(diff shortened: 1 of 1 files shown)


src/wordrec/wordrec.h
```diff
@@ -49,7 +49,7 @@ class TESS_API Wordrec : public Classify {
   virtual ~Wordrec() = default;
 
   // tface.cpp
-  void program_editup(const char *textbase, TessdataManager *init_classifier,
+  void program_editup(const std::string &textbase, TessdataManager *init_classifier,
```
stweil, line 52:
```suggestion
  void program_editup(const std::string& textbase, TessdataManager* init_classifier,
```

stweil:
Thank you!

amitdo:
@stweil,

You missed `TessdataManager *init_dict)`

stweil:
... and many more in Egor's changes. I am afraid we'll need a code formatter.

egorpugin:
Of course. I said some time ago that we need to call clang-format on our codebase. (And wanted to do it couple of days ago.)

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/master/.clang-format

Stefan and Egor, maybe you want make some changes to the formatting rules that you can agree on.

stweil:
Do we need changes for those rules?

egorpugin:
I don't care about style. I care about consistency.
I'll try to run clang-format today.

stweil:
I suggest to use latest clang-format-11. That would prevent differences from different versions of clang-format.

amitdo:
A further discussion about code style:

https://github.com/tesseract-ocr/tesseract/commit/afa476bc232257200982f09057a15234ea42a33b#commitcomment-48195505

amitdo:
>This will create some huge line changes which is not very good for git blame,

`git blame` has this optional argument:
`--ignore-revs-file <file> `

https://git-scm.com/docs/git-blame

PR #3323, opened 2021-03-12
Actions CI: Add vcpkg build for tesseract 4.1 (windows and linux)

Shreeshrii:


amitdo:
I hope we won't hit the maximum limits of GitHub Actions: 

https://docs.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

We are on the free account, I believe.

stweil:
We have currently a retention time of 90 days for artifacts and log files and could reduce that if we hit the storage limit. That would not help for the computation time limit of course. I did not find a report about our current usage, so we only can estimate it (or wait until we hit the limit).

Shreeshrii:
These are daily ci. Run once every night.
sw.yml runs for every commit.
unittest-disable-legacy is setup to run once a month, since it fails currently.

The cron frequency can be changed and the matrix combos can be reduced if we are hitting the limits.

I am not saving anything in cache so no storage is used.

Shreeshrii:
Workflow Name -  Approx.Time Taken

 autotools - 1hr 30 mins, daily
 cmake - 15 mins, daily
 sw - 40 mins every commit, 4+ hours for nightly run
 unittest - 1hr 35 mins, daily
 unittest-disablelegacy - 30 mins, once a month
 vcpkg - 10 mins, daily
 vcpkg-4.11 - 25 mins, daily

Shreeshrii:
Maybe the limits don't apply...

> GitHub Actions usage is free for both public repositories and self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage, depending on the product used with the account.

amitdo:
https://github.com/pricing

>**Free**

> Included 2,000 Actions minutes/month Free for public repositories 

Let's wait and see if we'll hit the time limit. Maybe they have exceptions to super popular projects like Tesseract.

Shreeshrii:
We want every open source project to be productive and use best practices, so Actions is free for the 40 million developers on GitHub to use with public repositories. For private repositories, Actions offers simple, pay-as-you-go pricing. 

https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/

amitdo:
'Free' does not imply 'without any limit'.

Nevertheless, It does seem that the 2000 minutes is for private repos.

So probably we don't need to worry about this.

PR #3325, opened 2021-03-13
Replace more STRING by std::string

stweil:
This also removes `STRING::add_str_int` and `STRING::add_str_double`.


egorpugin:
```TESS_API
const std::vector<std::string> split(const std::string &s, char c);```

amitdo:
Now that you have split() as a free function, I assume next step is to get rid of the split() **method**, right?

Are you planning to remove the STRING class entirely?

stweil:
Yes, that's my goal. https://github.com/stweil/tesseract/tree/string contains more removals.

PR #3329, opened 2021-03-15
Remove class STRING

stweil:


stweil:
There remains a single usage of `STRING`. Replacing that by `std::string` results in a memory leak which still has to be examined.

stweil:
> There remains a single usage of `STRING`

This is solved now. `STRING` is now completely removed.

egorpugin:
Great! Thank you.

amitdo:
Goodbye STRING.
:-)

Thanks!

PR #3332, opened 2021-03-16
Replace some GenericVector by std::vector

stweil:


src/ccutil/helpers.h
```diff
@@ -24,13 +24,19 @@
...
+#include <algorithm>  // for std::find
 #include <functional>
 #include <random>
 #include <string>
 #include <vector>
 
 namespace tesseract {
 
+template <class T>
+inline bool contains(const std::vector<T> &data, const T &value) {
```
egorpugin, line 36: no need to inline here

egorpugin (approved):

PR #3336, opened 2021-03-17
Replace more GenericVector by std::vector and remove GenericVector::choose_nth_item

stweil:


src/ccstruct/blobs.cpp
```diff
@@ -548,10 +548,10 @@ void TBLOB::GetEdgeCoords(const TBOX &box, std::vector<std::vector<int>> &x_coor
   y_coords.resize(box.width());
   CollectEdges(box, nullptr, nullptr, &x_coords, &y_coords);
   // Sort the output vectors.
-  for (int i = 0; i < x_coords.size(); ++i)
-    std::sort(x_coords[i].begin(), x_coords[i].end());
-  for (int i = 0; i < y_coords.size(); ++i)
-    std::sort(y_coords[i].begin(), y_coords[i].end());
+  for (auto coord : x_coords)
```
egorpugin, line 551: Copies are sorted, not original data. This is not the same as code before.

src/ccutil/unicharset.h
```diff
@@ -307,9 +307,9 @@ class TESS_API UNICHARSET {
 
   // Delete CHAR_FRAGMENTs stored in properties of unichars array.
   void delete_pointers_in_unichars() {
-    for (int i = 0; i < unichars.size(); ++i) {
-      delete unichars[i].properties.fragment;
-      unichars[i].properties.fragment = nullptr;
+    for (auto unichar : unichars) {
```
egorpugin, line 310: Also not a reference.

src/ccstruct/normalis.cpp
```diff
@@ -163,8 +162,8 @@ static void ComputeRunlengthImage(const TBOX &box,
   // Set a 2-d image array to the run lengths at each pixel.
   for (int ix = 0; ix < width; ++ix) {
     int y = 0;
-    for (int i = 0; i < y_coords[ix].size(); ++i) {
-      int y_edge = ClipToRange(y_coords[ix][i], 0, height);
+    for (auto y_coord : y_coords[ix]) {
```
egorpugin, line 165: If ClipToRange does not change first arg, copies may be expensive here (use `auto &`).

src/ccstruct/normalis.cpp
```diff
@@ -182,8 +181,8 @@ static void ComputeRunlengthImage(const TBOX &box,
   // Now set the image pixels the the MIN of the x and y runlengths.
   for (int iy = 0; iy < height; ++iy) {
     int x = 0;
-    for (int i = 0; i < x_coords[iy].size(); ++i) {
-      int x_edge = ClipToRange(x_coords[iy][i], 0, width);
+    for (auto x_coord : x_coords[iy]) {
```
egorpugin, line 184: same as prev

src/ccstruct/blobs.cpp
stweil: Right. I'll commit a fix.

src/ccstruct/normalis.cpp
stweil: `y_coord` is of type `int`, to this should be fine.

src/ccstruct/normalis.cpp
stweil: This is also an `int` value, so no need for a reference.

src/ccutil/unicharset.h
stweil: Right. I'll commit a fix.

PR #3337, opened 2021-03-18
Replace remaining choose_nth_item by std::nth_element

stweil:


src/textord/makerow.cpp
```diff
@@ -133,12 +133,7 @@ static int row_spacing_order( // sort function
   // converted ptr
   const TO_ROW *row2 = *reinterpret_cast<const TO_ROW *const *>(item2);
 
-  if (row1->spacing < row2->spacing)
-    return -1;
-  else if (row1->spacing > row2->spacing)
-    return 1;
-  else
-    return 0;
+  return row1->spacing < row2->spacing;
```
egorpugin, line 136: Can we remove reinterpret casts above? `void *` is so C-style.

egorpugin: Probably even inline lambda can be used.

stweil: I removed the type casts. lambda would be possible here and for other code, too. That can be done later.

PR #3341, opened 2021-03-18
Replace more GenericVector and PointerVector by std::vector

stweil:


stweil:
I am still looking for a better code pattern as a replacement for `PointerVector`.

PR #3344, opened 2021-03-19
Replace more GenericVector and PointerVector by std::vector

stweil:


egorpugin:
Regarding pointer vector and other pointer - I propose we'll look at them a bit later.
We need to setup basic guidlines for using automatic memory management.
Lots of apis accept raw pointers to own them (caller must not call delete) etc.

PR #3345, opened 2021-03-20
Replace more PointerVector by std::vector

stweil:


src/classify/intproto.cpp
```diff
@@ -688,9 +689,8 @@ INT_TEMPLATES Classify::ReadIntTemplates(TFile *fp) {
   /* variables for conversion from older inttemp formats */
   int b, bit_number, last_cp_bit_number, new_b, new_i, new_w;
   CLASS_ID class_id, max_class_id;
-  auto *IndexFor = new int16_t[MAX_NUM_CLASSES];
-  auto *ClassIdFor = new CLASS_ID[MAX_NUM_CLASSES];
-  auto **TempClassPruner = new CLASS_PRUNER_STRUCT *[MAX_NUM_CLASS_PRUNERS];
+  std::array<CLASS_ID, MAX_NUM_CLASSES> ClassIdFor;
+  std::array<CLASS_PRUNER_STRUCT *, MAX_NUM_CLASS_PRUNERS> TempClassPruner;
```
egorpugin, line 693: Use this with caution.
On recursive calls and big size of such arrays it may cause stack overflow.
This comes to mind because dynamic memory allocation probably was used intentionally.
We should double check this.

stweil: You are right, they are much too large. I now use `std::vector` which allocates dynamic memory.

PR #3350, opened 2021-03-21
Fix compiler warnings (mostly -Wsign-compare)

stweil:


stweil:
Those warnings are the result of replacing `STRING` and `GenericVector` (which used `int` for the size) by C++ `std::string` and standard containers (which use `size_t` for the size).

There still remain some more of those warnings.

PR #3351, opened 2021-03-21
Modernize code using clang-tidy

stweil:


egorpugin:
I'll wait for all greens here.

stweil:
@egorpugin, is it possible to build the unittests with cmake?

stweil:
> I'll wait for all greens here.

💚

egorpugin:
> is it possible to build the unittests with cmake?

It is possible, but not implemented.

The problem is in dependency management. I used cppan/sw for this.
Easy way is to try the same approach. (I'm using mostly windows.)
On linux deps could be plugged in faster.
See my sw code for tests https://github.com/tesseract-ocr/tesseract/blob/master/sw.cpp#L222

From https://www.kdab.com/clang-tidy-part-1-modernize-source-code-using-c11c14/
I see that clang-tidy uses `compile_commands.json` (`Generating a compile_commands.json` file section).
`cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..`

Sw can do the same:
`sw generate -Dwith-tests=1 -g compdb`
The result is in `.sw\g\compdb`, but you need to extract only unittests from there because it contains whole build.

zdenop:
IMHO cmake dependencies are problem for training only. AFAIK leptonica dependencies are "cmake ready" (or there is alternative solution e.g. gif).

PR #3354, opened 2021-03-22
Add braces to single line statements and modernize unittest code using clang-tidy

stweil:


egorpugin:
What is the purpose of braces to single line statements?
I'm asking this for myself. I'm trying to setup this code style, but sometimes it is easy for me to write just
```
if (x)
  expr;
```

stweil:
For me uniform looking code is easier to read. It also allows simple copying of for example an `if` statement no matter how many statements are in the conditional code. And the code in the block can be easily expanded to more than a single line (or multi-line blocks reduced to a single line) without adding (or removing) braces.

We can run clang-tidy from time to time and fix single line statements without braces, so such statements would not prevent new code from being merged.

stweil:
PS. I used `./configure CXX="bear --append -- clang++"` to create the `compile_commands.json` which is required for `run-clang-tidy`.

egorpugin:
> We can run clang-tidy from time to time

Definitely.
And clang-format too.
And also it is possible to add automatic checks to PRs that won't pass on bad formatting. I'm not sure about this. I think I'd prefer more relaxed PR style. We can run format later or refactor some things after merging.

PR #3355, opened 2021-03-22
Print command line options if run_command() failed.

eighttails:
It helps to reproduce failed command execution during training.

### DIFF
--- a/src/training/tesstrain_utils.sh
+++ b/src/training/tesstrain_utils.sh
@@ -79,7 +79,7 @@ run_command() {
     shift
     tlog "[$(date)] ${cmd} $@"
     if ! "${cmd}" "$@" 2>&1 | tee -a "${LOG_FILE}"; then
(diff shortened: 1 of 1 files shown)


stweil:
@eighttails, please consider using `tesstrain.py` (part of the tesstrain repository) instead of the old `tesstrain.sh`.

@Shreeshrii, do you think that we can remove the old shell scripts for training?

Shreeshrii:
>do you think that we can remove the old shell scripts for training?

Since 5.0 is quite different from 4.0.x, I think it is fine to remove them. 

I suggest you do it along with a new tag/release so that it is easy to identify.  

It will also require changes in documentation.  We should probably move Ray's tesstutorial to be as applicable for Tesseract 4.0.x and have new instructions for training/tutorial for 5.0.

Shreeshrii:
I have made some changes in tessdoc, moved training related info to tess4 subdirectory. Will Add new pages for tess5.

PR #3358, opened 2021-03-22
Avoid lots of messages for training with single line images

stweil:

### DIFF
--- a/src/ccstruct/imagedata.cpp
+++ b/src/ccstruct/imagedata.cpp
@@ -576,7 +576,8 @@ bool DocumentData::ReCachePages() {
       delete page;
     }
     pages_.clear();
-  } else {
+  } else if (loaded_pages > 1) {
+    // Avoid lots of messages for training with single line images.
     tprintf("Loaded %zu/%d lines (%d-%zu) of document %s\n", pages_.size(), loaded_pages,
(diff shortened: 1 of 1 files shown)


stweil:
I hope that nobody will miss those messages.

egorpugin:
In theory such messages should be in trace verbosity.
One log level is definitely not enogh.

Shreeshrii:
@stweil I see that the `Loaded` messages are no longer there. Thanks, it will certainly make the log files smaller.

If possible, as suggested by @egorpugin they can be kept for trace verbosity. It will be helpful when trying to track an error.

```
Loaded 1/1 lines (1-1) of document data/San-ground-truth/bugfix/san.Segoe_UI_Semi-Light.0000002.exp0_0.lstmf
Loaded 1/1 lines (1-1) of document data/San-ground-truth/bugfix/san.Mandala_Italic.0000080.exp0_0.lstmf
Loaded 1/1 lines (1-1) of document data/San-ground-truth/bugfix/san.Segoe_UI_Semi-Light_Italic.0000040.exp0_0
```

stweil:
All warning and error messages report the filename. Example:

    Compute CTC targets failed for data/gt/AustrianNewspapers/train/ONB_ibn_19110701_010/ONB_ibn_19110701_010.tif_line_1547100903812_34.lstmf!

So the removed message should not be needed for error tracking.

Issue #3083, opened 2020-08-24
RFC: allow flexible or better binarization


bertsky:
Tesseract has always included its own, internal binarization – which is **not** based on Leptonica and is of rather bad quality (custom global Otsu implementation without normalization). Leptonica does have lots of nice adaptive local normalization and thresholding implementations, but they are not utilized.

Since Tesseract 4.0, _recognition_ does not (under normal circumstances) use that binarized image, but uses the greyscale-converted raw image. If the input image was already bitonal, then it still works. However, what the LSTM model expects depends on what data it was _trained_ on. (Original tessdata models were all trained on artificial/clean greyscale IIRC.)

But the binarized image is still needed for all _segmentation_ and layout analysis (OSD, separator detection, picture detection). And to make matters worse, binarization **also** makes use of the greyscale image and threshold values from the binarizer – in order to get a better approximation/interpolation of blob outlines (`ComputeEdgeOffsets`). If the input image was already bitonal, then a fallback method is used which is not as accurate (`ComputeBinaryOffsets`).

Now the user (both CLI and API) is in a **dilemma**:
1. Present the originial/raw image:
   - Segmentation on some images may be suboptimal, because good binarization is hard.
   - Recognition might expect a (colour/contrast-) normalized or even binarized (e.g. from tesstrain) image and thus be suboptimal.
2. Present an externally binarized image: 
   - Segmentation on other images may be suboptimal, because the blob outlines are inaccurate.
   - Recognition might expect a greyscale image and thus be suboptimal.

So what do we do? Allow delegating to Leptonica's `pixContrastNorm`, `pixSauvolaBinarizeTiled` etc. methods via parameter variable? Or extend the API to allow passing the threshold values of an external binarization? How do we inform/document this and encapsulate an LSTM model's expectations?

There are many more aspects, but I just wanted to open the discussion.

If requested, I can provide example images (of bad segmentation due to internal Otsu or bad segmentation due to inaccurate interpolation from bitonal input; of recognition perplexed by background masked to white), as well as pointers to code.

wrznr:
@bertsky Many thanks for bringing up this important issue. It might indeed be helpful to illustrate the matter with sample images (in order to get the wave of comments rolling).

amitdo:
Thank for opening this RFC. Very nice analysis.

I will add my comments later.

amitdo:
[...]

PR #3359, opened 2021-03-22
Modernize code for OTSU and reduce public API further

stweil:
Remove thresholder.h from the public API.

### DIFF
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -336,15 +336,6 @@ public:
   void SetRectangle(int left, int top, int width, int height);
 
   /**
-   * In extreme cases only, usually with a subclass of Thresholder, it
-   * is possible to provide a different Thresholder. The Thresholder may
-   * be preloaded with an image, settings etc, or they may be set after.
-   * Note that Tesseract takes ownership of the Thresholder and will
-   * delete it when it it is replaced or the API is destructed.
-   */
-  void SetThresholder(ImageThresholder *thresholder);
-
-  /**
    * Get a copy of the internal thresholded image from Tesseract.
    * Caller takes ownership of the Pix and must pixDestroy it.
    * May be called any time after SetImage, or after TesseractRect.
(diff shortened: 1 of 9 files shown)


egorpugin:
Do we put it in 5.0?

stweil:
> Do we put this into 5.0?

If nobody objects, yes.

include/tesseract/baseapi.h
```diff
@@ -342,11 +342,7 @@ class TESS_API TessBaseAPI {
    * Note that Tesseract takes ownership of the Thresholder and will
    * delete it when it it is replaced or the API is destructed.
    */
-  void SetThresholder(ImageThresholder *thresholder) {
-    delete thresholder_;
-    thresholder_ = thresholder;
-    ClearResults();
-  }
+  void SetThresholder(ImageThresholder *thresholder);
```
egorpugin, line 345: This can't be a public API anymore, since user does not know how to inherit from ImageThresholder.

egorpugin: So should be removed in this case.

stweil: Right, so that should be removed, too, if it is not used anywhere.

stweil: I added a commit now which removes `TessBaseAPI::SetThresholder`, too.

stweil:
> If nobody objects, yes.

I merged the API change now. Let's see whether someone misses the removed parts. I don't think so, but if I was wrong, we can review that decision again.

egorpugin:
The image can be passed already thresholded.
Even preprocessed in more ways than tess could do itself.

amitdo:
IMO, you  removed a useful feature.

Unless you intend to replace it with something better. I don't think it should be removed.

Related issue: #3083

stweil:
I looked at that code because I want to allow better binarization and had the impression that it would not help.

amitdo:
Okay.

PR #3367, opened 2021-03-29
Replace most malloc / realloc / free by new / delete / std::vector

stweil:


egorpugin:
separate note
There are many static arrays over the codebase.
I think we can try to replace them with `std::array`. This might help in iterating over them.
Like
```
ADAPTED_CONFIG Config[MAX_NUM_CONFIGS];
->
std::array<ADAPTED_CONFIG, MAX_NUM_CONFIGS> Config;
```

amitdo:
https://clang.llvm.org/extra/clang-tidy/checks/modernize-avoid-c-arrays.html

stweil:
`clang-tidy` reports more than 600 warnings for code which can be updated to use `std::array`, but those updates have to be made manually.

egorpugin:
clang-tidy can't do that or it gives a lot of misuses and compilation errors?

stweil:
Citing the documentation linked above: "However, fix-it are potentially dangerous in header files and are therefore not emitted right now."

clang-tidy only warns, but does not change the code.

PR #3371, opened 2021-03-29
lstm: Include missing header

nickjwhite:
lstm/recodebeam.h was using PointerVector from ccutil/genericvector.h without including it.


egorpugin:
There is no any GenericVector usage in this file.

stweil:
> There is no any GenericVector usage in this file.

... but it still uses `PointerVector` which is also defined in `genericvector.h`.

egorpugin:
Oops, sorry. :)

Issue #3369, opened 2021-03-29
Endless loop


wix-andriusb:
### Environment
tesseract 4.1.1

```bash
uname -a
Darwin VL-C02WL1AYHTD6 19.6.0 Darwin Kernel Version 19.6.0: Tue Nov 10 00:10:30 PST 2020; root:xnu-6153.141.10~1/RELEASE_X86_64 x86_64
Linux ocr-5b7bf86f6-f6qsd 5.4.65-wix #1 SMP Thu Nov 19 15:24:12 UTC 2020 x86_64 GNU/Linux
```

### Current Behavior:

running tesseract in command line on this image https://bentkus.eu/ocr_while_true.gif does not finish after 1h
```
tesseract ocr_while_true.gif ocr_while_true --dpi 150
```
### Expected Behavior:
process should finish after in 2 minutes

### Suggested Fix:
I'll try to build and see why it never stops

stweil:
That GIF file is special: it includes 125 images.

amitdo:
This is a gif animation.

Convert it to static images and give them to 
tesserct as input.

stweil:
Other issues where OCR never finishes: #2196, #2288.

stweil:
> Convert it to static images [...]

The static images work fine. Nevertheless handling of animated GIF images has to be well defined, see my question above.

wix-andriusb:
Ok, thanks for the advice, I should handle this on my side, check for gif and slice and analyze it.
I tried other gifs and saw it finishing so I assumed that this should work too on this gif.

wix-andriusb:
Can you tell me maybe what tool you used to create the static images?

egorpugin:
> How should Tesseract handle animated GIF images? Create OCR for all images, or only for the first one, or refuse to process such files?

My answer is 'Create OCR for all images'

stweil:
> Can you maybe tell me what tool you used to create the static images?

I used `convert FROM.gif TO.png`.

wix-andriusb:
> My answer is 'Create OCR for all images'

Could be configurable through arguments with the default being do OCR for all images

egorpugin:
It's should be something like `cat 1.txt 2.txt 3.txt ...`
When we pass multiple images, they all must be processed. Same for multipage images (.gif, .tiff) if such format is enabled in leptonica.

stweil:
[...]

PR #3380, opened 2021-04-06
Optimize Gridsearch and proprietary list classes

stweil:


stweil:
This reduces the time for the test case in issue #3369 by a few percent:

261 s new code
269 s previous Tesseract 5 code
288 s Tesseract 4 code

PR #3382, opened 2021-04-06
Actions CI: Build Tesseract on Windows using cmake

Shreeshrii:
Thanks @zdenop


Shreeshrii:
See successful run at  https://github.com/Shreeshrii/tesseract/runs/2280456781?check_suite_focus=true

There seem to be a large number of warnings displayed during tesseract build. Are they expected? Should they be suppressed for CI?

egorpugin:
We have appveor for this?

PR #3387, opened 2021-04-08
lstmrecognizer.cpp: Call OutputStats() only when 'invert' is true

amitdo:


src/lstm/lstmrecognizer.cpp
```diff
@@ -343,33 +343,36 @@ bool LSTMRecognizer::RecognizeLine(const ImageData &image_data, bool invert, boo
...
-      }
-      *outputs = inv_outputs;
-      *inputs = inv_inputs;
-    } else if (re_invert) {
-      // Inverting was not an improvement, so undo and run again, so the
-      // outputs match the best forward result.
+  if (invert) {
+    float pos_min, pos_mean, pos_sd;
+    OutputStats(*outputs, &pos_min, &pos_mean, &pos_sd);
+    if (pos_mean < 0.5) {
```
stweil, line 349:
```suggestion
    if (pos_mean < 0.5f) {
```

stweil (approved):
Thanks.

Issue #2486, opened 2019-06-08
Issue 13611: tesseract-ocr/fuzzer-api: Direct-leak in C_BLOB::ConstructBlobsFromOutlines


stweil:
See [details](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13611). The bug is reproducible but I could not find a solution up to now.

stweil:
The bug is caused by a wrong implementation of `ELIST_ITERATOR`. Code which tries to extract all list elements like [line 190](https://github.com/tesseract-ocr/tesseract/blob/master/src/textord/tordmain.cpp#L190) and [line 202](https://github.com/tesseract-ocr/tesseract/blob/master/src/textord/tordmain.cpp#L202) in `tordmain.cpp` fail to extract the last element which gets lost.

stweil:
OSS-Fuzz closed issue 13611 because of changes in Tesseract code and opened a new issue 32246,  but both are identical.

stweil:
OSS-Fuzz reports both issues as fixed:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=13611
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=32246

PR #3392, opened 2021-04-10
Fix memory leak (OSS-Fuzz issue #32246)

stweil:


stweil:
This fixes issue #2486.

PR #3393, opened 2021-04-11
Fix division by zero during CJK training.

eighttails:
I met floating point error during training Japanese model.
Here is gdb output.
```
Program received signal SIGFPE, Arithmetic exception.
0x00007fcca7c14f5c in tesseract::LocalCorrelation::EstimateYFor (this=0x7ffdeae919d0, x=45, r=0.100000001)
    at src/textord/cjkpitch.cpp:165
165         return rc / vote;
```
This PR fixes the issue.

### DIFF
--- a/src/textord/cjkpitch.cpp
+++ b/src/textord/cjkpitch.cpp
@@ -158,10 +158,8 @@ public:
     float rc = 0;
     int vote = 0;
     for (int i = start; i < end; i++) {
-      if (values_[i].x != 0.0f) {
-        rc += values_[i].vote * x * values_[i].y / values_[i].x;
-        vote += values_[i].vote;
-      }
+      rc += values_[i].vote * x * values_[i].y / values_[i].x;
+      vote += values_[i].vote;
     }
 
     return vote == 0 ? 0.0f : rc / vote;
--- a/src/textord/cjkpitch.cpp
+++ b/src/textord/cjkpitch.cpp
@@ -158,11 +158,13 @@ public:
     float rc = 0;
     int vote = 0;
     for (int i = start; i < end; i++) {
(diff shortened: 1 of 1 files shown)


src/textord/cjkpitch.cpp
```diff
@@ -158,11 +158,13 @@ class LocalCorrelation {
     float rc = 0;
     int vote = 0;
     for (int i = start; i < end; i++) {
-      rc += values_[i].vote * x * values_[i].y / values_[i].x;
-      vote += values_[i].vote;
+      if (values_[i].x != 0.0f) {
```
egorpugin, line 161: What about this part? Does it give inf/nans?

src/textord/cjkpitch.cpp
```diff
@@ -158,11 +158,13 @@ class LocalCorrelation {
...
+        rc += values_[i].vote * x * values_[i].y / values_[i].x;
```
stweil, line 162: @eighttails, did you get a division by zero here, too? Or did it only happen in the return statement?

eighttails: No. I observed error only in the return statement.
This part is to avoid potential error.
I'm not sure if values_[i].x never be zero.

egorpugin: I think it would be better to fix only return value right now.

egorpugin:
Note: we need to squash such PRs on merging.

amitdo:
>Note: we need to squash such PRs on merging.

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits

PR #3395, opened 2021-04-11
Update AUTHORS

amitdo:


egorpugin:
@stweil @zdenop 

Is it very necessary to keep AUTHORS list?

stweil:
Are there any reasons not to keep it?

egorpugin:
We have git log.
It is not needed to keep such burden of maintaining every contributor.
The list could be very large.

Projects have MAINTAINERS list which is more functional (linux, llv, gcc).

PR #2784, opened 2019-11-26
Adding --print-fonts-table parameter & tessedit_font_id configuration option

Lucas-C:
The ability to "force select" the font has been requested a few times in the past, for example there on StackOverflow : https://stackoverflow.com/questions/13154150/explicitly-set-the-font-to-be-used-for-recognition-by-tesseract-ocr?rq=1

The new `--print-fonts-table` argument is helpful to select the value to set to the new `tessedit_font_id` configuration option.

PR #3406, opened 2021-04-29
Adding --print-fonts-table parameter & tessedit_font_id configuration option

Lucas-C:
Follow-up of https://github.com/tesseract-ocr/tesseract/pull/2784 after rebasing the branch

### DIFF
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -151,6 +151,11 @@ public:
   const char *GetStringVariable(const char *name) const;
 
   /**
+   * Print Tesseract fonts table to the given file.
+   */
+  void PrintFontsTable(FILE* fp) const;
+
+  /**
    * Print Tesseract parameters to the given file.
    */
   void PrintVariables(FILE *fp) const;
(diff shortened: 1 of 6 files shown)


amitdo:
What will happen when this is used with a model that does not have a data for the legacy engine?

I seems it will print an empty file.

Lucas-C:
The new CLI arguments are only available when `DISABLED_LEGACY_ENGINE` is not set.

Moreover no file is generated, only extra information on stdout

amitdo:
> The new CLI arguments are only available when `DISABLED_LEGACY_ENGINE` is not set.

I'm taking about the trained models. Models from the tessdata_best and tessdata_fast were only trained with the LSTM engine. They won't have font info.
 
> Moreover no file is generated, only extra information on stdout

So, I guess that "nothing" (empty string) will be printed with this option is used models that does not have font info.

Ideally, a message will be printed to stderr:
"This model does not have font information."

Lucas-C:
Feel free to submit a PR adding such message I guess.
The maintainers will review it

amitdo:
That was just a comment for possible future enhancement.

Anyway, thanks for your contribution.

PR #3408, opened 2021-04-29
fix compile error: PrintFontsTable() is for legacy builds only

GerHobbelt:
When compiling tesseract master branch with legacy engine DISABLED, these bits of code would throw a tantrum as they are for legacy only.

(Cherrypicked commit back onto mainline master.)

\# Conflicts:
\#	googletest


GerHobbelt:
Wow! That was quick! 😄 👍

PR #3410, opened 2021-04-30
[doc][fix] typo libicu-devel

M3ssman:
Just a small typo, it prints `libicu-devel` what should be `libicu-dev`


zdenop:
thanks

zdenop:
This is true, but for all other packages configure.ac uses `-dev`, so in that context `-devel` could be understood as typo.

Issue #3377, opened 2021-04-03
Tesseract seemingly stuck


MerlijnWajer:
### Environment

* **Tesseract Version**: Latest `master`
* **Commit Number**:  (`23ed59bd7bca777e4e104c4ee540843373aa9869`
* **Platform**: `Linux gentoo-x13 5.11.7-gentoo-dist #1 SMP Wed Mar 17 21:03:41 -00 2021 x86_64 AMD Ryzen 7 PRO 4750U with Radeon Graphics AuthenticAMD GNU/Linux`

### Current Behavior:

Tesseract hangs, seemingly never finishes

### Expected Behavior:

Tesseract doesn't hang and produces output normally

GDB backtrace (interrupted after more than 5 minutes):

```
merlijn@gentoo-x13 ~/archive/tesseract-src/tesseract $ time TESSDATA_PREFIX=/usr/share/tessdata LD_LIBRARY_PATH=`pwd` LD_LIBRARY_PATH=$LD_LIBARY_PATH:`pwd`/.libs gdb --args ./.libs/tesseract /tmp/sim_new-york-times_1900-01-11_49_15-603_0008.ppm - hocr
GNU gdb (Gentoo 10.1 vanilla) 10.1
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://bugs.gentoo.org/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./.libs/tesseract...
(gdb) r
Starting program: /home/merlijn/archive/tesseract-src/tesseract/.libs/tesseract /tmp/sim_new-york-times_1900-01-11_49_15-603_0008.ppm - hocr
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
[...]

Issue #1990, opened 2018-10-14
Tesseract thresholding eliminates text on bright background colors


jbarlow83:
### Environment

* **Tesseract Version**: 4.0.0-rc1-8-g1bee homebrew build
* **Platform**: macOS High Sierra

### Current Behavior:

When tesseract is run on the attached image, the text on highlight backgrounds is missing from output.

![highlight](https://user-images.githubusercontent.com/1825843/46923056-40f46480-cfc7-11e8-921e-a0ecfdf023dc.png)

```
tesseract -c tessedit_write_images=true  _.png stdout
Not highlighted text
```

The thresholder blacks out the text (this is tessinput.tif):

![tessinput](https://user-images.githubusercontent.com/1825843/46923075-76994d80-cfc7-11e8-903e-68a27814edb3.png)

### Expected Behavior:

Thresholder should treat highlights as background so that Tesseract recognizes all of the text.

amitdo:
I wonder why tesseract use its own otsu binarization and not leptonica's one.

You might want to test this image with leptonica's binarization options.

It also possible that your issue is cause by some filter done before or after the binarization.

amitdo:
I used gimp for image thresholding and gave the binary image to tesseract.

Output:
```
Not highlighted text

Highlighted text

 

 

Highlighted background
```

amitdo:
related: https://github.com/tesseract-ocr/tesseract/issues/242#issuecomment-191994220

amitdo:
Here's Tesseract's otsu:

https://github.com/tesseract-ocr/tesseract/blob/57970443b42b/src/ccstruct/otsuthr.cpp

zhangpy:
@amitdo Could you show me how to use gimp to change image thresholding? thanks

amitdo:
Gimp is a GUI tool. You might want to try imagemagick which is command line tool instead.

amitdo:
CC: @jbreiden

Shreeshrii:
>I wonder why does tesseract use its own otsu binarization and not leptonica's one.

https://github.com/DanBloomberg/leptonica/blob/master/src/binarize.c

```
 *  ===================================================================
 *  Image binarization algorithms are found in:
 *    grayquant.c:   standard, simple, general grayscale quantization
 *    adaptmap.c:    local adaptive; mostly gray-to-gray in preparation
 *                   for binarization
 *    binarize.c:    special binarization methods, locally adaptive and
 *                   global.
[...]

PR #3418, opened 2021-05-07
Add more binarization options

amitdo:
Use functions from Leptonica to provide more binarization options. The new options are: 1) Adaptive Otsu and 2) Sauvola (Tiled).

### DIFF
--- a/include/tesseract/publictypes.h
+++ b/include/tesseract/publictypes.h
@@ -278,6 +278,17 @@ enum OcrEngineMode {
   OEM_COUNT                    // Number of OEMs
 };
 
+/** 
+ * Except when OTSU_TRESH is chosen
+ * Leptonica is used for thresholding 
+ */
+enum ThreshMethod {
+  OTSU_TRESH,          // Legacy Tesseract's Otsu thresholding
+  ADAPTIVE_OTSU_TRESH,
+  SAUVOLA_TILED_TRESH,
+  TRESH_METHODS_COUNT, // Number of Thresholding methods
+};
+
 } // namespace tesseract.
 
 #endif // TESSERACT_CCSTRUCT_PUBLICTYPES_H_
--- a/include/tesseract/publictypes.h
+++ b/include/tesseract/publictypes.h
@@ -279,14 +279,14 @@ enum OcrEngineMode {
 };
 
 /** 
- * Except when OTSU_TRESH is chosen
+ * Except when Otsu is chosen
  * Leptonica is used for thresholding 
  */
-enum ThreshMethod {
-  OTSU_TRESH,          // Legacy Tesseract's Otsu thresholding
-  ADAPTIVE_OTSU_TRESH,
-  SAUVOLA_TILED_TRESH,
-  TRESH_METHODS_COUNT, // Number of Thresholding methods
+enum class ThreshMethod {
+  Otsu,         // Legacy Tesseract's Otsu thresholding
+  AdaptiveOtsu,
+  TiledSauvola,
+  Count,        // Number of Thresholding methods
 };
 
 } // namespace tesseract.
--- a/src/ccmain/thresholder.cpp
+++ b/src/ccmain/thresholder.cpp
@@ -20,6 +20,7 @@
 
 #include <cstdint> // for uint32_t
 #include <cstring>
+#include <tuple>
 
 #include "otsuthr.h"
 #include "thresholder.h"
@@ -182,6 +183,45 @@ void ImageThresholder::SetImage(const Image pix) {
   scale_ = 1;
   estimated_res_ = yres_ = pixGetYRes(pix_);
   Init();
+}
+
+std::tuple<bool, Image, Image, Image> ImageThresholder::Threshold(
+                                                         ThreshMethod method) {
+  Image pix_grey = nullptr;
+  Image pix_binary = nullptr;
+  Image pix_thresholds = nullptr;
+
+  if (image_width_ > INT16_MAX || image_height_ > INT16_MAX) {
+    tprintf("Image too large: (%d, %d)\n", image_width_, image_height_);
+    return std::make_tuple(false, nullptr, nullptr, nullptr);
+  }
+  
+  if (pix_channels_ == 0) {
+    // We have a binary image, but it still has to be copied, as this API
+    // allows the caller to modify the output.
+    Image original = GetPixRect();
+    pix_binary = original.copy();
+    original.destroy();
+    return std::make_tuple(false, nullptr, pix_binary, nullptr);
+  }
+
+  pix_grey = GetPixRectGrey();
+
+  if (method == OTSU_TRESH || method >= TRESH_METHODS_COUNT) {
+    method = ADAPTIVE_OTSU_TRESH;
+  }
+
+  int r;
+  if (method == ADAPTIVE_OTSU_TRESH) {
+    r = pixOtsuAdaptiveThreshold(pix_grey, 300, 300, 0, 0, 0.1, 
+                                 pix_thresholds.a(), pix_binary.a());
+  } else if (method == SAUVOLA_TILED_TRESH) {
+    r = pixSauvolaBinarizeTiled(pix_grey, 25, 0.40, 300, 300, pix_thresholds.a(), 
+                                pix_binary.a());
+  }
+  
+  bool ok = r == 0 ? true : false;
+  return std::make_tuple(ok, pix_grey, pix_binary, pix_thresholds);
 }
 
 // Threshold the source image as efficiently as possible to the output Pix.
--- a/src/ccmain/thresholder.cpp
+++ b/src/ccmain/thresholder.cpp
@@ -207,15 +207,15 @@ std::tuple<bool, Image, Image, Image> ImageThresholder::Threshold(
 
   pix_grey = GetPixRectGrey();
 
-  if (method == OTSU_TRESH || method >= TRESH_METHODS_COUNT) {
-    method = ADAPTIVE_OTSU_TRESH;
+  if (method == ThreshMethod::Otsu || method >= ThreshMethod::Count) {
+    method = ThreshMethod::AdaptiveOtsu;
   }
 
   int r;
-  if (method == ADAPTIVE_OTSU_TRESH) {
+  if (method == ThreshMethod::AdaptiveOtsu) {
     r = pixOtsuAdaptiveThreshold(pix_grey, 300, 300, 0, 0, 0.1, 
                                  pix_thresholds.a(), pix_binary.a());
-  } else if (method == SAUVOLA_TILED_TRESH) {
+  } else if (method == ThreshMethod::TiledSauvola) {
     r = pixSauvolaBinarizeTiled(pix_grey, 25, 0.40, 300, 300, pix_thresholds.a(), 
                                 pix_binary.a());
   }
--- a/src/ccstruct/image.h
+++ b/src/ccstruct/image.h
@@ -33,6 +33,8 @@ public:
   // service
   operator Pix *() const { return pix_; }
   Pix *operator->() const { return pix_; }
+  Pix **a() { return &pix_; }
+
 
   // api
   Image clone() const; // increases refcount
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -2108,20 +2108,41 @@ bool TessBaseAPI::Threshold(Pix **pix) {
     thresholder_->SetSourceYResolution(kMinCredibleResolution);
   }
   auto pageseg_mode = static_cast<PageSegMode>(static_cast<int>(tesseract_->tessedit_pageseg_mode));
-  Image im(*pix);
-  if (!thresholder_->ThresholdToPix(pageseg_mode, &im)) {
-    return false;
-  }
-  *pix = im;
+
+  Image pix_binary(*pix);
+  Image pix_grey;
+  Image pix_thresholds;
+
+  auto thresholding_method = static_cast<ThreshMethod>(static_cast<int>(tesseract_->thresholding_method));
+
+  if (thresholding_method == OTSU_TRESH) {
+    if (!thresholder_->ThresholdToPix(pageseg_mode, &pix_binary)) {
+      return false;
+    }
+    *pix = pix_binary;
+  
+    if (!thresholder_->IsBinary()) {
+      tesseract_->set_pix_thresholds(thresholder_->GetPixRectThresholds());
+      tesseract_->set_pix_grey(thresholder_->GetPixRectGrey());
+    } else {
+      tesseract_->set_pix_thresholds(nullptr);
+      tesseract_->set_pix_grey(nullptr);
+    }
+  } else {
+    auto [ok, pix_grey, pix_binary, pix_thresholds] = thresholder_->Threshold(thresholding_method);
+
+    if (!ok) {
+      return false;
+    }
+    *pix = pix_binary;
+
+    tesseract_->set_pix_thresholds(pix_thresholds);
+    tesseract_->set_pix_grey(pix_grey); 
+  } 
+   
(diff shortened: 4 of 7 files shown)


src/ccstruct/image.h
```diff
@@ -33,6 +33,8 @@ class TESS_API Image {
   // service
   operator Pix *() const { return pix_; }
   Pix *operator->() const { return pix_; }
+  Pix **a() { return &pix_; }
```
egorpugin, line 36: operator Pix**() ?

amitdo: It not really operator. It's a function call, but it acts like `&` operator.

egorpugin: I mean, `a()` is very very bad.
`operator Pix**()` must be used instead.

amitdo:
See #3083.

Fixes  #3377 (with Sauvola).

src/ccstruct/image.h
amitdo: Yes, I know it's not nice. I didn't know how to implement it using operator overloading. I don't write C++ code very often, so excuse me.

egorpugin: Try to replace it with `operator Pix**() const { return &pix_; }` and check if it works.

amitdo:
```
/../src/ccstruct/image.h:37:35: error: invalid conversion from ‘Pix* const*’ to ‘Pix**’ [-fpermissive]
   37 |   operator Pix**() const { return &pix_; }
      |                                   ^~~~~
      |                                   |
      |                                   Pix* const*

```

egorpugin: Try `operator const Pix**`.

egorpugin: Ok, this should work `operator Pix **() const { return (Pix **)&pix_; }`.

amitdo:
```
./../src/ccstruct/image.h: In member function ‘tesseract::Image::operator const Pix**() const’:
./../src/ccstruct/image.h:39:41: error: invalid conversion from ‘Pix* const*’ to ‘const Pix**’ [-fpermissive]
   39 |   operator const Pix**() const { return &pix_; }
      |                                         ^~~~~
      |                                         |
      |                                         Pix* const*

```

egorpugin: This is better `operator Pix **() { return &pix_; }` since we can modify `pix_` after return.

amitdo: with
`operator Pix* const*() const { return &pix_; }`

I don't get an error in `image.h`, but I get errors from other places.

amitdo:
```
/../src/ccmain/thresholder.cpp: In member function ‘void tesseract::ImageThresholder::SetImage(tesseract::Image)’:
./../src/ccmain/thresholder.cpp:157:12: error: ambiguous overload for ‘operator!=’ (operand types are ‘tesseract::Image’ and ‘std::nullptr_t’)
  157 |   if (pix_ != nullptr) {
      |       ~~~~~^~~~~~~~~~
./../src/ccmain/thresholder.cpp:157:12: note: candidate: ‘operator!=(Pix**, Pix**)’ <built-in>
./../src/ccmain/thresholder.cpp:157:12: note: candidate: ‘operator!=(Pix*, Pix*)’ <built-in>
./../src/ccmain/thresholder.cpp: In member function ‘virtual std::tuple<bool, tesseract::Image, tesseract::Image, tesseract::Image> tesseract::ImageThresholder::Threshold(tesseract::ThreshMethod)’:
./../src/ccmain/thresholder.cpp:217:34: error: cannot convert ‘tesseract::Image*’ to ‘PIX**’ {aka ‘Pix**’}
  217 |                                  &pix_thresholds, &pix_binary);
      |                                  ^~~~~~~~~~~~~~~
      |                                  |
      |                                  tesseract::Image*
In file included from ./../src/ccmain/thresholder.cpp:19:
/usr/include/leptonica/allheaders.h:152:146: note:   initializing argument 7 of ‘l_ok pixOtsuAdaptiveThreshold(PIX*, l_int32, l_int32, l_int32, l_int32, l_float32, PIX**, PIX**)’
  152 | LEPT_DLL extern l_ok pixOtsuAdaptiveThreshold ( PIX *pixs, l_int32 sx, l_int32 sy, l_int32 smoothx, l_int32 smoothy, l_float32 scorefract, PIX **ppixth, PIX **ppixd );
      |                                                                                                                                            ~~~~~~^~~~~~

```

egorpugin: I've pushed proper method. Please rebase.

amitdo: I get these errors:

```
./src/ccmain/thresholder.cpp: In member function ‘virtual std::tuple<bool, tesseract::Image, tesseract::Image, tesseract::Image> tesseract::ImageThresholder::Threshold(tesseract::ThreshMethod)’:
./../src/ccmain/thresholder.cpp:217:34: error: cannot convert ‘tesseract::Image*’ to ‘PIX**’ {aka ‘Pix**’}
  217 |                                  &pix_thresholds, &pix_binary);
      |                                  ^~~~~~~~~~~~~~~
      |                                  |
      |                                  tesseract::Image*
In file included from ./../src/ccmain/thresholder.cpp:19:
/usr/include/leptonica/allheaders.h:152:146: note:   initializing argument 7 of ‘l_ok pixOtsuAdaptiveThreshold(PIX*, l_int32, l_int32, l_int32, l_int32, l_float32, PIX**, PIX**)’
  152 | LEPT_DLL extern l_ok pixOtsuAdaptiveThreshold ( PIX *pixs, l_int32 sx, l_int32 sy, l_int32 smoothx, l_int32 smoothy, l_float32 scorefract, PIX **ppixth, PIX **ppixd );
      |                                                                                                                                            ~~~~~~^~~~~~
./../src/ccmain/thresholder.cpp:219:63: error: cannot convert ‘tesseract::Image*’ to ‘PIX**’ {aka ‘Pix**’}
  219 |     r = pixSauvolaBinarizeTiled(pix_grey, 25, 0.40, 300, 300, &pix_thresholds,
      |                                                               ^~~~~~~~~~~~~~~
      |                                                               |
      |                                                               tesseract::Image*
In file included from ./../src/ccmain/thresholder.cpp:19:
/usr/include/leptonica/allheaders.h:155:123: note:   initializing argument 6 of ‘l_ok pixSauvolaBinarizeTiled(PIX*, l_int32, l_float32, l_int32, l_int32, PIX**, PIX**)’
  155 | LEPT_DLL extern l_ok pixSauvolaBinarizeTiled ( PIX *pixs, l_int32 whsize, l_float32 factor, l_int32 nx, l_int32 ny, PIX **ppixth, PIX **ppixd );
      |                                                                                                                     ~~~~~~^~~~~~
make[2]: *** [Makefile:5704: src/ccmain/libtesseract_la-thresholder.lo] Error 1
```

egorpugin: You need `&pix_thresholds` -> `*pix_thresholds`.

include/tesseract/publictypes.h
```diff
@@ -278,6 +278,17 @@ enum OcrEngineMode {
   OEM_COUNT                    // Number of OEMs
 };
 
+/** 
+ * Except when OTSU_TRESH is chosen
+ * Leptonica is used for thresholding 
+ */
+enum ThreshMethod {
+  OTSU_TRESH,          // Legacy Tesseract's Otsu thresholding
```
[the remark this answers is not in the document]
amitdo: Yes, this is a typo.

amitdo:
@egorpugin, @stweil

Can you merge this PR and then do the changes related to the operator overloading?

egorpugin:
No. Why?

amitdo:
1) Why not? My patch work well and fix reported issues. What the problem to do any syntax improvement later?
2) I couldn't make it work (compile) with the new operator.

amitdo:
To be clear, I dont mean that you undo your changes to master.

include/tesseract/publictypes.h
```diff
@@ -278,6 +278,17 @@ enum OcrEngineMode {
   OEM_COUNT                    // Number of OEMs
 };
 
+/** 
+ * Except when OTSU_TRESH is chosen
+ * Leptonica is used for thresholding 
+ */
+enum ThreshMethod {
```
egorpugin, line 285: enum class?

amitdo: Sure, I can change it to enum class.

about style:

`OTSU_THRESH` ->
`otsu` ? 
`Otsu` ?
(droping the _thresh in both cases).

egorpugin: I think it should be Otsu.
You can also grep other `enum class`  vars in tess.

amitdo: https://github.com/tesseract-ocr/tesseract/search?q=%22enum+class%22&type=

'kName' is used in 3 out of 4 enum class cases.
This is google style. Should I also use it or should I drop the 'k' prefix?

egorpugin: No.

amitdo: >Should I also use it or should I drop the 'k' prefix?

> No.

No to which part of the question?

 1) Should I also use it
2) or should I drop the 'k' prefix?

egorpugin: Without `k`. Just `Otsu` etc.

amitdo:
I updated the PR. enum class is now used instead of plain enum.

The CI failures are not related to this PR.

amitdo:
Egor, feel free to get rid of `a()`.

src/ccmain/thresholder.cpp
```diff
@@ -184,6 +185,45 @@ void ImageThresholder::SetImage(const Image pix) {
   Init();
 }
 
+std::tuple<bool, Image, Image, Image> ImageThresholder::Threshold(
+                                                         ThreshMethod method) {
+  Image pix_grey = nullptr;
+  Image pix_binary = nullptr;
+  Image pix_thresholds = nullptr;
+
+  if (image_width_ > INT16_MAX || image_height_ > INT16_MAX) {
```
stweil, line 194: Where does this limit come from?

amitdo: I don't know. I just copied this part from another function that handles the legacy Otsu in the same file.

BTW, you can change `<climits>` to C++ `<limits>`. Not just in this file.

src/ccmain/thresholder.cpp
```diff
@@ -184,6 +185,45 @@ void ImageThresholder::SetImage(const Image pix) {
...
+  pix_grey = GetPixRectGrey();
+
+  if (method == ThreshMethod::Otsu || method >= ThreshMethod::Count) {
+    method = ThreshMethod::AdaptiveOtsu;
+  }
+
+  int r;
+  if (method == ThreshMethod::AdaptiveOtsu) {
+    r = pixOtsuAdaptiveThreshold(pix_grey, 300, 300, 0, 0, 0.1, 
+                                 pix_thresholds.a(), pix_binary.a());
```
bertsky, line 217: @amitdo I would advice making the tile size depenent on the pixel density of the input (`estimated_res_`) or on the pixel resolution (`image_width_` and `image_height_`). How large our tiles are relatively has a big impact: Is it likely to have white-only segments? Or (on the other extreme) will segments stretch across most of the page (conflating shaded and bright areas)?

bertsky: Furthermore, I think it would be useful to expose the `scorefract=0.1` default as a tunable Tesseract parameter.

src/ccmain/thresholder.cpp
```diff
@@ -184,6 +185,45 @@ void ImageThresholder::SetImage(const Image pix) {
...
+  } else if (method == ThreshMethod::TiledSauvola) {
+    r = pixSauvolaBinarizeTiled(pix_grey, 25, 0.40, 300, 300, pix_thresholds.a(), 
+                                pix_binary.a());
```
bertsky, line 220: @amitdo I am perplexed by this. According to Leptonica's API documentation, the 4th and 5th parameters are not tile **size** but tile **number** – so you don't get 300 but `image_width_ / 300` pixels here. 

Also, the _window size_ parameter of 25 seems **too low** for typical cases (like 300 DPI input) – remember that in the Sauvola formula, this needs to encompass enough statistics about each pixels environment to include both enough background and foreground. (I have found a good rule of thumb to be "window size equals DPI", because one inch of print usually contains enough whitespace around even large letters. I did experiment a lot with this myself, but cannot say much about the literature on the subject.)

Finally, I think the `factor=0.4` (which is the `k` parameter in the Sauvola formula, and usually is described as 0.34) should really be exposed as a Tesseract parameter.

bertsky (commented):
Great work! 

For robust quality, I think we need to optimise some details for these algorithms.

src/ccmain/thresholder.cpp
amitdo: HI @bertsky,  

>According to Leptonica's API documentation, the 4th and 5th parameters are not tile **size** but tile **number** – so you don't get 300 but `image_width_ / 300` pixels here.

Yes, you're right (but Leptonica is smart enough to correct my mistake in most cases). 

What should be the value here?

In another function, `pixSauvolaOnContrastNorm()` Leptonica has this code:

```
     /* Use tiles of size approximately 250 x 250 */
    pixGetDimensions(pix1, &w, &h, NULL);
    nx = L_MAX(1, (w + 125) / 250);
    ny = L_MAX(1, (h + 125) / 250);
    pixSauvolaBinarizeTiled(pix1, 25, 0.40, nx, ny, ppixth, &pixd);
```

bertsky: > Yes, you're right (but Leptonica is smart enough to correct my mistake in most cases).

Indeed, Leptonica does some range validation:
```C
    xrat = w / nx;
    yrat = h / ny;
    if (xrat < whsize + 2) {
        nx = w / (whsize + 2);
        L_WARNING("tile width too small; nx reduced to %d\n", procName, nx);
    }
    if (yrat < whsize + 2) {
        ny = h / (whsize + 2);
        L_WARNING("tile height too small; ny reduced to %d\n", procName, ny);
    }
```
So tiles will always be at least 1px larger than windows. (For us, of course, window size is equally problematic, as sketched above.)

> What should be the value here?
> 
> In another function, `pixSauvolaOnContrastNorm()` Leptonica has this code:

Hm, so they want to 250px tiles. That's not going to work if DPI is very large (say 1200) and therefore windows also need to be larger.

Since tile size is only about performance (memory / runtime trade-off), and window size only about quality (fg / bg statistics), I propose:

1. Use `estimated_res_` for DPI. (If input DPI is implausible, this will have been overriden by DPI estimation. In any case, the user can override DPI here as well.)
2. Set window size by multiplying DPI with a variable factor, let's say `thresholding_win_scale=0.5`. (The user can override that if she has a-priori knowledge about the material, e.g. 0.2 for small font size pages or 0.9 for title pages.)
3. Set tile size by dividing image size with a variable divisor, let's say `thresholding_tiles=8`, but ensure it is between window size and 1.

bertsky: Of course, we should first experiment on all these parameters with a suitable GT. Assuming we can definitely rule out any effects of binarization on neural recognition, all we need is layout evaluation. We could use PRImA tools for this on structural GT, or simple pixelwise mAP scoring...

amitdo: #3570 fixed the wrong tiles parameters in Sauvola.

src/ccmain/thresholder.cpp
amitdo: #3577 adds configurable variables to control thresholding.

il--ya:
> Use functions from Leptonica to provide more binarization options. The new options are: 1) Adaptive Otsu and 2) Sauvola (Tiled).

How do I use these options? I have a problem similar to #1990, and I'm looking for a solution.

Issue #1852, opened 2018-08-20
NKFC Normalization


Shreeshrii:
Tesseract uses NKFC Normalization.

As per http://unicode.org/reports/tr15/#Norm_Forms

> In the NFKC and NFKD forms, many formatting distinctions are removed, as shown in Figure 6. The “fi” ligature changes into its components “f” and “i”, the superscript formatting is removed from the “5”, and the long “s” is changed into a normal “s”.

>the composition phase of NFC and NFKC are the same—only their decomposition phase differs, with NFKC applying compatibility decompositions.

>Normalization Form KC additionally folds the differences between compatibility-equivalent characters that are inappropriately distinguished in many circumstances. For example, the halfwidth and fullwidth katakana characters will normalize to the same strings, as will Roman numerals and their letter equivalents. More complete examples are provided in Section 6, Examples and Charts.

See Figure 6 on  http://unicode.org/reports/tr15/#Norm_Forms

Shreeshrii:
I think this is the reason that superscripts are not recognized correctly. Should NKFC Normalization be changed to NFC?

Shreeshrii:
https://github.com/tesseract-ocr/tesseract/search?q=NFKChttps://github.com/tesseract-ocr/tesseract/search?q=NFC

PR #3420, opened 2021-05-11
Switch to NFC normalisation everywhere

nickjwhite:
Switch from NFKC to NFC normalisation (or switch from NFKD to NFD if decompose mode is requested). NFKC is not ideal for OCR, as characters which should be treated differently, such as superscripts and distinct alternative forms of letters, are treated as identical.

This is discussed in issue #1852.

### DIFF
--- a/src/training/pango/ligature_table.cpp
+++ b/src/training/pango/ligature_table.cpp
@@ -63,13 +63,13 @@ LigatureTable::LigatureTable()
 void LigatureTable::Init() {
   if (norm_to_lig_table_.empty()) {
     for (char32 lig = kMinLigature; lig <= kMaxLigature; ++lig) {
-      // For each char in the range, convert to utf8, nfkc normalize, and if
+      // For each char in the range, convert to utf8, nfc normalize, and if
       // the strings are different put the both mappings in the hash_maps.
       std::string lig8 = EncodeAsUTF8(lig);
       icu::UnicodeString unicode_lig8(static_cast<UChar32>(lig));
       icu::UnicodeString normed8_result;
       icu::ErrorCode status;
-      icu::Normalizer::normalize(unicode_lig8, UNORM_NFKC, 0, normed8_result, status);
+      icu::Normalizer::normalize(unicode_lig8, UNORM_NFC, 0, normed8_result, status);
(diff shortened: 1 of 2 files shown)


nickjwhite:
Sorry, this is causing some ligature tests to fail. I'll look into them and determine whether the tests need updating, or my patch does.

wollmers:
> Sorry, this is causing some ligature tests to fail. I'll look into them and determine whether the tests need updating, or my patch does.

IMHO you can remove the special rule

```
      // The icu::Normalizer maps the "LONG S T" ligature to "st". Correct that
      // here manually so that AddLigatures() will work as desired.
      if (lig8 == "\uFB05") {
        normed8 = "ſt";
      }
```

completely as NFKC of long_s + t is also st, and NFC keeps the long_s:

```
helmut@mbp:~$ uni print U+FB05
     cpoint  dec    utf-8       html       name
'ﬅ'  U+FB05  64261  ef ac 85    &#xfb05;   LATIN SMALL LIGATURE LONG S T (Lowercase_Letter)
helmut@mbp:~$ perl -e 'use Unicode::Normalize; use utf8; binmode(STDOUT,":encoding(UTF-8)"); print NFC("ﬅ"),"\n";'
ﬅ
helmut@mbp:~$ perl -e 'use Unicode::Normalize; use utf8; binmode(STDOUT,":encoding(UTF-8)"); print NFKC("ﬅ"),"\n";'
st
helmut@mbp:~$ perl -e 'use Unicode::Normalize; use utf8; binmode(STDOUT,":encoding(UTF-8)"); print NFC("ſt"),"\n";'
ſt
helmut@mbp:~$ perl -e 'use Unicode::Normalize; use utf8; binmode(STDOUT,":encoding(UTF-8)"); print NFKC("ſt"),"\n";'
st
```

Also there are many discretionary (typographic) ligatures like fi, fl, ij, ct, ſi, ſch, ſt, ſs, ſſ, ch, ck etc. It's a matter of the training data, if a user wants to transcribe long_s t to U+FB05, ſt or st, or even ch, ck to code points in the PUA. Same for other ligatures. For the traineddata delivered by the tesseract project it should be ſt.

PR #3428, opened 2021-05-15
Clean code for several recent modifications

stweil:


src/ccmain/thresholder.cpp
```diff
@@ -205,22 +204,19 @@ std::tuple<bool, Image, Image, Image> ImageThresholder::Threshold(
...
+  if (method == ThresholdMethod::TiledSauvola) {
     r = pixSauvolaBinarizeTiled(pix_grey, 25, 0.40, 300, 300, pix_thresholds,
                                 pix_binary);
+  } else {
+    // AdaptiveOtsu.
+    r = pixOtsuAdaptiveThreshold(pix_grey, 300, 300, 0, 0, 0.1,
+                                 pix_thresholds, pix_binary);
   }
 
-  bool ok = r == 0 ? true : false;
```
amitdo, line 223: I intend to add more options from Leptonica, but I guess I can keep this the last option.

stweil: That should not be a problem. You can also choose any order of the different methods. Just make sure that the last block is executed unconditionally. Then the compilers are happy.

src/api/baseapi.cpp
```diff
@@ -2107,15 +2107,12 @@ bool TessBaseAPI::Threshold(Pix **pix) {
...
-  auto pageseg_mode = static_cast<PageSegMode>(static_cast<int>(tesseract_->tessedit_pageseg_mode));
-
-  Image pix_binary(*pix);
-  Image pix_grey;
-  Image pix_thresholds;
 
   auto thresholding_method = static_cast<ThresholdMethod>(static_cast<int>(tesseract_->thresholding_method));
 
   if (thresholding_method == ThresholdMethod::Otsu) {
+    auto pageseg_mode = static_cast<PageSegMode>(static_cast<int>(tesseract_->tessedit_pageseg_mode));
```
amitdo, line 2114: You can drop it, it is not used in  ThresholdToPix().

src/ccmain/thresholder.cpp
```diff
@@ -190,11 +190,6 @@ std::tuple<bool, Image, Image, Image> ImageThresholder::Threshold(
   Image pix_binary = nullptr;
   Image pix_thresholds = nullptr;
 
-  if (image_width_ > INT16_MAX || image_height_ > INT16_MAX) {
```
amitdo, line 194: 1) As I said elsewhere, this block was copied from another function in the same file. You should delete the other block too.

amitdo: There might be security issues in allowing unlimited file size.

You should check if Leptonica has any image size limits.

egorpugin (approved):

src/ccmain/thresholder.cpp
```diff
@@ -205,22 +204,19 @@ std::tuple<bool, Image, Image, Image> ImageThresholder::Threshold(
...
 
   int r;
-  if (method == ThresholdMethod::AdaptiveOtsu) {
-    r = pixOtsuAdaptiveThreshold(pix_grey, 300, 300, 0, 0, 0.1,
-                                 pix_thresholds, pix_binary);
-  } else if (method == ThresholdMethod::TiledSauvola) {
+  if (method == ThresholdMethod::TiledSauvola) {
     r = pixSauvolaBinarizeTiled(pix_grey, 25, 0.40, 300, 300, pix_thresholds,
                                 pix_binary);
+  } else {
```
amitdo, line 208: Maybe print a warning message when method >= ThresholdMethod::Max ?

PR #3431, opened 2021-05-17
Fix crash in ObjectCache::DeleteUnusedObjects

nagadomi:
>for (auto i = cache_.size() - 1; i >= 0; i--) {
https://github.com/tesseract-ocr/tesseract/commit/4b5dd25b845d2ebf5ef7b15a86589d52f7b12495

The return type of .size() is unsigned, so the condition `i >= 0` will never be false and will crash.
This issue occurs when importing the Python `tesserocr` module.

```
% gdb --args bash python3
(gdb) run
>>> import tesserocr

Program received signal SIGSEGV, Segmentation fault.
tesseract::ObjectCache<tesseract::Dawg>::DeleteUnusedObjects (this=0x7ffff60d2c00 <tesseract::Dict::GlobalDawgCache()::cache>) at ./src/ccutil/object_cache.h:99
99            if (cache_[i].count <= 0) {
```

### DIFF
--- a/src/ccutil/object_cache.h
+++ b/src/ccutil/object_cache.h
@@ -95,7 +95,7 @@ public:
 
   void DeleteUnusedObjects() {
     std::lock_guard<std::mutex> guard(mu_);
-    for (auto i = cache_.size() - 1; i >= 0; i--) {
+    for (int i = (int)cache_.size() - 1; i >= 0; i--) {
       if (cache_[i].count <= 0) {
         delete cache_[i].object;
         cache_.erase(cache_.begin() + i);


src/ccutil/object_cache.h
```diff
@@ -95,7 +95,7 @@ class ObjectCache {
 
   void DeleteUnusedObjects() {
     std::lock_guard<std::mutex> guard(mu_);
-    for (auto i = cache_.size() - 1; i >= 0; i--) {
+    for (int i = (int)cache_.size() - 1; i >= 0; i--) {
```
stweil, line 98: The new fix is not the right one (C type cast). I think using a reverse iterator would be better.

stweil:
That's a bug which I caused in commit 4b5dd25b845d2ebf5ef7b15a86589d52f7b12495. Sorry for that. Obviously there is no unit test which tests that code.

src/ccutil/object_cache.h
nagadomi: reverse iterator version. It seems to work fine, but I don't know how to test it since I don't know how this class is used. It may be better to have you fix it again.
```c++
void DeleteUnusedObjects() {
  std::lock_guard<std::mutex> guard(mu_);
  for (auto it = cache_.rbegin(); it != cache_.rend(); ++it) { 
    if (it->count <= 0) {
      delete it->object;
      cache_.erase(it.base());
    }
  }
}
```

stweil: `apiexample_test` now tests that code (see commit d4bcbbd6defd97aeb5dbd1fbd58977dbd19563bd). I think your suggested code looks good.

stweil: ... looks good, but does not work. Code built with sanitizers crashes.

PR #3432, opened 2021-05-17
Refactor ObjectCache::DeleteUnusedObjects with reverse iterator

nagadomi:
Refactored #3431 according to @stweil 's code review.


nagadomi:
`std::vector::erase()` with reverse iterator seems a bit tricky, so the previous code may be safer.
Comments from people familiar with C++ are welcome.

PR #3434, opened 2021-05-18
Add --vertical_fontlist option to tesstrain.sh

nagadomi:
This Pull Request adds `--vertical_fontlist` option to `tesstrain.sh` to specify a list of fontnames to render vertical text.
The format for specifying a list of fontnames is the same as for `--font_list` option.
If `--vertical_fontlist <FONTS>` option is specified, it will override the `VERTICLA_FONTS` variable(defined in `language-specific.sh`) with the specified list of fontnames.

In the current version, the `VERTICAL_FONTS` variable is hardcoded in `language-specific.sh`. So, when creating training data for vertical text such as Japanese, users need to edit the source code even if they specify a list of fontnames with `--fontlist` and `--font_dir`.

### DIFF
--- a/src/training/language-specific.sh
+++ b/src/training/language-specific.sh
@@ -1179,6 +1179,10 @@ set_lang_specific_parameters() {
 
   # Default to 0 exposure if it hasn't been set
   test -z "$EXPOSURES" && EXPOSURES=0
+
+  # Overrides VERTICAL_FONTS if --vertical_fontlist option have been set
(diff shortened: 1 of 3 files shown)


stweil:
We dropped support for that shell code (it should be removed to make that clear), so the PR should not be merged.

Could you make those changes for the new Python code (in tesstrain repository)?

egorpugin:
> it should be removed to make that clear

Yes, just remove them then.

nagadomi:
I was planning to make the same pull request as the Python version of the tesstrain repository once this pull request was accepted.
If the shellcode is deprecated, I will close this pull request.

Issue #3184, opened 2020-12-18
Maximum supported image size


MerlijnWajer:
### Environment

* **Tesseract Version**: 4.1.1 and latest master
* **Platform**: `Linux gentoo-x230 5.6.18-grsec #2 SMP Tue Jul 7 18:17:17 CEST 2020 x86_64 Intel(R) Core(TM) i5-3320M CPU @ 2.60GHz GenuineIntel GNU/Linux`

### Current Behavior:

On large images, Tesseract fails like this:

```
Tesseract Open Source OCR Engine v4.1.1 with Leptonica
Warning: Invalid resolution 0 dpi. Using 70 instead.
Image too large: (2559, 37192)
Error during processing.
```

This is the image in question (large image!):
https://archive.org/download/manualzz-id-765154/765154_jp2.zip/765154_jp2/765154_0017.jp2

### Expected Behavior:

Tesseract would process the image  without erroring out.

### Comments

I don't know where exactly this limitation comes from. I see that the specific error comes from the Otsu thresholding code, but I am not sure if the limit of `2^15` (INT16_MAX) limit is actually also a leptonica maximum size limit.

Perhaps this is not considered a problem and the bug can be closed, but as it stands I am not sure what the best practice would be to OCR the image linked above. Perhaps the limit can be raised some, to `2^16` ?

stweil:
@MerlijnWajer, did you replace the image? I just loaded the image from the URL above, and it works because the size is 	 1706 x 24795 which is supported by the current tesseract.

Leptonica uses `l_int32` for image dimensions.

MerlijnWajer:
Yes, unfortunately someone changed the image to make it work for the existing item I linked to. I'll try to provide another sample file.

MerlijnWajer:
Here's another example:

https://wizzup.org/সহজ_কুরআন_শিক্ষা_0005.jp2

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/8b0c5405e2fa8cbecb3693c0074c5c8d1ae321b8/src/ccmain/thresholder.cpp#L188-L192

MerlijnWajer:
It looks like with `57b79742920cdda6d72e4fd7d0cab218db22f08b` this limited is removed, and now the output is: 

```
$ tesseract -c thresholding_method=2 /tmp/test.jp2 -
terminate called after throwing an instance of 'std::length_error'
  what():  cannot create std::vector larger than max_size()
```

[...]

PR #3435, opened 2021-05-19
Support image width and height larger than 32767

stweil:

### DIFF
--- a/src/ccstruct/points.cpp
+++ b/src/ccstruct/points.cpp
@@ -40,12 +40,32 @@ bool FCOORD::normalise() { // Convert to unit vec
   return true;
 }
 
+// Deserialize an ICOORD.
+// For compatibility reasons it uses unsigned 16 bit coordinates
+// instead of 32 bit coordinates.
 bool ICOORD::DeSerialize(TFile *f) {
-  return f->DeSerialize(&xcoord) && f->DeSerialize(&ycoord);
+  bool success = false;
+  uint16_t coord;
+  if (f->DeSerialize(&coord)) {
+    xcoord = coord;
+    if (f->DeSerialize(&coord)) {
+      ycoord = coord;
+      success = true;
+    }
+  }
+  return success;
 }
 
+// Serialize an ICOORD.
+// For compatibility reasons it uses unsigned 16 bit coordinates
+// instead of 32 bit coordinates.
 bool ICOORD::Serialize(TFile *f) const {
-  return f->Serialize(&xcoord) && f->Serialize(&ycoord);
+  uint16_t coord;
+  coord = xcoord;
+  auto success = f->Serialize(&coord);
+  coord = ycoord;
+  success = success && f->Serialize(&coord);
+  return success;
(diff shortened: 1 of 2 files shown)


stweil:
Please don't merge until these changes were better tested.

stweil:
The modifications here allow processing of large images which fixes issue #3184. I only tested the example in that issue and did not run performance tests or measure the increased memory usage.

stweil:
Known issues: code setting x / y / width / height to `INT16_MAX` or `-INT16_MAX` needs modifications, too.

amitdo:
>Please don't merge until these changes were better tested.

>This pull request is still a work in progress
Draft pull requests cannot be merged.

amitdo:
I wonder about the impact on regular size images.

How much more memory (in percents) will be consumed with this patch?

stweil:
> How much more memory (in percents) will be consumed with this patch?

I don't have numbers up to now.

egorpugin:
You changed int16->int32 a lot.
That must be a typedef (`using`) for a possible future changes?

stweil:
> That must be a typedef (`using`) for possible future changes?

Yes, the current Tesseract code uses `int16_t` which I replaced by `int32_t` for image dimensions.

I considered using a typedef, but don't think that there will be a future change for which a typedef would help. What kind of possible change do you think of?

egorpugin:
> What kind of possible change do you think of?

I agree, this may be subtle.
But from programmer point of view seeing `coord_t` instead of just `int32_t` annotate code much better.
You may note that in tess we have a lot of int vars which are hard to understand when navicating code here and there.

stweil:
>  But from programmer point of view seeing `coord_t` (or something like that) instead of just `int32_t` annotate code much better.

`coord_t` would violate the standards which reserve types ending with "_t".
I now use `TDimension`.

egorpugin:
> violate the standards

What standards?

stweil:
> What standards?

ISO C, see https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html.

stweil:
Should we make support for large images an optional feature in release 5.0.0 by adding a configure option, for example `--enable-large-images`?

egorpugin:
Or just ignore them.
If something big does not fit into 5.0.0 soon enough, just postpone.

amitdo:
If you choose to do it for 5.0.0, I suggest to add to the help message that explain this option: "experimental feature (use it only for testing purpose)" and in a comment in the code itself: "For package builders: We recommend not to enable this option because the feature is unstable/untested."

monotone:
> 

@stweil    Hi, Is this feat merged to main branched?  Why i just see this not changed at the tag 5.3.0 source code :
image

stweil:
No, it isn't merged. This pull request is still a draft.

monotone:
> 

it seems some years already, what‘s the reason?

monotone:
> 

it seems some years already, what‘s the reason?

PR #3437, opened 2021-05-21
ThresholdMethod::TiledSauvola -> ThresholdMethod::Sauvola

amitdo:
The fact that this method uses tiles is implementation detail. It does not change the result compared to Sauvola without tiles. The use of tiles minimize memory consumption.

### DIFF
--- a/include/tesseract/publictypes.h
+++ b/include/tesseract/publictypes.h
@@ -285,7 +285,7 @@ enum OcrEngineMode {
 enum class ThresholdMethod {
   Otsu,         // Legacy Tesseract's Otsu thresholding
   AdaptiveOtsu,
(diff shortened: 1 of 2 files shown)


include/tesseract/publictypes.h
```diff
@@ -285,7 +285,7 @@ enum OcrEngineMode {
 enum class ThresholdMethod {
```
stweil, line 285: @amitdo, why is `ThresholdMethod` declared in the public API? It looks like it is unused there, so maybe we should remove it for release 5.0.0.

amitdo: I don't remember right now why I put it there.You can move it elsewhere.

PR #288, opened 2016-03-21
Enable all ligatures available in a font for text2image rendering

nickjwhite:
This enables all OpenType ligatures for a specific font, where
available. Specifically, it explicitly enables the OpenType
features liga (standard ligatures), hlig (historical ligatures),
clig (contextual ligatures), and dlig (discretionary ligatures).

This feature requires Pango 1.38 or newer.

PR #3438, opened 2021-05-21
Raise Minimum required Pango version to 1.38.0

amitdo:


amitdo:
https://github.com/tesseract-ocr/tesseract/pull/288#issuecomment-802861225

PR #3467, opened 2021-06-24
Fix permanently disabled optimizations in CMake

nocun:
Macro CHECK_CXX_COMPILER_FLAG() will not update optimization variables if those were already defined.


zdenop:
Why this was merged? Removed code just set default state of variables used in compilation.
If compiler flag is supported than variable is correctly updated (otherwise nobody included me would have no SSE support on cmake build). 
So I would really like to see scenario where this code change solve problem.

egorpugin:
Default is always false, no need to set explicitly.
If some place cannot set a flag without these lines, that place must be fixed.

stweil:
> Default is always false

The default seems to be empty (which is handled like `FALSE`).

Old cmake output with wrong values for `HAVE_AVX` etc.:

```
[...]
-- HAVE_AVX: FALSE
-- HAVE_AVX2: FALSE
-- HAVE_FMA: FALSE
-- HAVE_SSE4_1: FALSE
-- MARCH_NATIVE_OPT: ON
-- HAVE_NEON: FALSE
-- Link-time optimization: FALSE
[...]
```

```
[...]
New cmake output (note the empty value of `HAVE_NEON`:

-- HAVE_AVX: 1
-- HAVE_AVX2: 1
-- HAVE_FMA: 1
-- HAVE_SSE4_1: 1
-- MARCH_NATIVE_OPT: ON
-- HAVE_NEON: 
-- Link-time optimization: FALSE
[...]
```

stweil:
I think it would be better to delay merging of new non-critical pull requests at least one day to allow comments by other contributors. In this case there was a build error which required a fix of course, but there might have been a better solution.

egorpugin:
That cmake output code must learn how to print empty vars.

nocun:
> So I would really like to see scenario where this code change solve problem.

When doing a clean cmake build these variables were not set by CHECK_CXX_COMPILER_FLAG() macros. See here:

https://stackoverflow.com/questions/25451254/getting-cmake-check-cxx-compiler-flag-to-work/49297190#49297190

> there might have been a better solution

What do you have in mind?

Issue #3473, opened 2021-06-29
Issues with swapping GenericVector for std::vector


nocun:
### Environment

* **Tesseract Version**: [master](https://github.com/tesseract-ocr/tesseract/commit/c583ecef2997586d86feed81ec58a179707b6705)
* **Commit Number**: [92b6c652](https://github.com/tesseract-ocr/tesseract/commit/92b6c652f36ceaf6ac8d77bc4e627117780a3e4c)..[65d882f9](https://github.com/tesseract-ocr/tesseract/commit/65d882f96e1fd4bb3c4cef43d07aece54eb83d36)
* **Platform**: Linux FC29 x86_64

### Problem
Replacing `GenericVector` with `std::vector` has resulted in a small execution time performance decrease (about 4% on my simple test). I am not sure whether some other bugs have been introduced this way.

Most of the issues I could see are from changing `generic_vector.init_to_size(n, x);` directly into `vector.resize(n, x);`. It is because `init_to_size()` not only resizes the vector but also resets *all* of its values (not only the new ones).

### Suggested Fix

Recheck the logic and update `GenericVector`'s changed lines:

gv.init_to_size(n, x) and gv.resize(n, x) should become:
```c++
v.clear();
v.resize(n, x);
```

gv.truncate(n) (other than truncate(0) which is just v.clear()) should become:
```c++
if (n < v.size()) {
    v.resize(n);
}
```

gv.resize_no_init(n) (I think this one was ok):
```c++
v.resize(n);
```

I have added an example in comments of a [commit](https://github.com/tesseract-ocr/tesseract/commit/9710bc0465fc0df1a5bc0e71177151e8c5901204#diff-1e9b4b30348fc110f27ca4095a364df6161b8ded2f04a2625674f06b09a1ab64) where the first performance hit was noted .

As far as I could see the code affected is within commit range [92b6c652](https://github.com/tesseract-ocr/tesseract/commit/92b6c652f36ceaf6ac8d77bc4e627117780a3e4c)..[65d882f9](https://github.com/tesseract-ocr/tesseract/commit/65d882f96e1fd4bb3c4cef43d07aece54eb83d36). I might be able to provide a PR but this could take some time.

stweil:
Thank you for this valuable hint. What do you mean with the "performance decrease" of 4 %? Increased character error rates? Larger execution time?

nocun:
[...]

PR #3474, opened 2021-06-29
Fix vector resize with init for all elements (issue #3473)

stweil:
Fixes: 9710bc0465fc0df1a5bc0e71177151e8c5901204


stweil:
Sorry, I just noticed that I only fixed src/ccmain, so please wait until I have fixes for the rest, too.

stweil:
> Sorry, I just noticed that I only fixed src/ccmain, so please wait until I have fixes for the rest, too.

Now all code is covered by this commit.

PR #3476, opened 2021-06-29
Fix for LSTM Diplopia issue

woodjohndavid:
This is a proposed fix for the LSTM diplopia problem where 2 characters are included in the LSTM output stream for the same physical position in the original image.  According to my review of trace output for a limited number of test cases, the issue occurs when there are 2 possible characters essentially 'competing' for the same spot, where one of those characters is a better match in the earlier timesteps but the second character (usually the better eventual match) becomes the better choice in later timesteps.  In this scenario, it is possible that there will be a beam which includes the first character choice and then adds the second character choice in the same beam after the first, once the first choice score has been reduced and it no longer appears in the TopN list.

This solution is limited to solving diplopia for 2 characters, but could be expanded to deal with a multiple character scenario.

This solution is also dependent upon assigning a value to variable kMinDiplopiaKey which is the minimum score (key value) for an output entry coming from the matrix which would be considered a likely valid character.  See the code for further details.

I have tested this only on a very small set of diplopia problems.  I am assuming that you folks have a much more extensive set of test cases to run this proposed change on to ensure that it has no unexpected results.


src/lstm/recodebeam.cpp
```diff
@@ -657,20 +667,46 @@ void RecodeBeamSearch::ComputeTopN(const float *outputs, int num_outputs, int to
...
     }
   }
+  float top_key = 0.0F;
+  float second_key = 0.0F;
+  bool found_first_code = false;
+  bool found_second_code = false;
   while (!top_heap_.empty()) {
     TopPair entry;
     top_heap_.Pop(&entry);
+    if (in_possible_diplopia_ && entry.data() == first_diplopia_code_)
```
stweil, line 677: Please use {}.

src/lstm/recodebeam.cpp
```diff
@@ -1189,6 +1229,20 @@ bool RecodeBeamSearch::UpdateHeapIfMatched(RecodeNode *new_node, RecodeHeap *hea
   return false;
 }
 
+// Determines if node can be added to heap based on possible diplopia status
+bool RecodeBeamSearch::AddToHeapIsAllowed(RecodeNode *new_node) {
+  if (!in_possible_diplopia_)
+    return true;
```
stweil, line 1235:
```suggestion
  if (!in_possible_diplopia_) {
    return true;
  }
```

src/lstm/recodebeam.h
```diff
@@ -425,6 +428,11 @@ class TESS_API RecodeBeamSearch {
   bool is_simple_text_;
   // The encoded (class label) of the null/reject character.
   int null_char_;
+  // Variables used in tracking possible diplopia case
```
stweil, line 431: Adding that new block after `bool is_simple_text_` would reduce the class size because it avoids another hole.

src/lstm/recodebeam.cpp
woodjohndavid: Hi Stefan,

Thanks for reviewing my code.  As this is the first time I have created a pull request, I am still learning the ropes and have a couple of questions.

For the comments you have made, should I make those corrections and then recommit to my branch JDWDIPLOPIA?  And then do I use the 'resolve conversation' button to let you know?

Dave

stweil: Yes, you can recommit to your branch, either by amending to existing commits (then you have to use a forced push) or by adding more commits. GitHub sometimes resolves conversations automatically.

woodjohndavid:
Stefan, I just committed the changes you suggested to branch JDWDIPLOPIA

Thanks,

Dave

src/lstm/recodebeam.cpp
stweil, line 1235:
```suggestion
  if (!in_possible_diplopia_) {
    return true;
  }
```

stweil:
> Stefan, I just committed the changes you suggested to branch JDWDIPLOPIA

I still don't see that latest changes. Did you push them to GitHub?

woodjohndavid:
Hi Stefan,

OK, apologies, newbie inexperience at this end.  I have no previous Git experience.  I am using GitHub Desktop for source control.  I had committed the changes at my end, but had not done the 'push to origin'.  Please check again now and hopefully you will find the changes there.

Sorry about that.

Dave

Robyer (commented):
Just few pedantic suggestions about code style consistency.

juligreen:
I'm sorry for making this kind of comment, but what's the status on this?

src/lstm/recodebeam.cpp
```diff
@@ -40,6 +40,11 @@ const int RecodeBeamSearch::kBeamWidths[RecodedCharID::kMaxCodeLen + 1] = {
 
 static const char *kNodeContNames[] = {"Anything", "OnlyDup", "NoDup"};
 
+// the minimum diplopia key is the minimum score (key) from
+// the network output to qualify as a likely 'real' character
+// for the purposes of identifying possible diplopia
```
stweil, line 45:
```suggestion
// The minimum diplopia key is the minimum score (key) from
// the network output to qualify as a likely 'real' character
// for the purposes of identifying possible diplopia.
```

src/lstm/recodebeam.cpp
```diff
@@ -657,20 +667,48 @@ void RecodeBeamSearch::ComputeTopN(const float *outputs, int num_outputs, int to
...
         top_code_ = entry.data();
+        top_key = entry.key();
       } else {
         second_code_ = entry.data();
+        second_key = entry.key();
       }
     }
   }
+  // need to identify if we are in a potential diplopia situation
+  // or if we already are, then determine if it is ended
```
stweil, line 697:
```suggestion
  // Need to identify if we are in a potential diplopia situation
  // or if we already are, then determine if it is ended.
```

src/lstm/recodebeam.cpp
```diff
@@ -1189,6 +1231,21 @@ bool RecodeBeamSearch::UpdateHeapIfMatched(RecodeNode *new_node, RecodeHeap *hea
   return false;
 }
 
+// Determines if node can be added to heap based on possible diplopia status
```
stweil, line 1234:
```suggestion
// Determines if node can be added to heap based on possible diplopia status.
```

src/lstm/recodebeam.h
```diff
@@ -377,6 +377,9 @@ class TESS_API RecodeBeamSearch {
   // Searches the heap for an entry matching new_node, and updates the entry
   // with reshuffle if needed. Returns true if there was a match.
   bool UpdateHeapIfMatched(RecodeNode *new_node, RecodeHeap *heap);
+  // Determines if new node can be added to the heap for the current beam.
+  // Returns false if we are in possible diplopia situation
```
stweil, line 381:
```suggestion
  // Returns false if we are in possible diplopia situation.
```

src/lstm/recodebeam.h
```diff
@@ -423,6 +426,11 @@ class TESS_API RecodeBeamSearch {
   // True if the input is simple text, ie adjacent equal chars are not to be
   // eliminated.
   bool is_simple_text_;
+  // Variables used in tracking possible diplopia case
```
stweil, line 429:
```suggestion
  // Variables used in tracking possible diplopia case.
```

src/lstm/recodebeam.h
```diff
@@ -423,6 +426,11 @@ class TESS_API RecodeBeamSearch {
...
+  // Refer to ComputeTopN routine for use of these variables
```
stweil, line 430:
```suggestion
  // Refer to ComputeTopN routine for use of these variables.
```

src/lstm/recodebeam.cpp
```diff
@@ -1138,6 +1176,10 @@ void RecodeBeamSearch::PushHeapIfBetter(int max_size, int code, int unichar_id,
     if (UpdateHeapIfMatched(&node, heap)) {
       return;
     }
+    // check to see if node is possible diplopia
```
stweil, line 1179:
```suggestion
    // Check to see if node is possible diplopia.
```

src/lstm/recodebeam.cpp
```diff
@@ -65,6 +70,9 @@ RecodeBeamSearch::RecodeBeamSearch(const UnicharCompress &recoder, int null_char
     , beam_size_(0)
     , top_code_(-1)
     , second_code_(-1)
+    , in_possible_diplopia_(false)
+    , first_diplopia_code_(-1)
+    , second_diplopia_code_(-1)
```
stweil, line 75:
```suggestion
```

stweil:
> what's the status on this

Meanwhile there are some merge conflicts which must be resolved.

But the most important thing is that we need to test the changes that they improve the diplopia issue and that they don't introduce a regression. So it would help if whoever has done OCR before and after those changes and can confirm that it improves the OCR result without making some recognition worse could report it here.

juligreen:
It doesn't seem to resolve the issue for me:

Recognized text: `けげんこうちょう…?知らない町だな。`
(The first character is duplicated)

Original Image:
screenshot(575)

After applying my optimizations:
image

To make sure I'm actually using the right build, here are my steps to compile:
* checking out the repo
* checking out the PR with `gh pr checkout 3476`
* `./autogen.sh`
* `./configure`
* `./make`

johndavidwood:
Hello all:

Sorry, but I have no previous experience with open source development mechanisms, so not sure how to move things forward.

The changes I made in this pull request have been successfully tested by me, but admittedly only on a limited set of data and the fairly specific requirements I have, namely English text from computer screen contents with specific limited fonts being used.  So indeed there is much more testing needed, but I have no mechanism for doing this.

I also have no idea how to deal with the merge conflicts mentioned.

For the test results reported by Juligreen, I would suggest that you investigate different values for the variable kMinDiplopiaKey.  This value is critical to the identification of potential diplopia, and probably should be make into a configuration setting.  But again, how to do that is beyond my level of experience.

woodjohndavid:
OK I have fixed the merging issue I think and committed them to my branch.  Hopefully I did it properly.  @stweil please let me know.

ohk2kt3t4:
With a dataset that produced 103 "diplopia" affected results, this branch reduced it down to only 38.

woodjohndavid:
Thanks @ohk2kt3t4 that is good to know.  Could you please attach a couple of the images you used where the diplopia was not successfully eliminated?  I would like to take a look and see if I can figure out why.

ohk2kt3t4:
Unfortunately I cannot share the images, but I can try help you investigate if it is trivial enough.

ohk2kt3t4:
All my images are of part numbers which have the same number of characters and follow the same character placement pattern (e.g. the first two characters are always digits): image.

When "diplopia" occurs, the resulting length of the OCR is always longer, and usually because tesseract has output two characters in the place of one (see a lot of "0O", "B8" pairs). So in that sense I could automatically detect them, but it might not be useful for other usecases. Perhaps could try generating some images programmatically and see if can end up with some that will trigger diplopia..

MerlijnWajer:
I am also seeing this problem in some cases, were we try to automatically OCR a lot of text, store it in spreadsheets, and then manually verify the results. I can pull in this pull request and re-run that OCR, and check if the diplopia problems are reduced or gone, if that's helpful. It'd be a narrow/specific set of mostly images like this:

image

Please let me know if that's helpful, and I'll try to do the evaluation. The images should be public, so I could share results where diplopia is (potentially) not eliminated.

woodjohndavid:
@MerlijnWajer it would be very helpful if you could test my diplopia fix as much as possible, and report your findings here.  Also, if there are any diplopia examples which remain, and if you are able to share your images, please attach them here after your testing so I can take a further look.

I do know that the fix I posted was limited in scope, but it seemed to correct the diplopia cases we have generally encountered in our own use of Tesseract.  Typically those cases involved situations where one particular character is fairly closely matched at the beginning of the given image segment, but then a second character turns out to be a better match as that image segment is further traversed.  This can result in one or more beams which contain both characters, with the likelihood that such a beam will get a higher score and therefore be considered as the best match.

The fix as it currently stands does not address cases where there are more than 2 potential matching characters.

woodjohndavid:
@ohk2kt3t4 I understand that you are unable to share your images on this site, but perhaps you could just extract a few specific part numbers where diplopia is still occurring using my fix, and just attach partial images of those.

exander77:
Not perfect, but huge improvement, I support pulling this. I don't think there is any negative side effect.

woodjohndavid:
As mentioned earlier on, I have no previous experience participating in an open source community.  However, I am a very experienced developer and am interested in working on Tesseract issues, particularly those that are related to OCR of non-textual data, like part numbers, codes, etc.  It is particularly in those cases that diplopia is a problem when it occurs.

I have asked earlier but need to ask again.  How is it decided that things like the change that I have put forward in this pull request actually get included in the primary Tesseract release?  What should I be doing to move this forward?

As indicated earlier in this thread, there have been some fairly large test runs done by @ohk2kt3t4 which seem to indicate that the fix in this pull request does fix a large percentage of diplopia cases (although not all) and does not seem to have negative side effects.  So I am not sure why it hasn't moved forward, even though it is not a full solution to the diplopia issue.

In the meanwhile, I am continuing to work with the code, and trying to see if I can come up with a more universal solution.

exander77:
@woodjohndavid Hello, I myself have come here to see if this fixes my wrongly positioned boxes issues, but it doesn't, I wrote a lot here: https://github.com/tesseract-ocr/tesseract/issues/3477

I am interested in this:
> As I see it, therefore the LSTM matrix processing using the NetworkIO interface needs to add to its return values (in addition to the possible character and the likelihood score) the starting pixel location of the possible match, and the horizontal size of the potential match image from the train data. Once that is done, the rest should be relatively straightforward.

Can you point me to the code? I have a hard time navigating it.

woodjohndavid:
Yes, the code is hard to follow.  And I agree with you that the ultimate solution should be that the LSTM engine returns the match coordinates in some form.  However, it does not at this point, and I do not understand the LSTM engine operation sufficiently well to figure out how to get it to return those coordinates.

So what I have been working on is the code that runs after the LSTM engine does its thing, to see if there is a way to solve the diplopia problem at that stage.  This pull request is my initial attempt on this, which is partially but not entirely successful.  It does NOT make any attempt to correct the inaccurate box dimensions.

One thing you can try (which I did also) is to reduce the "timestep" size.  That does improve the box dimensions with the code as it is otherwise, although when diplopia occurs they are still messed up.  However, it seems that reducing the "timestep" would require full re-training of the LSTM model.

exander77:
> However, it does not at this point, and I do not understand the LSTM engine operation sufficiently well to figure out how to get it to return those coordinates.

Yeah, that pretty much summarizes my experiences so far. There is a tonne of comments in the code, but not really any explanation of the operation.

exander77:
@woodjohndavid I am not ever sure where are the character positions calculated.

liuyl07:
@MerlijnWajer Thanks for guiding me to this PR!
@woodjohndavid Hello, I tested the below image with your fixes in the PR with different kMinDiplopiaKey values (0.25,0.5,0.75), but the result (DOT **0**O4N 6VHPPC) keeps unchanged with an extra ZERO in bold.
TesseractInputImageSingle

Could you please take a look if it is possible to improve on this case?

liuyl07:
Updates for the my question above.

By printing the below log in the function RecodeBeamSearch::ExtractBestPathAsWords(), it is obvious that there are 2 possible characters (zero 0 and big O) essentially 'competing' for the same spot
image

However, it seems that various kMinDiplopiaKey values don't help for this particular case...

liuyl07:
@woodjohndavid Sorry to trouble you. From all your comments in the pull request, I think you should be definitely interested in some cases where your fixes can be further improve. Could you take a look for my example which meets the conditions of your fixes, i.e., diplopia for 2 characters (0 and O).

woodjohndavid:
Another update @liuyl07

I did run your sample on my system, and did NOT have the same result as you did.  I did not get the diplopia '0O' I just got the 'O'

Dellu:
Is this code merged?
I am getting a lot of diplopia in my projects lately. I was wondering if it can reduce it.

tfmorris:
Sounds like this is superseded by #4211. Should it be closed?

PR #3481, opened 2021-06-30
Fix possible UB when accessing empty vector's data

nocun:
Make sure that the empty vector's data is never used, `at()` will throw an exception in that case.

This may allow compilers to better optimize code.

### DIFF
--- a/src/lstm/networkscratch.h
+++ b/src/lstm/networkscratch.h
@@ -148,7 +148,8 @@ public:
       vec_ = scratch_space_->vec_stack_.Borrow();
       vec_->reserve(reserve);
       vec_->resize(size);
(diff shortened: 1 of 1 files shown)


stweil:
I'd expect that code using `at()` is slower, so I am not sure whether this is really an improvement. Did you test code size and performance?

nocun:
I did not perform large scale tests so can't be sure if this is faster in all cases.

In my test case (single long line) this looks to be faster. Init() is not called that many times (about 30) and there should be virtually no exception related performance hit cause nothing is thrown.

nocun:
On the other hand it may be beneficial to do some testing on non-x86 platforms but I do not have those available atm.

Issue #3482, opened 2021-06-30
CMake clang build fail


nocun:
### Environment

* **Tesseract Version**: master (currently [38f0fdcd](https://github.com/tesseract-ocr/tesseract/commit/38f0fdcd888ebea26d44ea11c8a3365eb73e3232))
* **Platform**: Linux (possibly Windows) when compiling with Clang 7

### Problem

The `CHECK_CXX_COMPILER_FLAG()` macros seem to still cause problems.

E.g. in [CMakeLists.txt:180](https://github.com/tesseract-ocr/tesseract/blob/master/CMakeLists.txt#L180):
```
...
    CHECK_CXX_COMPILER_FLAG("-mfpu=neon" HAVE_NEON)
```

x86 GCC does not support this flag so `HAVE_NEON` remains unset.
On the same x86 machine but using Clang this line sets the flag to 1.

The reason is that `g++ -mfpu=neon` will error out but `clang++ -mfpu=neon` just gives a warning:
```
clang-7: warning: argument unused during compilation: '-mfpu=neon' [-Wunused-command-line-argument]
```
And the process will end with status 0 (no error). Actually you can put anything there and CMake will always set this flag to 1 when using Clang:
```
clang-7: warning: argument unused during compilation: '-mfpu=kettle' [-Wunused-command-line-argument]
```

This causes current cmake-clang jobs failures. I am unsure how to fix it easily.

stweil:
@nocun, do you want to try pull request #3484? I'd merge that if it fixes this issue.

nocun:
@stweil yes, this works, thank you!

PR #3484, opened 2021-07-02
Use CMAKE_SYSTEM_PROCESSOR to check for SIMD support options

stweil:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,16 +140,12 @@ if(COMPILER_SUPPORTS_MARCH_NATIVE)
     set(MARCH_NATIVE_OPT ON)
 endif()
 
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64.*|AARCH64.*")
- set(AARCH64 TRUE)
-endif()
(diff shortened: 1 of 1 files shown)


stweil:
Maybe this fixes issue #3482.

CMakeLists.txt
```diff
@@ -140,16 +140,12 @@ if(COMPILER_SUPPORTS_MARCH_NATIVE)
...
 endif()
 
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64.*|AARCH64.*")
- set(AARCH64 TRUE)
-endif()
+message("CMAKE_SYSTEM_PROCESSOR=<${CMAKE_SYSTEM_PROCESSOR}>")
 
-if(AARCH64)
+# Flags for SIMD support
+set(HAVE_NEON FALSE)
```
nocun, line 146: I didn't test the ARM platform but shouldn't this `set()` be removed? I think it may interfere with line [224](https://github.com/tesseract-ocr/tesseract/pull/3484/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR224).

stweil: You are right, thank you. Fixed in commit e0af8d12e6c8a58cda998cff975dc97bb63e425a.

PR #3486, opened 2021-07-05
Add TFloat data type for neural network

stweil:
Up to now Tesseract used double for training and recognition
with "best" models.

This commit replaces double by a new data type `TFloat` which
is double by default, but float if FAST_FLOAT is defined.

Ideally this should allow faster training.

### DIFF
--- a/src/arch/dotproductavx.cpp
+++ b/src/arch/dotproductavx.cpp
@@ -29,6 +29,28 @@ namespace tesseract {
 
 // Computes and returns the dot product of the n-vectors u and v.
 // Uses Intel AVX intrinsics to access the SIMD instruction set.
+#if defined(FAST_FLOAT)
+float DotProductAVX(const float *u, const float *v, int n) {
+  const unsigned quot = n / 8;
+  const unsigned rem = n % 8;
+  __m256 t0 = _mm256_setzero_ps();
+  for (unsigned k = 0; k < quot; k++) {
+    __m256 f0 = _mm256_loadu_ps(u);
+    __m256 f1 = _mm256_loadu_ps(v);
+    f0 = _mm256_mul_ps(f0, f1);
+    t0 = _mm256_add_ps(t0, f0);
+    u += 8;
+    v += 8;
+  }
+  alignas(32) float tmp[8];
+  _mm256_store_ps(tmp, t0);
+  float result = tmp[0] + tmp[1] + tmp[2] + tmp[3] + tmp[4] + tmp[5] + tmp[6] + tmp[7];
+  for (unsigned k = 0; k < rem; k++) {
+    result += *u++ * *v++;
+  }
+  return result;
+}
+#else
 double DotProductAVX(const double *u, const double *v, int n) {
   const unsigned quot = n / 8;
   const unsigned rem = n % 8;
@@ -57,6 +79,7 @@ double DotProductAVX(const double *u, const double *v, int n) {
   }
   return result;
 }
+#endif
 
 } // namespace tesseract.
 
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -96,7 +96,11 @@ bool SIMDDetect::sse_available_;
 static TFloat DotProductAccelerate(const TFloat* u, const TFloat* v, int n) {
   TFloat total = 0;
   const int stride = 1;
+#if defined(FAST_FLOAT)
+  vDSP_dotpr(u, stride, v, stride, &total, n);
+#else
   vDSP_dotprD(u, stride, v, stride, &total, n);
+#endif
   return total;
 }
 #endif
--- a/src/arch/dotproductfma.cpp
+++ b/src/arch/dotproductfma.cpp
@@ -31,26 +31,26 @@ namespace tesseract {
 // Uses Intel FMA intrinsics to access the SIMD instruction set.
 #if defined(FAST_FLOAT)
 float DotProductFMA(const float *u, const float *v, int n) {
-  const unsigned quot = n / 8;
-  const unsigned rem = n % 8;
+  const unsigned quot = n / 16;
+  const unsigned rem = n % 16;
   __m256 t0 = _mm256_setzero_ps();
   __m256 t1 = _mm256_setzero_ps();
   for (unsigned k = 0; k < quot; k++) {
     __m256 f0 = _mm256_loadu_ps(u);
     __m256 f1 = _mm256_loadu_ps(v);
     t0 = _mm256_fmadd_ps(f0, f1, t0);
-    u += 4;
-    v += 4;
+    u += 8;
+    v += 8;
     __m256 f2 = _mm256_loadu_ps(u);
     __m256 f3 = _mm256_loadu_ps(v);
     t1 = _mm256_fmadd_ps(f2, f3, t1);
-    u += 4;
-    v += 4;
+    u += 8;
+    v += 8;
   }
   t0 = _mm256_hadd_ps(t0, t1);
-  alignas(32) float tmp[4];
+  alignas(32) float tmp[8];
   _mm256_store_ps(tmp, t0);
-  float result = tmp[0] + tmp[1] + tmp[2] + tmp[3];
+  float result = tmp[0] + tmp[1] + tmp[2] + tmp[3] + tmp[4] + tmp[5] + tmp[6] + tmp[7];
   for (unsigned k = 0; k < rem; k++) {
     result += *u++ * *v++;
   }
--- a/src/arch/dotproductfma.cpp
+++ b/src/arch/dotproductfma.cpp
@@ -29,6 +29,34 @@ namespace tesseract {
 
 // Computes and returns the dot product of the n-vectors u and v.
 // Uses Intel FMA intrinsics to access the SIMD instruction set.
+#if defined(FAST_FLOAT)
+float DotProductFMA(const float *u, const float *v, int n) {
+  const unsigned quot = n / 8;
+  const unsigned rem = n % 8;
+  __m256 t0 = _mm256_setzero_ps();
+  __m256 t1 = _mm256_setzero_ps();
+  for (unsigned k = 0; k < quot; k++) {
+    __m256 f0 = _mm256_loadu_ps(u);
+    __m256 f1 = _mm256_loadu_ps(v);
+    t0 = _mm256_fmadd_ps(f0, f1, t0);
+    u += 4;
+    v += 4;
+    __m256 f2 = _mm256_loadu_ps(u);
+    __m256 f3 = _mm256_loadu_ps(v);
+    t1 = _mm256_fmadd_ps(f2, f3, t1);
+    u += 4;
+    v += 4;
+  }
+  t0 = _mm256_hadd_ps(t0, t1);
+  alignas(32) float tmp[4];
+  _mm256_store_ps(tmp, t0);
+  float result = tmp[0] + tmp[1] + tmp[2] + tmp[3];
+  for (unsigned k = 0; k < rem; k++) {
+    result += *u++ * *v++;
+  }
+  return result;
+}
+#else
 double DotProductFMA(const double *u, const double *v, int n) {
   const unsigned quot = n / 8;
   const unsigned rem = n % 8;
@@ -55,6 +83,7 @@ double DotProductFMA(const double *u, const double *v, int n) {
   }
   return result;
 }
+#endif
 
 } // namespace tesseract.
 
--- a/src/arch/dotproductsse.cpp
+++ b/src/arch/dotproductsse.cpp
@@ -30,6 +30,15 @@ namespace tesseract {
 
 // Computes and returns the dot product of the n-vectors u and v.
 // Uses Intel SSE intrinsics to access the SIMD instruction set.
+#if defined(FAST_FLOAT)
+TFloat DotProductSSE(const TFloat *u, const TFloat *v, int n) {
+  TFloat total = 0.0;
+  for (int k = 0; k < n; ++k) {
+    total += u[k] * v[k];
+  }
+  return total;
+}
+#else
 double DotProductSSE(const double *u, const double *v, int n) {
   int max_offset = n - 2;
   int offset = 0;
@@ -78,6 +87,7 @@ double DotProductSSE(const double *u, const double *v, int n) {
   }
   return result;
 }
+#endif
 
 } // namespace tesseract.
 
--- a/src/arch/dotproductsse.cpp
+++ b/src/arch/dotproductsse.cpp
@@ -31,12 +31,63 @@ namespace tesseract {
 // Computes and returns the dot product of the n-vectors u and v.
 // Uses Intel SSE intrinsics to access the SIMD instruction set.
 #if defined(FAST_FLOAT)
-TFloat DotProductSSE(const TFloat *u, const TFloat *v, int n) {
-  TFloat total = 0.0;
-  for (int k = 0; k < n; ++k) {
-    total += u[k] * v[k];
+float DotProductSSE(const float *u, const float *v, int n) {
+  int max_offset = n - 4;
+  int offset = 0;
+  // Accumulate a set of 4 sums in sum, by loading pairs of 4 values from u and
+  // v, and multiplying them together in parallel.
+  __m128 sum = _mm_setzero_ps();
+  if (offset <= max_offset) {
+    offset = 4;
+    // Aligned load is reputedly faster but requires 16 byte aligned input.
+    if ((reinterpret_cast<uintptr_t>(u) & 15) == 0 &&
+        (reinterpret_cast<uintptr_t>(v) & 15) == 0) {
+      // Use aligned load.
+      sum = _mm_load_ps(u);
+      __m128 floats2 = _mm_load_ps(v);
+      // Multiply.
+      sum = _mm_mul_ps(sum, floats2);
+      while (offset <= max_offset) {
+        __m128 floats1 = _mm_load_ps(u + offset);
+        floats2 = _mm_load_ps(v + offset);
+        floats1 = _mm_mul_ps(floats1, floats2);
+        sum = _mm_add_ps(sum, floats1);
+        offset += 4;
+      }
+    } else {
+      // Use unaligned load.
+      sum = _mm_loadu_ps(u);
+      __m128 floats2 = _mm_loadu_ps(v);
+      // Multiply.
+      sum = _mm_mul_ps(sum, floats2);
+      while (offset <= max_offset) {
+        __m128 floats1 = _mm_loadu_ps(u + offset);
+        floats2 = _mm_loadu_ps(v + offset);
+        floats1 = _mm_mul_ps(floats1, floats2);
+        sum = _mm_add_ps(sum, floats1);
+        offset += 4;
+      }
+    }
   }
-  return total;
+  // Add the 4 sums in sum horizontally.
+#if 0
+	alignas(32) float tmp[4];
+	_mm_store_ps(tmp, sum);
+	float result = tmp[0] + tmp[1] + tmp[2] + tmp[3];
+#else
+  __m128 zero = _mm_setzero_ps();
+  // https://www.felixcloutier.com/x86/haddps
+  sum = _mm_hadd_ps(sum, zero);
+  sum = _mm_hadd_ps(sum, zero);
+  // Extract the low result.
+  float result = _mm_cvtss_f32(sum);
+#endif
+  // Add on any left-over products.
+  while (offset < n) {
+    result += u[offset] * v[offset];
+    ++offset;
+  }
+  return result;
 }
 #else
 double DotProductSSE(const double *u, const double *v, int n) {
@@ -48,7 +99,8 @@ double DotProductSSE(const double *u, const double *v, int n) {
   if (offset <= max_offset) {
     offset = 2;
     // Aligned load is reputedly faster but requires 16 byte aligned input.
-    if ((reinterpret_cast<uintptr_t>(u) & 15) == 0 && (reinterpret_cast<uintptr_t>(v) & 15) == 0) {
+    if ((reinterpret_cast<uintptr_t>(u) & 15) == 0 &&
+        (reinterpret_cast<uintptr_t>(v) & 15) == 0) {
       // Use aligned load.
       sum = _mm_load_pd(u);
       __m128d floats2 = _mm_load_pd(v);
--- a/src/arch/intsimdmatrix.h
+++ b/src/arch/intsimdmatrix.h
@@ -115,7 +115,7 @@ struct TESS_API IntSimdMatrix {
   static const IntSimdMatrix *intSimdMatrix;
   // Only available with NEON.
   static const IntSimdMatrix intSimdMatrixNEON;
-  // Only available with AVX2 / SSE.
+  // Only available with AVX2 / AVX / FMA / SSE.
   static const IntSimdMatrix intSimdMatrixAVX2;
   static const IntSimdMatrix intSimdMatrixSSE;
 };
--- a/src/arch/intsimdmatrixavx2.cpp
+++ b/src/arch/intsimdmatrixavx2.cpp
@@ -86,54 +86,45 @@ static inline __m128i load64_to_128(const int8_t *wi_) {
 }
 
 #if defined(FAST_FLOAT)
-static inline void ExtractResults8(__m256i result, const int8_t *wi, const float *scales,
-                                   float *v) {
-  __m128i w128 = load64_to_128(wi);          // 8x8bit vals in bottom of 128bit reg
+
+static inline void ExtractResults8(__m256i result, const int8_t *wi,
+                                   const float *scales, float *v) {
+  __m128i w128 = load64_to_128(wi); // 8x8bit vals in bottom of 128bit reg
   __m256i w256 = _mm256_cvtepi8_epi32(w128); // 8x32bit vals in 256bit reg
   __m256i bias_scale = _mm256_set_epi32(127, 127, 127, 127, 127, 127, 127, 127);
-  __m256d scale01234567 = _mm256_loadu_ps(scales);
-  //~ __m256d scale4567 = _mm256_loadu_ps(scales + 8);
+  __m256 scale01234567 = _mm256_loadu_ps(scales);
   w256 = _mm256_mullo_epi32(w256, bias_scale); // 8x32 <bias * 127>
   result = _mm256_add_epi32(result, w256);     // result += bias * 127
-  __m256 res01234567 = _mm256_cvtepi32_ps(_mm256_castsi256_si128(result));
+  __m256 res01234567 = _mm256_cvtepi32_ps(result);
   result = _mm256_permute4x64_epi64(result, 2 + (3 << 2));
-  __m256d res4567 = _mm256_cvtepi32_pd(_mm256_castsi256_si128(result));
-  res01234567 = _mm256_mul_pd(res01234567, scale01234567);
-  //~ res4567 = _mm256_mul_pd(res4567, scale4567);
+  res01234567 = _mm256_mul_ps(res01234567, scale01234567);
   _mm256_storeu_ps(v, res01234567);
-  //~ _mm256_storeu_pd(v + 4, res4567);
 }
 
-static inline void ExtractResults16(__m256i result0, __m256i result1, const int8_t *&wi,
-                                    const float *&scales, float *&v) {
+static inline void ExtractResults16(__m256i result0, __m256i result1,
+                                    const int8_t *&wi, const float *&scales,
+                                    float *&v) {
   __m128i w8 = _mm_loadu_si128(reinterpret_cast<const __m128i *>(wi));
   // 8x8bit vals in bottom of 128bit reg
-  const __m256i bias_scale = _mm256_set_epi32(127, 127, 127, 127, 127, 127, 127, 127);
+  const __m256i bias_scale =
+      _mm256_set_epi32(127, 127, 127, 127, 127, 127, 127, 127);
   __m256i w256 = _mm256_cvtepi8_epi32(w8); // 8x32bit vals in 256bit reg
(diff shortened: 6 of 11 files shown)


stweil:
Using `float` instead of `double` requires a build with `-DFAST_FLOAT`.

While floating point calculations should be faster with `float`, it also has some overhead because (de-)serialization requires additional steps as long as the traineddata files store double values.

The new code passes most unit tests (`lstm_test` still fails). I am still checking whether it is faster or not. My first test with `lstm_squashed_test` was not as expected because the execution time even increased a little.

egorpugin:
What is t in tfloat? type?
Maybe `using tfloat = ...` (lowercase name)?

stweil:
> What is t in tfloat?

Tesseract float?

stweil:
I could use `TessFloat` instead of `TFloat` ...

egorpugin:
Primitive typedefs could be just snake_case?

nocun:
I am really curious what the results are. Personally I do not have any experience with `float` vs. `double` speed comparison, Alexandrescu in his talk mentions that the two should be similar ([see slide 22](https://www.slideshare.net/andreialexandrescu1/three-optimization-tips-for-c)).

There is also the issue of type mismatch in some places (for example [simdetect.cpp:97](https://github.com/tesseract-ocr/tesseract/pull/3486/files#diff-c27c6eb89f41325812b7587bba780d89ad5fc230f87ddb7f51590d4971b73552R97)). In some cases compiler will still operate on `double` and then convert back to `float`.

Although after checking on CompilerExplorer it looks like some `float` with `double` comparisons (like `-1.0`, `0.0`, `1.0`) do not involve a promotion conversion (at least on x86).

src/arch/simddetect.cpp
```diff
@@ -84,16 +84,16 @@ bool SIMDDetect::sse_available_;
...
   for (int k = 0; k < n; ++k) {
     total += u[k] * v[k];
   }
   return total;
 }
 
 // Compute dot product using std::inner_product.
-static double DotProductStdInnerProduct(const double *u, const double *v, int n) {
+static TFloat DotProductStdInnerProduct(const TFloat *u, const TFloat *v, int n) {
   return std::inner_product(u, u + n, v, 0.0);
```
nocun, line 97:
```suggestion
  return std::inner_product(u, u + n, v, TFloat(0.0));
```

stweil: Right, that would largely improve this function for the `float` case.

src/arch/dotproductavx.cpp
```diff
@@ -29,6 +29,47 @@ namespace tesseract {
 
 // Computes and returns the dot product of the n-vectors u and v.
 // Uses Intel AVX intrinsics to access the SIMD instruction set.
+#if defined(FAST_FLOAT)
+float DotProductAVX(const float *u, const float *v, int n) {
```
amitdo, line 33: FAST_FLOAT16 -> FAST_FLOAT ?

Not just in this line, but also in other places in this file.

stweil: I used `FAST_FLOAT16` up to now exclusively in `dotproductavx.cpp` to disable some code which handled 16 (instead of 8) float values per iteration and which had a negative impact on the performance.

GerHobbelt:
See also #3490 -- just a rough first run on medium hardware (laptop i7) looks like this shaves off a at-least-noticable amount. Haven't got time to dive in more right now and run VTune or anything, but even 10% off on the DotProduct would be great as that's eating upwards of 90% off total product run-time on my rigs. More floats multiplied per clock is a big win then IMO.

(#3490 has preliminary code for AVX2, etc. fitted for TFloat. When I file a proper pullreq and you like it, then someone with more AVX/AVX2 savvy should peer-review that hacky edit. 😊 )

stweil:
> More floats multiplied per clock is a big win then IMO.

Yes, the DotProduct function (for AVX / AVX2 / SSE / NEON / native / ...) uses most of the processor time with best models, especially during training runs.

Ideally using `float` (4 bytes) instead of `double` (8 bytes) data types could double the speed if the calculation is limited by the memory bandwidth and when SIMD instructions can handle for example 8 `float` operations instead of 4 `double` operations in the same time. Getting 10 % off is already a nice improvement, but I still hope to get nearer to 50 % (maybe 30 or 40 %) off.

amitdo:
Once we have this float 32 option ready, next step could be OpenMP offloading, to calculate the dot product in GPU with just a few lines of code.

stweil:
Do you have code examples for that?

stweil:
One issue with the current Tesseract code is the mix of the two implementations "fast" and "best" in the same classes. It results in lots of places which check for `int_mode`, and the classes waste memory because they contain data for both implementations although only one is used at a time.

Ideally "fast" and "best" should use separate classes, and the "best" classes should be template classes which allow `float` and `double`. Then the Tesseract code could support "fast" and two variants of "best" without recompilation.

amitdo:
Regarding OpenMP offloading, here is one source:
https://developer.ibm.com/technologies/systems/articles/gpu-programming-with-openmp/#ref2

GerHobbelt:
Re the OpenMP bit: might we be better off using BLAS (or BLIS?) primitives with a OpenCL/etc. supporting library and thus leave the whole microtuning of math operations on multiple platforms to someone else? 

I know it's not going to be as simple as that, but my current concern is that *I* at least am not running off the usual gcc+linux rigs for this (MSVC+Windows mainly, on not always the quickest hardware), nor are my users, so something like clBLAS (OpenCL+BLAS) looks more promising to me than OpenMP right now (OpenMP is at spec 5.something and MSVC2019 is at.... 2.0 (/openMP compiler flag)? Plus I don't see where MSVC has that alleged "GPU support" in their OpenMP implementation, so... 🤔 🤔 

BLAS/BLIS comes with SDOT, DDOT, etc. plus several BLAS libs out there mention AMD+ARM (OpenCL) support, next to ubiquitous NVidia. At least the way I read their pages.

Meanwhile OpenMP `#pragma omp` promises fewer lines of code than the entire setup and cleanup I see in the usual OpenCL & CUDA samples. BLAS libs *look* like they take care of this under the hood, but I haven't looked deep enough to sure yet.

Anyway, this is new ground for me, so disregard or correct if I'm saying dumb things here. 🍼

GerHobbelt:
Ok, some performance number for FAST_FLOAT:

# Is FAST_FLOAT useful?

TL;DR: *YES*.

Here's the results from a MSVC2019 (latest) + VTune (latest) set of profile runs:

```

DOUBLE / REL 64

Function                        CPU Time        Module  Function (Full) Source File     Start Address
tesseract::DotProductSSE        4.661s  tesseract-unittests.exe tesseract::DotProductSSE(double const *,double const *,int)     dotproductsse.cpp       0x140101b40
DotProductNativeTF              2.224s  tesseract-unittests.exe DotProductNativeTF(double const *,double const *,int)   main.cpp        0x1400270f0
tesseract::DotProductFMA        2.027s  tesseract-unittests.exe tesseract::DotProductFMA(double const *,double const *,int)     dotproductfma.cpp       0x140101860
DotProductDBL                   1.952s  tesseract-unittests.exe DotProductDBL(double const *,double const *,int)        main.cpp        0x1400270a0
tesseract::DotProductNative     1.883s  tesseract-unittests.exe tesseract::DotProductNative(double const *,double const *,int)  dotproduct.cpp  0x140101ca0
tesseract::DotProductAVX        1.660s  tesseract-unittests.exe tesseract::DotProductAVX(double const *,double const *,int)     dotproductavx.cpp       0x1401019d0
_stdio_common_vfprintf          0.597s  ucrtbase.dll    _stdio_common_vfprintf  [Unknown]       0x180022930



FLOAT / REL 64

Function / Call Stack           CPU Time        Module  Function (Full) Source File     Start Address
tesseract::DotProductSSE        2.173s  tesseract-unittests.exe tesseract::DotProductSSE(float const *,float const *,int)       dotproductsse.cpp       0x140101b30
tesseract::DotProductNative     1.817s  tesseract-unittests.exe tesseract::DotProductNative(float const *,float const *,int)    dotproduct.cpp  0x140101c90
DotProductDBL                   1.687s  tesseract-unittests.exe DotProductDBL(float const *,float const *,int)  main.cpp        0x1400271f0
tesseract::DotProductAVX_8      1.244s  tesseract-unittests.exe tesseract::DotProductAVX_8(float const *,float const *,int)     main.cpp        0x1400270a0
DotProductNativeTF              0.991s  tesseract-unittests.exe DotProductNativeTF(float const *,float const *,int)     main.cpp        0x140027260
tesseract::DotProductAVX_A      0.881s  tesseract-unittests.exe tesseract::DotProductAVX_A(float const *,float const *,int)     main.cpp        0x140027120
tesseract::DotProductFMA        0.856s  tesseract-unittests.exe tesseract::DotProductFMA(float const *,float const *,int)       dotproductfma.cpp       0x1401017d0
tesseract::DotProductAVX        0.846s  tesseract-unittests.exe tesseract::DotProductAVX(float const *,float const *,int)       dotproductavx.cpp       0x140101980
_stdio_common_vfprintf          0.626s  ucrtbase.dll    _stdio_common_vfprintf  [Unknown]       0x180022930
```

What's this say: DotProduct of ANY KIND will be about 50% faster, so @stweil was on the money. 😄 

This is using #3490 + fixes (to be posted later, running out of time again - RL), i.e. FAST_FLOAT, done pervasively as coded by bleeding edge commits from @stweil + all AVX/SEE/FMA dorproduct calls ported as well (that's in #3490 and will be re-submitted in a cleaner pullreq later)

REL64 = Windows 10, 64bit build, using MSVC with AVX2 and other Release Build optimizations enabled.


For comparison, as an aside, here are my Debug Build numbers: 64-bit MSVC build, but with all debug flags ON: much slower, but the same relative speed up:

```
FLOAT 

Function / Call Stack           CPU Time        Module  Function (Full) Source File     Start Address
tesseract::DotProductNative     16.672s tesseract-unittests.exe tesseract::DotProductNative(float const *,float const *,int)    dotproduct.cpp  0x14019cc60
DotProductDBL                   15.519s tesseract-unittests.exe DotProductDBL(float const *,float const *,int)  main.cpp        0x1400186e0
DotProductNativeTF              15.406s tesseract-unittests.exe DotProductNativeTF(float const *,float const *,int)     main.cpp        0x140018760
tesseract::DotProductSSE        8.211s  tesseract-unittests.exe tesseract::DotProductSSE(float const *,float const *,int)       dotproductsse.cpp       0x14019c870
tesseract::DotProductAVX        5.571s  tesseract-unittests.exe tesseract::DotProductAVX(float const *,float const *,int)       dotproductavx.cpp       0x14019c4a0
tesseract::DotProductAVX_8      5.224s  tesseract-unittests.exe tesseract::DotProductAVX_8(float const *,float const *,int)     main.cpp        0x140017e70
tesseract::DotProductAVX_A      5.065s  tesseract-unittests.exe tesseract::DotProductAVX_A(float const *,float const *,int)     main.cpp        0x140018140
tesseract::DotProductFMA        4.022s  tesseract-unittests.exe tesseract::DotProductFMA(float const *,float const *,int)       dotproductfma.cpp       0x14019c110



DOUBLE:

Function / Call Stack           CPU Time        Module  Function (Full) Source File     Start Address
DotProductDBL                   30.756s tesseract-unittests.exe DotProductDBL(double const *,double const *,int)        main.cpp        0x140017e70
DotProductNativeTF              17.317s tesseract-unittests.exe DotProductNativeTF(double const *,double const *,int)   main.cpp        0x140017ef0
tesseract::DotProductSSE        15.845s tesseract-unittests.exe tesseract::DotProductSSE(double const *,double const *,int)     dotproductsse.cpp       0x14019bf50
tesseract::DotProductAVX        11.916s tesseract-unittests.exe tesseract::DotProductAVX(double const *,double const *,int)     dotproductavx.cpp       0x14019bbc0
tesseract::DotProductFMA        7.616s  tesseract-unittests.exe tesseract::DotProductFMA(double const *,double const *,int)     dotproductfma.cpp       0x14019b870
```


## What are the AVX_A and AVX_8 versions? And TF vs DBL?

DBL is Native, done with all double's.
TF is a correction to the commits from @stweil as I merged them yesterday: all-float calculus in Native (there was still a single `double` in there for the sum).

Aside: resulting values differ in the 6-7th digit between SEE/AVX/FMA and Native versions. There's also that same inaccuracy visible in double vs. float so output changes ever so slightly. (Makes sense given IEEE754, anyway.)

AVX_A is DotProductAVX optimized for ALIGNED LOADS vs unaligned (`load` vs. `loadu`, as is already in original tesseract SSE code)

AVX_8 is like ~FAST_FLOAT16, i.e. 8 floats per round, regular AVX is 16 floats per round (I took out the FLOAT16 conditional logic there).

Do note that the ALIGNED code is exercised only a few times, relatively speaking, due to the way the test/benchmark code has been hacked together (intentionally! as I was testing the stability of the code changes).

I don't have a good answer why the FMA code is so much faster than the rest. 🤔  -- At least the validation code in the test assures me that the output is the same value, so it's not due to bugs "helping" that one.

Here's the test/benchmark code snippet used, so you can see what was done, globally speaking:

```
bool approx_eq(double a, double b)
{
	auto diff = a - b;
	if (diff == 0.0) return true;
	// take the log of both, as we know all incoming values will be positive,
	// so we can check the precision easily, i.e. at which significant digit
	// did the difference occur? ::
	a = log(a);
	b = log(b);
	diff = a - b;
	return (diff >= -1e-5 && diff <= 1e-5);
}

void testDP(void)
{
#define AMOUNT   655360
#define STEP	 1       // offset jump between tests; determines (indirectly) the number of memory-aligned tests vs. UNaligned tests executed. (STEP=4 would be 16-byte step @ float, i.e. each test aligned)
	static TFloat arr1[AMOUNT];
	static TFloat arr2[AMOUNT];

	for (int i = 0; i < AMOUNT; i++)
	{
		arr1[i] = 2.0 + i * 0.005;
		arr2[i] = 7.0 + i * 0.005;
	}

	TFloat total[10];
	const int size = 8192;     // size of vector(s) to feed the dot-product functions

	// start with aligned storage, then step through, including unaligned storage accesses
	int round = 0;
	for (int step = 0; step + size < AMOUNT; (step += STEP), round++)
	{
		TFloat* p1 = arr1 + step;
		TFloat* p2 = arr2 + step;

		total[0] = DotProductDBL(p1, p2, size);
		const double sollwert = total[0];
		//total[1] = DotProduct(p1, p2, size);
		total[1] = DotProductNative(p1, p2, size);
		total[2] = DotProductNativeTF(p1, p2, size);
		total[3] = DotProductAVX(p1, p2, size);
		total[4] = DotProductFMA(p1, p2, size);
		total[5] = DotProductSSE(p1, p2, size);
#if defined(FAST_FLOAT)
		total[6] = DotProductAVX_8(p1, p2, size);
		total[7] = DotProductAVX_A(p1, p2, size);
		total[8] = sollwert;             // regrettably, my i7 doesn't have AVX512 support.
		//total[8] = DotProductAVX_512(p1, p2, size);
#else
		total[6] = sollwert;
		total[7] = sollwert;
		total[8] = sollwert;
#endif
		// check calculations: this is always active to ensure we haven't got any obvious bugs in there (we got a few! -- fixed!)
		for (int i = 1; i <= 8; i++)
		{
			if (!approx_eq(total[i], sollwert))
				fprintf(stderr, "step %d: total %d = %lf %s\n", step, i, (double)total[i], approx_eq(total[i], sollwert) ? "(PASS)" : "*****FAIL!*****");
		}
		// only load the console output for information once in a while: we're running a benchmark here...
		if (round % 10000 == 0)
		{
			for (int i = 1; i <= 8; i++)
			{
				fprintf(stderr, "step %d: total %d = %lf %s\n", step, i, (double)total[i], approx_eq(total[i], sollwert) ? "(PASS)" : "*****FAIL!*****");
			}
		}
	}

	exit(7);
}

```

GerHobbelt:
Oh, in case anyone wonders why DBL isn't the same speed everywhere: it's still TFloat vectors coming in, so only the *calculus* is done as all-double. Hence different numbers for the same func in DOUBLE and FLOAT scenarios.

GerHobbelt:
Latest @stweil results on MSVC2019 latest, /openmp:experimental (https://devblogs.microsoft.com/cppblog/improved-openmp-support-for-cpp-in-visual-studio/) on lightly loaded i7 4700 laptop (has AVX2, not AVX512):

image

TL;DR: `#pragma opm` (the DotProductNative() function) is on par with the hand-optimized AVX ones. The rest of the bunch are about the same.

The CPUtime:Self column has these 0 numbers because the compiler clearly decided to inline a few of these functions, but not all of them. (This does also affect your debugging experience in release builds, BTW: the inlined stuff is not steppable)

Sad discovery right now: Vtune doesn't cope well with templatized functions: it somehow bundles them together once they are inside another template instance (the benchmark runner in this case):

image

which not exactly how the benchmark is coded:

```
// The benchmark runner:
void run_tfloat_benchmark(void) {
  run_tfloat_benchmark<float>();
  run_tfloat_benchmark<double>();
}
```

Anyway, the diff between the double and float run is 2:1 and that matches the visual experience of the running benchmark quite well: the progress feedback via fprintf() in the double run feels about twice as slow as the float run part.

stweil:
> Is FAST_FLOAT useful?

My tests now also show a performance gain. I have run `lstm_squashed_test` on a number of different platforms. Most runs used the default implementation with `double`, but selected runs were done with `float`.

Results:

- double: fastest systems are AMD EPYC 7502 (32 s) and Mac mini (30 s)
- float: fastest systems are AMD EPYC 7502 (28 s) and Mac mini (22 s)

```
// Comparison of execution time with different dot product implementations.
// time lstm_squashed_test
// time DOTPRODUCT=accelerate lstm_squashed_test
// time DOTPRODUCT=generic lstm_squashed_test
// time DOTPRODUCT=native lstm_squashed_test
// Results for Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz (g++ 8.3.0)
// DotProduct (default)    53 s
// DotProductGeneric      105 s
// DotProductNative       139 s
// Results for Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz (clang 7)
// DotProduct (default)    47 s
// DotProductGeneric       99 s
// DotProductNative        55 s
// Results for AMD EPYC 7502 (g++ 10.2.1, single threaded):
// DotProduct (default)    36 s
// DotProductGeneric       91 s
// DotProductNative        87 s
// Results for AMD EPYC 7502 (g++ 10.2.1, single threaded, float):
// DotProduct (default)    28 s
// Results for AMD EPYC 7502 (clang 11, single threaded):
// DotProduct (default)    32 s
// DotProductGeneric       76 s
// DotProductNative        37 s
// Results for (g++ 10.2.1, single threaded):
// Results for Macbook Intel Core i5 2.4 MHz (clang):
// DotProduct (default)    60 s
// DotProductAccelerate    78 s
// DotProductGeneric      108 s
// DotProductNative        65 s
// Results for Apple M1 (clang):
// DotProduct (default)    60 s
// DotProductAccelerate    33 s
// DotProductGeneric       64 s
// DotProductNative        30 s
// Results for Apple M1 (clang, float):
// DotProduct (default)    61 s
// DotProductNative        22 s
// Results for Power8 3425 MHz (g++ 10.2.1, single threaded):
// DotProduct (default)   179 s
// DotProductGeneric      179 s
// DotProductNative       130 s
// Results for NVIDIA Jetson Xavier (g++ 9.3.0, single threaded):
// DotProduct (default)   172 s
// DotProductGeneric      180 s
// DotProductNative       179 s
// Results for NVIDIA Jetson Xavier (g++ 9.3.0, single threaded, float):
// DotProduct (default)   169 s
// DotProductNative       96 s
// Results for NVIDIA Jetson Xavier (clang 11, single threaded):
// DotProduct (default)   181 s
// DotProductGeneric      185 s
// DotProductNative       107 s
```

GerHobbelt:
@stweil: *very* nice set of hardware rigs you got there! 😄 

And some interesting results! 

From far away, it looks like the ranking of the 'Native' variant is highly dependent on which compiler (and its settings, probably) is used.

What I've seen on my dev rig so far is that MSVC2019 (v16.10.3) is quite smart in its optimizations:

- when I turned on the AVX512 compiler setting, without any other change in the code, I got immediate crashes in the Native DotProduct loops as MSVC apparently had decided this could do with some AVX512 opcodes (which my old i7/4700 doesn't support, hence the crashes) -- which makes me wonder about the actual *worth* of the compiler feature detection code in Tesseract, as that will only work when the compiler optimizations are turned OFF -- otherwise MSVC will decide on its own where to apply AVX/AVX2/AVX512 opcodes if the compiler switches allowed it at build time.

  Don't know what to do about that, without reverting to a separate "non-optimized compile" build for "old hardware". 🤔 
- Your change to `float` cuts time in about half. With AVX on top of that, you get another halving of the cost. That's for DotProduct at least, across the board. *Wow!*

Asides:
- I've created and run a benchmark for the matrix codes, but it turns out there that matrix::Init, i.e. the 2D fill of `GENERIC_2D_ARRAY` is eating up 50% or worse, no matter what I did. (last edit there: SHA-1: 648d5dfb85bf078cc60491da2b86fc9f0fc23f17) The matrix calculus work doesn't want to rise about ~10% cost in that mini-benchmark.
- Ran a PDF bulktest benchmark tonight, which is my own mupdf fork, which, as part of the bulk PDF processing, kicks up Tesseract (embedded in the mupdf code) to do OCR work on the rendered pages), so that serves as a reasonable check how well it'll do in actual practice & at user sites (design: one PDF per thread; bulktest runs the PDFs sequentially, so usually only 1 core is loaded; that will of course change when running multiple PDFs in parallel, which is what will happen with the end user app; bulktest is coded sequentially to keep log & diagnostics as simple as possible)
  + observed CPU cost share of DotProduct (float+AVX, so BEST): ~20% of *total* (including lots of PDF text/metadata activities, so it's not just 'tesseract' taking time in there)

     --> is the 50% gain useful in general practice/use? **definitely YES!**

     --> how much gain do we get? Bit hard to say without also measuring the old style, but given the new cost number, the gain at userland level would be estimated at 15-20% speed gain. (Doubling the time of `dpAVX` doesn't double its cost %)
  + Tesseract already had OpenMP multithread `#pragma` macros in the LSTM code, which make it pretty tough for VTune to produce any legible numbers. With `/openmp:experimental` turned on, the entire bulktest (thus LSTM + DotProduct as the costliest ones, next to mupdf draw) is noticably faster and loading all 4 cores, though still suboptimally if you care about that stuff. Weird loads reported in VCOMP dll, but that looks like it's the OpenMP support code racking up a *lot* of NtYieldExecution calls, but that's outside my control as that's MSVC runtime lib internals.
  + When I forcibly turn the OpenMP multithread macros OFF in LSTM, the bulktest run slows down noticably (so OpenMP::experimental in MSVC does indeed distribute the LSTM work across the cores) but the reported numbers become easier to decode: this is where I come to the conclusion that DotProductAVX takes up ~20% of the CPU time. The rest is mostly consumed by LSTM::Forward (~25%) and the weird one in the bunch now that I turned OFF OpenMP multithreading: the VCOMP NtYieldExecution stuff from MS is reported as 50% cost. 😨 Must be something wrong there or VTune going 🤡 .

---

Here's a snap how it looks with all OpenMP support dialed up to the max in MSVC2019: note the 4 cores at work when there's bits of LSTM to distribute (4 cores = 4 brown graphs!):

image

And here's that same report showing the DotProductAVX score and where it is called from (top-down view again -- the cryptic func identifiers are OpenMP/MSVC multithreading at work):

image

---

Here's a **second** bulktest run, now with the OpenMP multithread macros DISABLED in LSTM as mentioned above, so we get a single core load and easier to read numbers; also visible in the core graphs is the no-activity there: the red stuff is all VCOMP (Microsoft internals) spinning like mad waiting for OpenMP work that will never arrive):

image

(The rest of the cost (61%!) is all VCOMP Spinning/MS runtime internals ("SwitchToThread" in KERNELBASE.dll and children thereof.)

---

The way I read this, is that OpenMP is definitely *quite experimental* over at Microsoft -- this is the latest available compiler you're looking at. *Granted, on hardware that's a couple of years old, but that VCOMP/NtYield stuff feels a bit *yucky* IMO.*

On the other hand, it looks like the optimizations done by hand (AVX, SSE, etc. DotProduct and Matrix codes, float vs. double now) look pretty solid, at least on this hardware. That means my users will get faster OCR results -- still slow, but definitely faster! 👍 

(🤔 What you also MAY note here is that the matrix calls don't feature *anywhere* in these benchmarks. Was that a happy optimization once that's now not important anymore, or is the matrix code more involved with **training** tesseract?  Just asking as I'm a tesseract n00b. 😄 )

GerHobbelt:
@stweil: switched the code to using your `DotProductAVX1` instead of `DotProductAVX`, i.e. processing twice the FPs per round in the loop: that change drops the 19.4% cost number in bulktest down a *little bit*: to 18.7%. Not much, but still another about-half percent gain. So AVX1 is slightly better on this hardware (similar to other tests posted earlier today; depends a bit on the way the loops are coded, but anyway).

I wonder what it'll accomplish (if anything) on the other hardware+compiler rigs.

amitdo:
@stweil's data, formatted in a table.

https://github.com/amitdo/DotProduct/blob/main/dot-product-benchmark.md

stweil:
Switching from `double` to `float` for faster training has an unexpected side effect: it also seems to reduce the time for recognition with fast models significantly. While most computation with fast models uses integer, there is also a smaller amount of floating point arithmetics involved. I now have run a first test using one of the images which are part of our test repository: `tesseract test/testing/8087_054.3G.tif 8087_054.3G`. Running that 10 times takes 21.22 s with the default `double` and 19.94 s with the new code using `float`. So the saving is 1.28 s or 6 %. I think that's good news for all who do mass production of OCR like for example @MerlijnWajer. Of course this result still has to be confirmed with additional tests.

amitdo:
There are two sets of fast models. One in the tessdata repo and the second in the tessdata_fast repo.

The one in tessdata is MUCH faster than 'best', with a very minimal loss in accuracy. I believe it's suitable for mass digitization projects.

nagadomi:
`lstmtraining --stop_training --convert_to_int ...` command on this branch causes a crash.
```
LSTMRecognizer::Serialize(&mgr_, &fp):Error:Assert failed:in file src/training/unicharset/lstmtrainer.cpp, line 998
```

stweil:
> `lstmtraining --stop_training --convert_to_int ...` command on this branch causes a crash.

It works for me. Can you debug which part of `LSTMRecognizer::Serialize` fails?

nagadomi:
It seems that `WeightMatrix::Serialize` for `int_mode_==true` is not implemented.
https://github.com/stweil/tesseract/blob/a9e27a5d21834496e21e92e4d5213c7649209424/src/lstm/weightmatrix.cpp#L260-L262

stweil:
> It seems that `WeightMatrix::Serialize` for `int_mode_==true` is not implemented.

Right, my test was invalid and I got the same failure. This is fixed now.

nagadomi:
It worked. Thank you.

amitdo:
About the issue with f32 dot product for training. It might be related to:
https://en.wikipedia.org/wiki/Floating-point_arithmetic#Accuracy_problems

I suggest you try training with fma instead of avx. https://en.wikipedia.org/wiki/Multiply%E2%80%93accumulate_operation#In_floating-point_arithmetic

stweil:
The pull request is now ready for merging. I squashed my own commits and added three commits from @GerHobbelt which fixed my dot product implementations and added another one.

Still missing: documentation, continuous integration and more unittest code.

I want to add a configure parameter to enable `float` for builds using autotools. Would `--enable-float32` be a good choice? Or are there other suggestions?

How should `float` builds be made with cmake?

For continuous integration it might be sufficient to run some builds with `float` and others with `double`. I think that duplicating the number of builds would be too much.

egorpugin:
I was thinking about template parameter TFloat, not compiler definition.

stweil:
> I was thinking about template parameter TFloat, not compiler definition.

Yes, that should be addressed by future commits. But that also requires separating the classes for fast and best models.

egorpugin:
I elaborate a bit more.

IMO we should aim on real C++ with template metaprogramming.
Not compiler switches, but templated code.

Yes, this will involve more header only stuff, and other related things, burden.
But I see this as more powerful and preferable way.

stweil:
Using more templates is fine if it helps maintaining the code and does not cost performance and memory resources.

Here I have a very concrete need: I want to run more training, and LSTM training is currently very slow. So making it faster is important for me. Templates cannot be used for the low level SIMD code, and I think they cannot be used here as long as we don't have different objects for fast and best LSTM.

The changes suggested here don't prevent using template code later to make executables which support `double` and `float`, so users can select which variant they want to have.

amitdo:
@stweil and @GerHobbelt, thank you!

amitdo:
>I want to add a configure parameter to enable float for builds using autotools. Would --enable-float32 be a good choice? Or are there other suggestions?

Please add this option to the autotools build.

nagadomi:
I also agree that adding the autotools option is a good idea.
I used the following command to build and check it. But I'm not sure if this is the right way.
```
./configure CPPFLAGS=-DFAST_FLOAT --prefix ~/local
```

GerHobbelt:
That is (was, rather) my pullreq 7 on @stweil's fork. Should I reissue on
mainline now 3486 has been merged?

stweil:
> Please add this option to the autotools build.

With pull request #3510 it will be possible to use `./configure --enable-float32`.

stweil:
> Once we have this float 32 option ready, next step could be OpenMP offloading, to calculate the dot product in GPU with just a few lines of code.

I tried that meanwhile, it basically worked but took much more time. The faster calculation of the dot product does not help because before that the data for that calculation has to be transferred from the CPU RAM to the GPU memory.

So either the GPU code must begin at a higher level and transfer scales and weights only once initially from CPU to GPU memory, or one needs a GPU which shares memory with the CPU and avoids the memory transfer overhead.

amitdo:
Intel has 'DL Boost' in some Xeon CPUs released since 2019. It is also available in desktop CPUs released a few months ago. It supports `float32 += bfloat16 * bfloat16` for vectors.

PR #3491, opened 2021-07-13
Tfloat patch 2 - bugfix for FMA FAST_FLOAT

GerHobbelt:
See commit: just merged your latest this morning and saw you had the same mistake as I had in #3490: 8 floats fit in 256 bits, contrasting 4 doubles. This corrects that and should produce correct numbers for DotProduct.


GerHobbelt:

GerHobbelt:
Now filed correctly here: https://github.com/stweil/tesseract/pull/1

PR #3492, opened 2021-07-13
Tfloat patch 1 - TFloat used consistently in prototypes

GerHobbelt:
Just saw these this morning; anyway: question is: do we want the various *specific implementations* to have TFloat in their function definition (correctly matching the typedef) or do we want the compiler to match up float <-> TFloat and double <-> TFloat instead?

(if you decide on the latter direction, that would be using the C++ compiler and removing the FAST_TFLOAT #ifdef/#else conditions in the AVX/SSE codes, but it would also mean slightly more work for the compiler, so slightly longer compile times. 🤔  Just thinking out loud and wondering what your preferred coding style for these beasts would be, in the end.)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1236,6 +1236,7 @@ check_PROGRAMS += commandlineflags_test
 check_PROGRAMS += dawg_test
 endif # ENABLE_TRAINING
 check_PROGRAMS += denorm_test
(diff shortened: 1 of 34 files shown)


GerHobbelt:

GerHobbelt:
Reissued as https://github.com/stweil/tesseract/pull/2

PR #3493, opened 2021-07-13
Tfloat patch 3: implements DotProductSSE() for FAST_FLOAT

GerHobbelt:
extract from #3490. Tested to produce the correct numbers in the benchmark mentioned in #3486


GerHobbelt:

PR #3494, opened 2021-07-13
Tfloat patch 4: bugfixes for AVX2 FAST_FLOAT Extract8+16 implementations

GerHobbelt:
Extract from #3490 - bugfixing the AVX2 Extract8+16 codes, where there's lines like `__m256d scale01234567 = _mm256_loadu_ps(scales)`, i.e. loading float vectors into double vector types.  Extract from #3490.

Note: next pullreq is a reduced version of this: less code duplication for bleeding edge tfloat branch.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1236,6 +1236,7 @@ check_PROGRAMS += commandlineflags_test
 check_PROGRAMS += dawg_test
 endif # ENABLE_TRAINING
 check_PROGRAMS += denorm_test
+check_PROGRAMS += dotproduct_test
 if !DISABLED_LEGACY_ENGINE
 check_PROGRAMS += equationdetect_test
 endif # !DISABLED_LEGACY_ENGINE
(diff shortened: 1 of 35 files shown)


egorpugin:
Hi,

You are doing patches wrong.

GerHobbelt:
Note: #3495 is this one (#3494) PLUS FAST_FLOAT condition only applied to the ExtractXYZ calls, as the others are good to go with only their prototype adjusted from double --> TFloat. Hence #3495 is only moving code compared to this one, no code change. (I don't know what diff tools you use, but thus this one (#3494) would be easier to diff/review, and then verify that #3495 is only copy/cut/paste work, resulting in a much larger diff)

GerHobbelt:
> Hi,
> 
> You are doing patches wrong.

Crap. Yep, seen it. 😊 

Discard. Will re-issue.

GerHobbelt:
🤔  I used the github link and didn't watch carefully that the bugger ref'd against against instead of tfloat.  Checked against my own visual commit graph and these were correct, but definitely wholly wrong to submit against master. *ugh*.

GerHobbelt:
Re-issued as https://github.com/stweil/tesseract/pull/4.

PR #3500, opened 2021-07-16
CMake: fix cross-build to iOS/tvOS/watchOS

SpaceIm:
On these OS, executables are bundle, and they need a BUNDLE DESTINATION, otherwise CMake configuration fails.
See https://cmake.org/cmake/help/latest/policy/CMP0006.html


egorpugin:
But you do not provide BUNDLE DESTINATION in this PR.

SpaceIm:
I provide a common destination for all types (runtime, bundle etc).

stweil:
Thanks.

PR #3501, opened 2021-07-17
Solve clang reporting unused variable in  ExtractMicros function

zdenop:
```
..\src\classify\mf.cpp(47,14): warning: unused variable 'f' [-Wunused-variable]
  for (auto &f : features) {
             ^
```


zdenop:
I do not understand why there are 4 commits instead of one...

zdenop:
so msvc does not understand it...

stweil:
> I do not understand why there are 4 commits instead of one...

I squashed them for the merge, so now it's only a single commit.

zdenop:
thanks

PR #3509, opened 2021-07-25
Add unittest for static TessBaseAPI object

stweil:


stweil:
The new unittest would have detected a regression introduced by pull request #3330.

PR #3516, opened 2021-08-06
Remove submodule abseil

stweil:


stweil:
@egorpugin, the CI builds fail since I added dotproductneon.cpp because cmake builds try to compile that file also on Intel platforms. Do I have to use conditional compilation in the file, or is it possible to exclude it from builds on non ARM hosts?

amitdo:
https://github.com/google/googletest

>### Coming Soon
>
>We are planning to take a dependency on Abseil.

So it will become an indirect dependency.

PR #3517, opened 2021-08-07
Write image filename in ALTO output and reduce size of renderer classes

stweil:


stweil:
This fixes issue #2700.

PR #3539, opened 2021-08-24
Use model prefix also for submodels

stweil:
Fix also a regression in the for loop which handles submodels.

Fixes: 0d91c700c062 ("Modernize code in Tesseract::init_tesseract")

### DIFF
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -23,6 +23,8 @@
 #  include "config_auto.h"
 #endif
 
+#include <regex> // for std::regex_match
+
 #include "control.h"
 #include "matchdefs.h"
 #include "pageres.h"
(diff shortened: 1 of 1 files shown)


stweil:
Some language models load submodels (using `tessedit_load_sublangs` in their config), for example `jpn` loads `jpn_vert`.

If such models are installed in a subdirectory (for example `best`), they can be accessed by using that subdirectory name as a prefix (`best/jpn`). Then the submodels should use the same prefix (`best/jpn_vert`) automatically.

This is achieved by this pull request.

src/ccmain/tessedit.cpp
```diff
@@ -247,6 +249,16 @@ static bool IsStrInList(const std::string &str, const std::vector<std::string> &
 void Tesseract::ParseLanguageString(const std::string &lang_str, std::vector<std::string> *to_load,
                                     std::vector<std::string> *not_to_load) {
   std::string remains(lang_str);
+  // Look whether the model file uses a prefix which must be applied to
+  // included model files as well.
+  std::regex e("(.*)/[^/]*");
+  std::cmatch cm;
+  std::string prefix;
+  std::regex_match(lang.c_str(), cm, e, std::regex_constants::match_default);
+  if (cm.size() == 2) {
```
egorpugin, line 258: `if (regex_match(...`

stweil: Thanks, I changed that now.

src/ccmain/tessedit.cpp
```diff
@@ -23,6 +23,8 @@
 #  include "config_auto.h"
 #endif
 
+#include <regex> // for std::regex_match
```
egorpugin, line 26: Std headers must go last.

Issue #3544, opened 2021-08-30
Obtaining all pages on tessinput.tif when using "get.images" configuration.


diogoalmiro:
### Environment
Tested on windows 10 - 64 with:
```
tesseract v5.0.0-alpha.20210506
 leptonica-1.78.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.3) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
 Found libarchive 3.5.0 zlib/1.2.11 liblzma/5.2.3 bz2lib/1.0.6 liblz4/1.7.5 libzstd/1.4.5
 Found libcurl/7.77.0-DEV Schannel zlib/1.2.11 zstd/1.4.5 libidn2/2.0.4 nghttp2/1.31.0
```
and on a docker alpine container:
```
tesseract 4.1.1
 leptonica-1.81.1
  libgif 5.2.1 : libjpeg 8d (libjpeg-turbo 2.1.0) : libpng 1.6.37 : libtiff 4.2.0 : zlib 1.2.11 : libwebp 1.2.0
 Found AVX2
 Found AVX
 Found FMA
 Found SSE
```

### Current Behavior:
When running `tesseract multiple-page.tif - get.images` tesseract overwrites `tessinput.tif` until it stops. Efectively only allowing to see the last page processed.

### Expected Behavior:
Either create a multiple page `tessinput.tif` or creating `tessinput-n.tif` for each page.

amitdo:
You face a similar issue if you ocr more than one single page. In this case the solution should be to change `tessinput.tif` to `original-input-name-bw.tif`.

diogoalmiro:
@amitdo I think it is acceptable to keep the name `tessinput.tif`. Once tesseract ends, I move the file to a more appropriate one.
The main problem is when running on a multiple page we can't know when to move the file, neither will tesseract "wait" for us to move the file.

diogoalmiro:
Maybe a better solution would be to write to OUTPUTBASE.tessinput.tif simmilarly to any other config file and on this note also change the logfile to OUTPUTBASE.tesseract.log for consistency.

> CONFIGFILE
> 
[...]

PR #3545, opened 2021-09-01
Rename processed TIFF output file and add page number if needed (fixe…

stweil:
…s issue #3544)

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1276,7 +1276,12 @@ bool TessBaseAPI::ProcessPage(Pix *pix, int page_index, const char *filename,
 
   if (tesseract_->tessedit_write_images) {
     Pix *page_pix = GetThresholdedImage();
-    pixWrite("tessinput.tif", page_pix, IFF_TIFF_G4);
+    std::string output_filename = output_file_ + ".processed";
+    if (page_index > 0) {
+      output_filename += std::to_string(page_index);
+    }
+    output_filename += ".tif";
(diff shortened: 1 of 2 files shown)


stweil:
Example:

```
ls -1 test/testing/4*tif >/tmp/files
tesseract /tmp/files /tmp/out get.images
```

This will create `out.processed.tif`, `out.processed1.tif` and `out.processed2.tif`

src/api/baseapi.cpp
```diff
@@ -1276,7 +1276,12 @@ bool TessBaseAPI::ProcessPage(Pix *pix, int page_index, const char *filename,
 
   if (tesseract_->tessedit_write_images) {
     Pix *page_pix = GetThresholdedImage();
-    pixWrite("tessinput.tif", page_pix, IFF_TIFF_G4);
+    std::string output_filename = output_file_ + ".processed";
```
amitdo, line 1279: This file is the result of thresholding, without other image preprocessing. 

I suggest this form:

`std::string output_filename = output_file_ + "-binary";`
or
`std::string output_filename = output_file_ + "-bw";`
or
`std::string output_filename = output_file_ + "-thresholded";`

I prefer `-` over `.`. IMO, the dot as file extension should only be used for standard extensions, not for custom, application-specific file extension.

stweil: The current documentation says "Write processed input images to file". Can we be sure that there won't be other kinds of preprocessing in the future? You already added new thresholding methods.

Using `-` instead of `.` would be fine for me if others agree, too,

Issue #3564, opened 2021-09-16
Failed to open tr file: shapeclustering. Segmentation fault (core dumped)


Ortega-Dan:
------------------------

### Environment

* **Tesseract Version**: tesseract 5.0.0-beta-20210815-6-g1d3d
* **Commit Number**: unknown
* **Platform**: "Linux danort-pc 5.4.0-84-generic 94-Ubuntu SMP Thu Aug 26 20:27:37 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux"

### Current Behavior:
Segmentation fault (core dumped) on shapeclustering execution:
```bash
:~$ shapeclustering -F font_properties -U unicharset -O "separados.unicharset" separados.test.tr

Reading shapeclustering ...
Failed to open tr file: shapeclustering
Reading separados.test.tr ...
Segmentation fault (core dumped)
```

### Expected Behavior:
Run without issues and produce a "shapetable" file

### Suggested Fix:
Unknown

Also reported by someone else on Stack Overflow: https://stackoverflow.com/questions/67613545/tesseract-failed-to-open-tr-file-issue
and it works fine on Tesseract v5.0.0-alpha.20200328

(Attachments: files generated before the shapeclustering command [while not really specifically needed to reproduce the issue])
[FilesGeneratedAndNeededUpToTheFailureStep.zip](https://github.com/tesseract-ocr/tesseract/files/7178968/FilesGeneratedAndNeededUpToTheFailureStep.zip)

zdenop:
So you are trying to train (legacy) tesseract 3 engine with tesseract 5? 
Do you know and understand what you are doing?

zdenop:
OK. Official training doc for tesseract 3 is here:
https://github.com/tesseract-ocr/tessdoc/blob/main/tess3/Training-Tesseract-3.03%E2%80%933.05.md

Your training seems to fail in the first step: filename convention. Try to fix it, rerun training, and come back with the result (if it does not help.)

Ortega-Dan:
> and [this comment](https://github.com/tesseract-ocr/tesseract/issues/3547#issuecomment-912339918) seems to be relevant too.

That comment says "so unless Buyi993-bit tries to train for tesseract 3, this part of docs should be relevant:" then... that part is not relevant because Yes, I am trying to train for legacy tessreact which is usable and works very well for custom trainings on tesseract 5.

Again, it even works on the versión mentioned in the last comment.

[...]

PR #3566, opened 2021-09-17
Simplify function LoadTrainingData and fix mastertrainer_test

stweil:


stweil:
This is a better fix for issue #3564 which does not break `mastertrainer_test`.

Issue #1256, opened 2018-01-02
can not link tesseract built with cmake to my app


avestnik:
I am working on large research project and need to run the same app on different platforms, so I have to use cmake. the problem is I fail to link tesseract with my app, it complains about missing tesseract lib.

I built up virtual machine with ubuntu 17.10 to demonstrate the problem
```
root@tess:~/w/test/build# cmake ..
-- The C compiler identification is GNU 7.2.0
... [cmake output skipped]
-- Build files have been written to: /root/w/test/build
root@tess:~/w/test/build# make
Scanning dependencies of target testtess
[ 50%] Building CXX object CMakeFiles/testtess.dir/main.cpp.o
[100%] Linking CXX executable testtess
/usr/bin/ld: cannot find -ltesseract
collect2: error: ld returned 1 exit status
CMakeFiles/testtess.dir/build.make:94: recipe for target 'testtess' failed
make[2]: *** [testtess] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/testtess.dir/all' failed
make[1]: *** [CMakeFiles/testtess.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```

tesseract built from most recent source with cmake and installed to ~/local

app is quite simple:
main.cpp
```
#include <iostream>
#include <tesseract/baseapi.h>

int main() {
    tesseract::TessBaseAPI tess;
    tess.Init("/tmp/tessdata", "eng");

    std::cout << "Hello, World!" << std::endl;
    return 0;
}
```

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.9)
project(testtess)

set(CMAKE_CXX_STANDARD 11)

set(CMAKE_PREFIX_PATH $ENV{HOME}/local)

find_package(Tesseract CONFIG REQUIRED)
include_directories(${Tesseract_INCLUDE_DIRS})

add_executable(testtess main.cpp)

target_link_libraries(testtess ${Tesseract_LIBRARIES})
```

I guess the problem lies in descrepancies between TesseractConfig.cmake which uses tessaract for the library name
```
set(Tesseract_LIBRARIES tesseract)
```
and TesseractTargets-noconfig.cmake which uses libtesseract
```
set_property(TARGET libtesseract APPEND PROPERTY IMPORTED_CONFIGURATIONS NOCONFIG)
set_target_properties(libtesseract PROPERTIES
```

[...]

PR #3568, opened 2021-09-17
Fix linking Tesseract to project using CMake on Linux

Robyer:
In the past Tesseract library was wrongly named as "liblibtesseract" so it was needed to use "-llibtesseract" as argument for linker. The name was fixed in commit 52cac3a42ef2b823b9367b9ee2d1d84f8eda4ea0 but the ${Tesseract_LIBRARIES} variable still wrongly holds "libtesseract" and causes linker error when one is using it in target_link_libraries(...). This commit fixes that.

### DIFF
--- a/cmake/templates/TesseractConfig.cmake.in
+++ b/cmake/templates/TesseractConfig.cmake.in
@@ -23,6 +23,6 @@ include(${CMAKE_CURRENT_LIST_DIR}/TesseractTargets.cmake)
 @PACKAGE_INIT@
 
 set_and_check(Tesseract_INCLUDE_DIRS "@PACKAGE_INCLUDE_DIR@")
-set(Tesseract_LIBRARIES libtesseract)
+set(Tesseract_LIBRARIES tesseract)
 
 check_required_components(Tesseract)


Robyer:
For more context you can see discussion on issue #1256

zdenop:
I am afraid this is not the correct fix to the problem. On Windows I have to use something like this (to use  `${Tesseract_LIBRARIES}` for linking because the library name is `tesseract50`):

```
if (WIN32)
    target_link_libraries(${PROJECT_NAME} 
	   ${Tesseract_LIBRARIES}${Tesseract_VERSION_MAJOR}${Tesseract_VERSION_MINOR}
	   ${Leptonica_LIBRARIES})  #
endif() 
```

So instead of using a plain library name, maybe `OUTPUT_NAME` should be used here somehow. I tried to follow leptonica cmake settings, but I was not successful. So help of cmake expert is need. 

Another problem is why leptonica (as tesseract dependacy) is not automatically added and needs to be included manually.

However, this works (and it adds leptonica automatically):
```
    target_link_libraries(${PROJECT_NAME} Tesseract::libtesseract) 
```

egorpugin:
everyone must use:
```
    target_link_libraries(${PROJECT_NAME} Tesseract::libtesseract) 
```

Robyer:
@zdenop Shouldn't that `if` condition be in the modified file then? If it is used for both Linux and Windows and on both platforms the library has different name? That would solve it.

amitdo:
About "**They** use..."
https://github.com/DanBloomberg/leptonica/commits/master/cmake/templates/LeptonicaConfig.cmake.in
:-)

Robyer:
@amitdo Hmm :-)

@zdenop I created PR to use the correct value also on WIN32, see #3578

PR #3570, opened 2021-09-20
Fix wrong tiles parameters in Sauvola

amitdo:
Thanks to Robert Sachunsky @bertsky who pointed out the issue.


amitdo:
Adapted from:

https://github.com/DanBloomberg/leptonica/blob/bc21d9f184ce6d9e28e0d37ea4ad38cdbce08297/src/binarize.c#L484

https://github.com/DanBloomberg/leptonica/blob/bc21d9f184ce6d9e28e0d37ea4ad38cdbce08297/src/binarize.c#L885-L889

amitdo:
The CodedQL runner failed at **build time**.

Issue #3326, opened 2021-03-14
hOCR renderer could perhaps write scan_res to the ocr_page metadata


MerlijnWajer:
### Environment

* **Tesseract Version**: 5.0.0-alpha-20201231

### Current Behavior:

I've found that Tesseract will scale the `x_fsize` if the DPI is passed at the command line (via `--dpi`) or if the DPI if embedded in the metadata of the image itself (exif tags, etc). However, it will not do this when it attempts to estimate the DPI. All of this is fine and makes sense to me, but for a consumer of the hOCR file, the actual DPI that was used to scale the x_fsize is nowhere to be found. That is: Tesseract writes the `x_fsize` in printer points, but doesn't specify what it thought the DPI of the image was.

### Expected Behavior:

Tesseract provides an entry in the hOCR files that specifies the identified DPI that was used to scale various values in the hOCR file.

### Suggested Fix:

The hOCR spec has a `scan_res` field for the `ocr_page` entry: http://kba.cloud/hocr-spec/1.2/#scan_res

Using this `scan_res` field, consumers of the hOCR output file can construct the size of the text in pixels.

I could conjure up a patch to include this property in the hOCR output of every `ocr_page` element, but *only* if it was detected in the image metadata, or passed via the command line. In other cases, Tesseract will guess what the DPI is, but does not actually adjust the `x_fsize` attribute.

MerlijnWajer:
I currently went for this. Not sure what to do with the x_res.

```
From 9f82e08f84e74737d2b8a13f6b8fbc2105aef11b Mon Sep 17 00:00:00 2001
From: Merlijn Wajer <merlijn@wizzup.org>
Date: Mon, 15 Mar 2021 19:31:25 +0100
Subject: [PATCH] hocrrenderer: write scan_res

---
 src/api/hocrrenderer.cpp | 8 +++++---
 1 file changed, 5 insertions(+), 3 deletions(-)

diff --git a/src/api/hocrrenderer.cpp b/src/api/hocrrenderer.cpp
index f4504da9..738de926 100644
--- a/src/api/hocrrenderer.cpp
+++ b/src/api/hocrrenderer.cpp
@@ -175,9 +175,11 @@ char* TessBaseAPI::GetHOCRText(ETEXT_DESC* monitor, int page_number) {
   } else {
     hocr_str << "unknown";
   }
-  hocr_str << "\"; bbox " << rect_left_ << " " << rect_top_ << " "
[...]

PR #3571, opened 2021-09-21
hocrrenderer: write scan_res property to the ocr_page

MerlijnWajer:
This will make Tesseract emit the DPI of the document, if known at OCR
time. This is requird to properly interpret the x_fsize (font size)
property of words, since Tesseract scales the font size to the DPI.

See issue #3326 (https://github.com/tesseract-ocr/tesseract/issues/3326)

### DIFF
--- a/src/api/hocrrenderer.cpp
+++ b/src/api/hocrrenderer.cpp
@@ -173,8 +173,11 @@ char *TessBaseAPI::GetHOCRText(ETEXT_DESC *monitor, int page_number) {
   } else {
     hocr_str << "unknown";
   }
+
   hocr_str << "\"; bbox " << rect_left_ << " " << rect_top_ << " " << rect_width_ << " "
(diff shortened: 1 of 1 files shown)


src/api/hocrrenderer.cpp
```diff
@@ -26,6 +26,7 @@
 #endif
 #include <tesseract/renderer.h>
 #include "tesseractclass.h" // for Tesseract
+#include "thresholder.h" // for ImageThresholder
```
amitdo, line 29: Please use the API instead:

https://github.com/tesseract-ocr/tesseract/blob/60fd2b4abaa9c5c5c42d32db57576bc95d28a78a/src/api/baseapi.cpp#L956-L960

MerlijnWajer: Done, thanks.

amitdo:
Thanks!

PR #3577, opened 2021-09-24
Add configurable variables to control thresholding

amitdo:


amitdo:
@egorpugin,  sw build for Ubuntu is broken.

src/ccmain/tesseractclass.cpp
```diff
@@ -79,6 +79,30 @@ Tesseract::Tesseract()
                  "Thresholding "
                  "method: 0 = Otsu, 1 = Adaptive Otsu, 2 = Sauvola",
                  this->params())
+    , INT_MEMBER(thresholding_sauvola_window_half_width, 25,
```
bertsky, line 82: Like I said, 50 or 100 would be more appropriate for 300 DPI input. 

Also, like I explained, it would make more sense to tie this to the DPI (`estimated_res_`) by some (tunable) factor. Say: `thresholding_sauvola_window_dpi_factor` (defaulting to 0.33 or 0.66; I would use the full width instead of the half width for the param).

src/ccmain/tesseractclass.cpp
```diff
@@ -79,6 +79,30 @@ Tesseract::Tesseract()
...
+                 this->params())
+    , double_MEMBER(thresholding_sauvola_kfactor, 0.40,
+                 "Factor for reducing threshold due to variance. Must be 0 or positive number",
+                 this->params())
+    , INT_MEMBER(thresholding_adaptive_otsu_tile_x, 300,
+                 "Desired tile width. Actual size may vary. Must be >= 16",
+                 this->params())
+    , INT_MEMBER(thresholding_adaptive_otsu_tile_y, 300,
+                 "Desired tile height. Actual size may vary. Must be >= 16",
+                 this->params())
```
bertsky, line 93: As I said, it's not a good encapsulation of complexity for the user to think about absolute tile sizes. This depends so much on DPI and on RAM size. Rather, it should be the number of tiles which the full image is split into (`thresholding_num_tiles`?) or the relative tile size (`thresholding_tile_size_inch`?).

src/ccmain/tesseractclass.cpp
```diff
@@ -79,6 +79,30 @@ Tesseract::Tesseract()
...
+    , INT_MEMBER(thresholding_adaptive_otsu_smooth_x, 0,
+                 "Half-width of convolution kernel applied to threshold array: "
+                 "Use 0 for no smoothing",
+                 this->params())
+    , INT_MEMBER(thresholding_adaptive_otsu_smooth_y, 0,
+                 "Half-width of convolution kernel applied to threshold array: "
+                 "Use 0 for no smoothing",
+                 this->params())
```
bertsky, line 101: In the interest of simplicity, let's not distinguish x and y here, only a single value for both.

bertsky (commented):
Finally, I think there is one missing parameter that might be most useful practically: the threshold to actually apply (after computing the background level). It should default to 0.5 but 0.2 would yield more bg and 0.7 would yield more fg. (Perhaps not necessary if we have a decent auto-contrast, but this is difficult.)

egorpugin:
>  build for Ubuntu is broken.

I've built sw on fedora34 which has newer libc.
I'll check it.

---

As for the changes.
I said earlier that tess should have an api taking binary image for OCR.
It will be much more flexible than tuning dozens of parameters of tess.
User will provide his fine-tuned b/w input image.

bertsky:
> I said earlier that tess should have an api taking binary image for OCR.
> It will be much more flexible than tuning dozens of parameters of tess.
> User will provide his fine-tuned b/w input image.

We should offer both IMO – see [my proposal](https://github.com/tesseract-ocr/tesseract/issues/3083#issuecomment-918145451) for something like `SetThresholdedImage` / `SetBinaryImage` / 2-ary `SetImage` and `GetEstimatedResolution`. @amitdo, IIUC you wanted to keep such changes out of 5.0?

src/ccmain/tesseractclass.cpp
amitdo: I prefer to mirror Leptonica's parameters. If Leptonica provides them both it means they can have different values.

I also tried to use similar (but not identical) names for the configurable parameters to Leptonica's variables names.

bertsky: Of course you can smooth differently in x and y direction, but this will not be useful or necessary 99% of the time.

I strongly recommend _not_ just mimicing Leptonica parameters here: this is an implementation detail. Leptonica's job is to provide an implementation for certain mathematical operations (as a backend), Tesseract's is to do practical document processing with that (as a frontend). You want to _hide_ details and complexity – this will actually give the users _more_ leverage.

src/ccmain/tesseractclass.cpp
amitdo: The calculation to estimate the resolution is done in the layout analysis stage. The result is not available in the thresholding stage.

https://github.com/tesseract-ocr/tesseract/blob/60fd2b4abaa9c5c5c42d32db57576bc95d28a78a/src/ccmain/pagesegmain.cpp#L320-L328

bertsky: > The calculation to estimate the resolution is done in the layout analysis stage. The result is not available in the thresholding stage.

Yes, I forgot about that. Well, for binarization, I guess it's good enough to use the DPI from input meta (plus the plausibilization and fallback to 72)...

src/ccmain/tesseractclass.cpp
amitdo: >As I said, it's not a good encapsulation of complexity for the user to think about absolute tile sizes.

But that's what other implementations do.

I want to keep the tiles size and window size as absolute values, like the other implementations do.

Nevertheless, I plan to provide an option to turn on a bool parameter to scale these absolute numbers depending on the dpi info which is embedded in the image (or user provided dpi that Tesseract already supports).

With this option turn on, the absolute values will be related to the base dpi of 300x300. If the actual dpi is 600x600 the tiles and the window will be scaled up.

amitdo:
This PR will be useful to many advanced users.

Later PR will provide the scaling option I mentioned in another comment.

I also want to add some more thresholding  methods from Leptonica.

I plan to add the option to pick an external grayscale-binary-thresholds_array trio. This will be used by even more advanced users.
This one is not guaranteed to be implemented in time for 5.0.0.

amitdo:
I need to rethink about the binarization done by Tesserct.

src/ccmain/tesseractclass.cpp
bertsky: > But that's what other implementations do.
> 
> I want to keep the tiles size and window size as absolute values, like the other implementations do.

It would help naming a few of those implementations. Normally I would agree and see that as a strong argument, but in this case I believe we are comparing implementations of different tasks – binarization or other low-level tools on the one hand, and robust layout analysis or other high-level tools on the other.

> Nevertheless, I plan to provide an option to turn on a bool parameter to scale these absolute numbers depending on the dpi info which is embedded in the image (or user provided dpi that Tesseract already supports).
> 
> With this option turned on, the absolute values will be related to the base dpi of 300x300. If the actual dpi is 600x600 the tiles and the window will be scaled up.

That would be a good compromise, the crux would be to make the parameter names and their description readily understandable.

bertsky:
> This PR will be useful for many advanced users.

Agreed in principle, but I still think the concrete choices for parameter names and defaults should be different, as argued above.

> Later PR will provide the scaling option I mentioned in another comment.
> I also want to add some more thresholding methods from Leptonica.
> I plan to add the option to use an external grayscale-binary-thresholds_array trio. This will be used by even more advanced users.
> This one is not guaranteed to be implemented in time for 5.0.0.

Then why close this one so soon?

egorpugin:
@amitdo 

The most flexible approach is to allow user to pass his prepared image.
Ocr tools does ocr. Let others do their best (segmentation of text blocks, binarization).

amitdo:
>Ocr tools does ocr. Let others do their best (**segmentation of text blocks**, binarization).

A large part of Tesseract is dedicated to layout analysis.

amitdo:
@bertsky,

>Then why close this one so soon?

One reason is Egor's comment. Why should I bother with this PR with this consistent feedback?

zdenop:
@amitdo : You can not expect that everybody will praise you or be excited about your work. We have different aims, we use different tools, but we share what we find useful. We need consensus only in cases where one approach is blocking/limit something else. IMO this is not the case.

Also, I prefer a ready solution (even not perfect) than waiting for some great new feature which is not available. Because it is easier to remove old code than to use not existing code. ;-)

So I suggest finishing this PR, so the community can test it and just it is useful for them too.

amitdo:
Zdenko, very nice comment.

zdenop:
@egorpugin 

> I said earlier that tess should have an api taking binary image for OCR.

AFAIK user can give binary image for OCR via [SetImage](https://github.com/tesseract-ocr/tesseract/blob/ebb214c4438ccfa32952a4907a27295f1f46be60/src/api/baseapi.cpp#L617) or do I miss something? 

> Let others do their best (segmentation of text blocks, binarization).

Can you suggest another maintained C++ opensource library/tool for document analyzing? 
Because I got different feedback: it was claimed that tesseract is the best open-source c++ document analysis tool they found. Actually, it was told to me that they use tesseract for document analyze and for OCR (e.g. numbers on invoices) they use a custom OCR engine. Even there was a request to put tesseract document analyze to separate library, so they do not need to install OCR part...

amitdo:
I'm going to merge this PR as-is today.

The scaling request can be done later on top of the absolute values. I don't promise I'll implement it.

amitdo:
>AFAIK user can give binary image for OCR via SetImage or do I miss something?

See @bertsky comment here:
https://github.com/tesseract-ocr/tesseract/issues/3083#issue-684853460

egorpugin:
@zdenop 

> AFAIK user can give binary image for OCR via SetImage or do I miss something?

If SetImage does not do further thresholding, it's ok.

> it was claimed that tesseract is the best open-source c++ document analysis tool they found.

Consider following.
Tess has one algorithm. Some images will work fine, others are not.
Now we take all other possible algorithms. Their count is infinite.
Obviously, Inf > 1.
What approach is more flexible?

And I'm not saying about something existing.
Someony may always do their own, custom image preprocessing.

@amitdo 

I'm not saying these changes are bad or unneeded. It is completely fine to have them.

src/ccmain/tesseractclass.cpp
amitdo: I changed `smooth_x`, `smooth_y` to just `smooth` as suggested.

src/ccmain/tesseractclass.cpp
amitdo: I change the name to `window_size` and the value to `51`.

amitdo:
Please don't merge yet.

src/ccmain/tesseractclass.cpp
bertsky: Ok, then let's postpone the exploitation of DPI to a later time, together with new API to set bilevel and background level images and to query the estimated DPI.

bertsky:
@egorpugin @zdenop you are both right I think: Tesseract is a good practical OLR tool which covers many different materials and languages more or less robustly and very fast, too. There are weaknesses for sure, like complex layouts with lots of tables or graphical elements or widely varying font sizes, historical documents with faint ink or show-through or very dense typesetting or ornaments, images with bad photography like shadows or keystone distortion. A page frame detection is missing entirely, and OSD and deskewing are not nearly as good as the rest. For some of these problems, you'll find good automatic solutions, and often very good interactive/GUI tools. But still, Tesseract's automatic OLR is still unmatched for productive use in the Open Source world. I'm certain this will soon change when more and more well packaged and well tested data-driven OLR tools pour in from the research breakthrough that the deep learning / object detection methods brought into the field. But Tesseract with its flexible API and wide coverage of languages and scripts will still be relevant for OCR workflows. And its OLR will always be an important baseline, perhaps even an adjunct to the neural models in some kind of hybrid approach.

That said, again, I believe Tesseract should try to do its best by its own means, and things like binarization and table OLR can still be improved (and at the same time made more user-friendly). But its API should also be kept as flexible as necessary to better interoperate with other preprocessing, OLR, and OCR tools in a custom workflow.

src/ccmain/tesseractclass.cpp
amitdo: @bertsky,

After further thinking, I decided to change the code to use relative values as you suggested. I will send a new PR.

amitdo: Apart from `window_size` and `tile_size` what other parameters need to be scaled using the DPI info? `smooth_size`? Anything else?

bertsky: > Apart from `window_size` and `tile_size` what other parameters need to be scaled using the DPI info? `smooth_size`? Anything else?

Yes, the size of the smoothing kernel would make sense for that, too. (Perhaps the parameter could use typographic `pt` – 1/72" – as unit.) 

Not sure about `tile_size` though – it could still be just a divisor of the full image size (`num_tiles` or sth)...

The others should not depend on pixel density.

amitdo: I don't understand the `num_tiles` suggestion. It implies that the actual size is meaningless and the only thing we care about is RAM. BTW, I don't let users set `tile_size` for Sauvola, so this is only relevant for Otsu. 

I also don't understand `pt` for smoothing.

Bottom line: Is it reasonable to go with what I was thinking to implement for relative values and use floats for all the three parameters with the formula: `scaled_size = relative_size * y_resolution` ?

Example to parameters values:
`0.2` for `window_size`,  `1.0` for `tile_size`, `0.01` for `smooth_size`.

This is consistent and it will be easy to me to explain in the parameters' description string.

Also, I can tweak the smooth parameter name:
`smooth_size` -> `smooth_block_size`?   `smooth_kernel_size`?

Another acceptable option is that I'll pass this whole task to you. You'll do 100% what you want by sending a new PR in the next few days. Let me know if you want to accept this offer.

bertsky: > I don't understand the `num_tiles` suggestion. It implies that the actual size is meaningless and the only thing we care about is RAM. BTW, I don't let users set `tile_size` for Sauvola, so this is only relevant for Otsu.

Sorry, I had confused `pixOtsuAdaptiveThreshold`'s tiles with something else I saw in Olena. You are right – these do behave much like Sauvola's windows, so they should be parameterized in the same, DPI-relative way.

> I also don't understand `pt` for smoothing.

Smoothing is done via kernels of (usually) quite small px sizes, so you'd have to use very small float numbers to express this as DPI factor or multiples of 1". Typographic points, on the other hand, are in the right range and pretty easily understandable for a user. (For example, if I know I have a 12pt font, I can assume a 4pt filter will suffice keeping bg around the textlines coherent.) 

But it's just an idea. Using resolution/DPI is probably more transparent, because that estimate/metadatum might be way off, so the user will understand that parameters based on that might be way off too. Hiding behind _derived_ physical measures like point or inch could obscure that.

> Bottom line: Is it reasonable to go with what I was thinking to implement for relative values and use floats for all the three parameters with the formula: `scaled_size = relative_size * y_resolution` ?
> 
> Example to parameters values: `0.2` for `window_size`, `1.0` for `tile_size`, `0.01` for `smooth_size`.
> 
> This is consistent and it will be easy to me to explain in the parameters' description string.

Sure, why not. That way they are all the consistent (and the default values speak for themselves).

> Also, I can tweak the smooth parameter name: `smooth_size` -> `smooth_block_size`? `smooth_kernel_size`?
> 

Yeah, `smooth_kernel_size` sounds better to me.

> Another acceptable option is that I'll pass this whole task to you. You'll do 100% what you want by sending a new PR in the next few days. Let me know if you want to accept this offer.

No, by all means, please go ahead (I must still find time to unwrap my changes and branches and finally get the ALTO thing finished)!

amitdo: Okay.

Another question. Currently the default for smooth is 0 - no smoothing. That what Leptonica do in its Otsu example program. Should I change the default to >0? 0.01? Something else?

bertsky: > Currently the default for smooth is 0 - no smoothing. That what Leptonica do in its Otsu example program. Should I change the default to >0? 0.01? Something else?

You mean in `prog/otsutest2`? There are non-zero examples in `prog/binarize_fuzzer` (30px) and `prog/binarize_set` (2px) though. I have not used that algorithm myself in practice yet. So I cannot recommend a default.

Could you try a few settings on a realistic example, against various tile sizes?

amitdo: >Could you try a few settings on a realistic example, against various tile sizes?

I think I'll choose some defaults and pass the testing task to the community. If necessary, the default can be changed later on.

amitdo: @bertsky, thanks for your feedback.

amitdo: @bertsky,

I keep changing my mind about thresholding support in Tesseract. Your comment on testing made me rethink about supporting Leptonica's Otsu ('LeptOtsu').

My last thought on this subject:

I think I'll drop LeptOtsu. I added LeptOtsu, because I thought It can outperform the built-in Otsu. Currently I don't have a proof that It can achieve this goal. This makes me feel uncomfortable with keeping the code that use LeptOtsu. Remember that we are in second beta and we will probably move soon to rc and to the final release.

Leptonica's Sauvola is already known to be a good alternative to Tesseract's built-in Otsu, so I'll keep it.

bertsky: > thanks for your feedback.

Thank you for considering it and being so thorough!

> I think I'll drop LeptOtsu. I added LeptOtsu, because I thought It can outperform the built-in Otsu. Currently I don't have a proof that It can achieve this goal. This makes me feel uncomfortable with keeping the code that use LeptOtsu. Remember that we are in second beta and we will probably move soon to rc and to the final release.

But @wollmers already did a [first test](https://github.com/tesseract-ocr/tesseract/issues/3083#issuecomment-877663240) – and that showed your LeptOtsu to be superior not only to TessOtsu but to LeptSauvola (on the example at hand), while also being fastest.

Moreover, this merely wraps existing functionality in Leptonica. I'd like to think that the Leptonica creators had good reasons to provide this implementation besides Sauvola.

Also, this is just an option, nothing will break or degrade by integrating it. In the very least, your PR will lead to a better understanding and more evidence in the field.

I very much recommend keeping it included (while also making it as easy and effective to use as possible, hence the request for complexity-reducing parameters).

amitdo: >But @wollmers already did a first test – and that showed your LeptOtsu to be superior not only to TessOtsu but to LeptSauvola (on the example at hand), while also being fastest.

https://github.com/tesseract-ocr/tesseract/issues/3083#issuecomment-877663240

>Sauvola is an improvement compared to (legacy) global binarisation. 
>...
>Adaptive Otsu (Leptonica) is a problem on pages with large areas of empty but noisy background, e. g. title pages.

@wollmers

How many pages were included in your tests? There are only 4 samples in the folder you linked to.

Also, if you want to add your opinion, you are invited to do so.

wollmers: As I wrote in the original test report:

> It's a trial and not a systematic test.

There are two folders

https://github.com/wollmers/ocr-tess-issues/tree/main/issues/issue_3083_binarisation

3 historical pages known to have problems.

https://github.com/wollmers/ocr-tess-issues/tree/main/issues/issue_3083_binarisation_grey

4 examples of grey corner cases

Conclusion:

> Legacy method is not so bad and should be kept in Tesseract as long as the other methods are not better in all cases.

> **Some behaviours of the binarisation methods of Leptonica look like issues, maybe parameters, algorithm or implementation.**

In some systematic performance comparisons e.g. [Novel Adaptive Binarization Method for Degraded Document Images](https://www.techscience.com/cmc/v67n3/41584) Adaptive Otsu is under the top performers:

Bildschirmfoto 2021-10-03 um 01 35 11

Maybe there is an implementation difference between Adaptive Otsu (moving window) and Leptonica Otsu (tiled), because Leptonica Otsu creates quadratic blocks in larger areas of empty paper, either emphasizing noise and shine through, or creating black blocks on clean white areas (and white blocks in black areas). This is a bug, because a clean white area in the window should not become black. But this has no impact on CER, because such areas are not classified as text. It's just ugly.

Thus my vote is: **keep all 3 methods** and finish this step of configurable variables with defaults based on 300 DPI (or relative to the real or estimated DPI). We need more systematic tests to explore better or smarter parameters.

Read a lot of more recent papers about binarisation, ground truth for it and quality measurements for binarisation  without ground truth. Some sound promising, but also sound slower.

PR #3578, opened 2021-09-24
cmake: Improve configuration

Robyer:
- Use correct library name in TesseractConfig.cmake on all platforms
- Expose Tesseract_VERSION and Tesseract_VERSION_* variables in TesseractConfig.cmake


zdenop:
thanks!

PR #3595, opened 2021-10-14
Thresholding: Change the window & tile size params to relative numbers

amitdo:
They are relative to the pixel density of the image.


amitdo:
Cc: @bertsky

Issue #3590, opened 2021-10-04
makebox doesn't output horizontal coordinates of textangle 90 content


rmast:
### Environment
```
tesseract 4.1.1 and 5.0.0-beta-20210916
Linux 5.4.0-81-generic #91-Ubuntu SMP Thu Jul 15 19:09:17 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

using nld language.
the language nld (and eng) from https://github.com/tesseract-ocr/tessdata with these sizes:
15400601 eng.traineddata
8903736 nld.traineddata

Makebox does output this for a part in the image that has vertically oriented text (textangle 90):
(all horizontal coordinates and widths are 0)
```
2 1968 0 1982 0 0
0 1985 0 1998 0 0
8 2001 0 2014 0 0
4 2016 0 2030 0 0
- 2041 0 2049 0 0
2 2059 0 2073 0 0
/ 2074 0 2082 0 0
2 2083 0 2097 0 0
2 2116 0 2130 0 0
5 2133 0 2146 0 0
1 2150 0 2158 0 0
9 2165 0 2179 0 0
8 2181 0 2195 0 0
0 2197 0 2211 0 0
```

This is the image that was used for this data:
[210913.nog.2-000na.zip](https://github.com/tesseract-ocr/tesseract/files/7280732/210913.nog.2-000na.zip)

A similar issue was filed earlier https://github.com/tesseract-ocr/tesseract/issues/2340, but the issuer https://github.com/dev884 didn't provide any pointer to his fix, he has no code at all in his account.

### Expected Behavior:
I would expect the horizontal coordinates to resemble the ones in the word oriented hocr-output of the same region of the picture.
[...]

PR #3599, opened 2021-10-18
Repair boundingbox of individual characters of textangle 90 text

rmast:
Solution to issue #3590 (makebox doesn't output horizontal coordinates of textangle 90 content).

I followed these lines back to 2010, there has been no-one fiddling with these lines, however they were most suspect of excluding RIL_SYMBOL from the matrix transformation at textangle 90.

TBOX rotate operations don't seem expensive, so it's not known why the exclusion for RIL_SYMBOL has ever been introduced.

### DIFF
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -41,6 +41,7 @@
 #endif
 
 #if defined(HAVE_AVX) || defined(HAVE_AVX2) || defined(HAVE_FMA) || defined(HAVE_SSE4_1)
+// See https://en.wikipedia.org/wiki/CPUID.
 #  define HAS_CPUID
 #endif
 
@@ -94,6 +95,7 @@ bool SIMDDetect::avx_available_;
 bool SIMDDetect::avx2_available_;
 bool SIMDDetect::avx512F_available_;
 bool SIMDDetect::avx512BW_available_;
+bool SIMDDetect::avx512VNNI_available_;
 // If true, then FMA has been detected.
 bool SIMDDetect::fma_available_;
 // If true, then SSe4.1 has been detected.
@@ -171,6 +173,7 @@ SIMDDetect::SIMDDetect() {
         avx2_available_ = (ebx & 0x00000020) != 0;
         avx512F_available_ = (ebx & 0x00010000) != 0;
         avx512BW_available_ = (ebx & 0x40000000) != 0;
+        avx512VNNI_available_ = (ecx & 0x00000800) != 0;
       }
 #      endif
     }
@@ -201,6 +204,7 @@ SIMDDetect::SIMDDetect() {
         avx2_available_ = (cpuInfo[1] & 0x00000020) != 0;
         avx512F_available_ = (cpuInfo[1] & 0x00010000) != 0;
         avx512BW_available_ = (cpuInfo[1] & 0x40000000) != 0;
+        avx512VNNI_available_ = (cpuInfo[2] & 0x00000800) != 0;
       }
 #      endif
     }
--- a/src/arch/simddetect.h
+++ b/src/arch/simddetect.h
@@ -47,6 +47,10 @@ public:
   static inline bool IsAVX512BWAvailable() {
     return detector.avx512BW_available_;
   }
+  // Returns true if AVX512 Vector Neural Network Instructions are available.
+  static inline bool IsAVX512VNNIAvailable() {
+    return detector.avx512VNNI_available_;
+  }
   // Returns true if FMA is available on this system.
   static inline bool IsFMAAvailable() {
     return detector.fma_available_;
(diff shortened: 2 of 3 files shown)


p12tic:
@stweil @zdenop Is there anything that could be done to get this PR merged? I can create a separate PR rebased on top of latest main branch tip.

I have looked into all call paths that could potentially be affected by the code change. There aren't too many of them in the first place and all of them are within the auxiliary functionality of getting the results of tesseract out, not the recognition itself. As a consequence, the risk of a regression is probably relatively small.

The following is the list of APIs affected:
 - PageIterator::BoundingBoxInternal when level == RIL_SYMBOL
 - PageIterator::BoundingBox when level == RIL_SYMBOL
 - PageIterator::GetImage when level == RIL_SYMBOL
 - PageIterator::GetBinaryImage when level == RIL_SYMBOL
 - TessBaseAPI::GetComponentImages when level == RIL_SYMBOL
 - TessBaseAPI::GetConnectedComponents 
 - TessBaseAPI::GetBoxText
 - output of hOCR renderer when hocr_char_boxes is enabled
 - output of BoxText renderer

zdenop:
I am sorry, but I have a minimum spare time for tesseract. PR seems to be interesting, but as this effect API, it should be well tested including effect on training.

p12tic:
@wollmers This is great to hear. Is there any way to help? I could translate _very_ high-level directions into working code :-) Answering a small number of questions should take much less time than doing the implementation.

To me it seems that annotating ground truth images with correct bounding boxes is work that is not complicated in principle, but just needs a lot of effort. This would be a perfect task for an external developer like me to accomplish.

I'm assuming that you don't want to go the route of rendering text and OCRing the result images back, like when doing LSTM training in certain cases. In this case the character positions are essentially already known. Well, at least that's my understanding which could be completely wrong.

wollmers:
@p12tic 

> To me it seems that annotating ground truth images with correct bounding boxes is work that is not complicated in principle, but just needs a lot of effort for automation and reviewing. This would be a perfect task for an external developer like me to accomplish.

Sorry, mismatched this PR with [PR 3787](https://github.com/tesseract-ocr/tesseract/pull/3787). For 3787 (normal text without rotation) I wrote an approach at the weekend. See [ocr-bbox-gt](https://github.com/wollmers/ocr-bbox-gt) in prototypish Perl (without the dependencies, not published yet). If you can read it you can port it to your favourite language, which is maybe Python.

For text angles other than 0 degrees, the text image can be rotated before OCR and the bounding boxes geometrically transformed back. For degrees other than a multiple of 90 a polygon notation is needed, something like `x1,y1 x2,y2 x3,y3`.

Just use a clean image of text, which has no recognition errors (CER 0.0). That's the case for the sample image in 3787. Then use a legacy model with `--oem 0` which provides nearly perfect bounding boxes. 

Now we can check the quality as follows:

- count the width (and height) per character in a first scan of the bboxes
- select the width with the highest frequency ("best width")
- compare the best width against the actual width (`abs(width1 - width2)`) 
- count deviations per deviation 

Of course this works only with clean, generated images in one and the same font, style and size. But we want to isolate the problem, reduce it only to bbox errors, thus want to exclude all other seasons for errors.

As text one page of the Human Rights Declaration (available in ~500 languages) can be used. Format it with a popular font, export as PDF, pdftoimage, tesseract. That's the work to get ground truth. Then measure the errors compared between ground truth, before patch, after patch.

With legacy only a few characters have deviation:

```
$ tesseract pr_3787.png pr_3787.oem0.psm6.lat.png -l t5data/lat --oem 0 --psm 6 
	--tessdata-dir  /usr/local/share/tessdata makebox hocr txt pdf

$charboxes: 1706

width frequency per character
[...] # all other characters have only one width
r 15: 88,  16: 3,  17: 2, 
s 15: 156, 
t 13: 107,  15: 10,  14: 7,  12: 3,  16: 3, 
u 24: 138,  25: 5, 
v 24: 20,  25: 1,

width errors
    exact: 1653 (0.9689) # color green
    in   : 50   (0.0293) # within +/-2; color orange
    out  : 3    (0.0018) # 3 't' with width 16; color red
```
One of the 3 errors (deviation 3 pixels):

Bildschirmfoto 2022-05-02 um 11 02 03

It would be easy to correct this few remaining errors in a website (import the bboxes as JSON and wite the corrections back). Then the resulting bbox file is the ground truth.

With CTC/LTSM Tesseract release 5.1.0 it looks like this:

```
$ tesseract pr_3787.png pr_3787.psm6.png -l deu  --psm 6 
	--tessdata-dir  /usr/local/share/tessdata makebox hocr txt pdf

width errors
    exact: 1304 (0.7644)
    in   : 89   (0.0522)
    out  : 314  (0.1841)
```

The same part of the image with CTC/LTSM:

Bildschirmfoto 2022-05-02 um 11 05 33

rmast:
@wollmers wrote 

> Sorry, mismatched this PR with [PR 3787](https://github.com/tesseract-ocr/tesseract/pull/3787). 

Yes, I first wasn't able to understand what my rotation fix had to do with your response, but as I now also have run into a bounding-box issue I'll be glad trying your PR to see if that fixes it. I'll first see whether I can satisfactorily get it running with LSTM before reverting to OEM 0 for my bounding boxes. My fix doesn't fix straight up bounding boxes.

Issue #3604, opened 2021-10-22
Inappropriate string causes abnormal loop.


junsik-kim0:
### Environment

* **Tesseract Version**: 5.0.0-beta-20210916
* **Commit Number**: 
* **Platform**: Ubuntu 18.04

### Current Behavior:
A issue was found within the tesseract. The issue was found with a fuzzer based on the test-code"DoesNotConvertSpace". DoesNotConvertSpace API doesn't seem to be a public api. But this issue is related to utf8_step which it seems to be public api. I gave input string, utf8_step return 2. but real string size is 1.

```
std::string StringRenderer::ConvertFullwidthLatinToBasicLatin(const std::string &str) {
   std::string half_str;
   UNICHAR::const_iterator it_end = UNICHAR::end(str.c_str(), str.length());                   it_end : 0x77aa051                              
   for (UNICHAR::const_iterator it = UNICHAR::begin(str.c_str(), str.length()); it != it_end; ++it) {            it: 0x7aa050, 0x7aa052..    
```

In the code above, the ++ operator changes the memory address as much as the step returned by utf8_step function.
But if there is a mismatch between the utf8_step return value and the real string size, the syntax will loop abnormally.

crash log
```
==30255==ERROR: AddressSanitizer: stack-overflow on address 0x7ffea5a16000 (pc 0x000000c21f17 bp 0x7ffea5a14150 sp 0x7ffea5a14080 T0)
    #0 0xc21f17 in tesseract::UNICHAR::utf8_step(char const*) /root/fuzz-test-generation/exp/tesseract/src/ccutil/unichar.cpp:155:48
    #1 0xc21f17 in tesseract::UNICHAR::const_iterator::operator*() const /root/fuzz-test-generation/exp/tesseract/src/ccutil/unichar.cpp:174:19
    #2 0x4ea753 in tesseract::StringRenderer::ConvertFullwidthLatinToBasicLatin(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /root/fuzz-test-generation/exp/tesseract/src/training/pango/stringrenderer.cpp:709:45
    #3 0x4d9df6 in main /ssd_array/utopia_paper/fuzz-test-generation/exp/tesseract/case7.cc:12:9 
```

Steps to reproduce:

  1. git clone https://github.com/tesseract-ocr/tesseract.git

[...]

PR #3623, opened 2021-11-03
Fix Inappropriate string causes abnormal loop

junsik-kim0:
Soultion to issue https://github.com/tesseract-ocr/tesseract/issues/3604 (Inappropriate string causes abnormal loop).

If increment operator(++) returns a value longer than the string length, the loop may behave abnormally.
So it has been modified to work only when the range is smaller than string length.

### DIFF
--- a/include/tesseract/unichar.h
+++ b/include/tesseract/unichar.h
@@ -143,8 +143,8 @@ public:
     friend bool operator!=(const CI &lhs, const CI &rhs) {
       return !(lhs == rhs);
     }
-    friend bool operator<=(const CI &lhs, const CI &rhs) {
-      return lhs.it_ <= rhs.it_;
+    friend bool operator<(const CI &lhs, const CI &rhs) {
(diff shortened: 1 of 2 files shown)


stweil (commented):
Did you test your code change? It looks wrong.

afosscontact:
Oops.. I just ran the make command, and it seems that the build was successful.
I'm sorry for not checking properly. I modified source code.

stweil:
Please check also your Git settings (`git config --global user.name`). Your commits currently show this:

    Author: afosscontact <afosscontact@gmail.com>

"afosscontact" does not look like a real name.

junsik-kim0:
I replaced my username with my real name.

amitdo:
stweil,

Should we merge this PR?

stweil:
No, these changes are not required to fix a real issue, and they would hide problems with bad strings. I therefore closed the issue and close the pull request as well.

Issue #3620, opened 2021-11-01
CI unittest-disablelegacy fails since some time


stweil:
There was a build issue (missing include statements) which is fixed now, but two CI tests fail.

Shreeshrii:
Reported in https://github.com/tesseract-ocr/tesseract/issues/3226

amitdo:
LSTM does not support equation detection. only the legacy engine does.

amitdo:
I now see that you closed this issue because it duplicated #3226 which is still open.

amitdo:
>There was a build issue (missing include statements) which is fixed now

Can you give me a link to the reported build failure related to the missing include statement?

stweil:
https://github.com/stweil/tesseract/runs/3479277718?check_suite_focus=true

amitdo:
IMO, the right solution is to disable the delete in disable legacy mode (it will stay null all the time anyway in this mode). The header should not be included.

amitdo:
Or just ignore the warning because in this case it's harmless IMO.

stweil:
That code also triggers a clang crash. So ignoring of the warning is not a good solution.

Disabling the delete is fine if all other code with `equ_detect_` is disabled, too.

amitdo:
You should consider all files that is disabled in autotools, including headers as deleted files for disable legacy mode. If there is an issue the solution should not be re-include them. but to disable some more code.

A crash is bad. so it needs to be fixed. I wanted to avoid more ugly ifdefs, thinking this code is harmless, but it was not a good idea to avoid them.

amitdo:
The crash is a compiler time crash. A compiler should not crash due to errors in code it compiles. This crash is a clang bug. They telling you:

>PLEASE submit a bug report to https://bugs.llvm.org/

PR #3624, opened 2021-11-03
Improve the disable legacy build

amitdo:
Disable more code related to equation detection and osd.

### DIFF
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -49,7 +49,9 @@ class UNICHARSET;
 
 class Dawg;
 class Dict;
+#ifndef DISABLED_LEGACY_ENGINE
 class EquationDetect;
+#endif // ndef DISABLED_LEGACY_ENGINE
 class PageIterator;
 class ImageThresholder;
 class LTRResultIterator;
@@ -766,7 +768,9 @@ protected:
 protected:
   Tesseract *tesseract_;          ///< The underlying data object.
   Tesseract *osd_tesseract_;      ///< For orientation & script detection.
+#ifndef DISABLED_LEGACY_ENGINE
   EquationDetect *equ_detect_;    ///< The equation detector.
+#endif // ndef DISABLED_LEGACY_ENGINE
   FileReader reader_;             ///< Reads files from any filesystem.
   ImageThresholder *thresholder_; ///< Image thresholding module.
   std::vector<ParagraphModel *> *paragraph_models_;
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1897,15 +1897,13 @@ void TessBaseAPI::End() {
   if (osd_tesseract_ == tesseract_) {
     osd_tesseract_ = nullptr;
   }
-#endif // ndef DISABLED_LEGACY_ENGINE
-  delete tesseract_;
-  tesseract_ = nullptr;
-#ifndef DISABLED_LEGACY_ENGINE
   delete osd_tesseract_;
   osd_tesseract_ = nullptr;
   delete equ_detect_;
   equ_detect_ = nullptr;
 #endif // ndef DISABLED_LEGACY_ENGINE
+  delete tesseract_;
+  tesseract_ = nullptr;
   input_file_.clear();
   output_file_.clear();
   datapath_.clear();
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -29,7 +29,9 @@
 #include "dict.h"       // for Dict
 #include "elst.h"       // for ELIST_ITERATOR, ELISTIZE, ELISTIZEH
 #include "environ.h"    // for l_uint8
+#ifndef DISABLED_LEGACY_ENGINE
 #include "equationdetect.h" // for EquationDetect, destructor of equ_detect_
+#endif // ndef DISABLED_LEGACY_ENGINE
 #include "errcode.h" // for ASSERT_HOST
 #include "helpers.h" // for IntCastRounded, chomp_string
 #include "host.h"    // for MAX_PATH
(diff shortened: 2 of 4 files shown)


egorpugin:
Is it possible to drop legacy engine in some of future releases?

amitdo:
I think @stweil still wants to keep it.

Note that this build also drops osd because it uses the legacy engine.

src/api/baseapi.cpp
```diff
@@ -1889,15 +1893,19 @@ void TessBaseAPI::End() {
...
   }
+#endif // ndef DISABLED_LEGACY_ENGINE
   delete tesseract_;
   tesseract_ = nullptr;
+#ifndef DISABLED_LEGACY_ENGINE
   delete osd_tesseract_;
   osd_tesseract_ = nullptr;
   delete equ_detect_;
   equ_detect_ = nullptr;
+#endif // ndef DISABLED_LEGACY_ENGINE
```
stweil, line 1908: That lines can be moved up into the previous conditional block.

stweil (changes requested):

stweil:
> Note that this build also drops osd because it uses the legacy engine.

That part is still not complete and could disable more lines.

include/tesseract/baseapi.h
```diff
@@ -766,7 +768,9 @@ class TESS_API TessBaseAPI {
 protected:
   Tesseract *tesseract_;          ///< The underlying data object.
   Tesseract *osd_tesseract_;      ///< For orientation & script detection.
+#ifndef DISABLED_LEGACY_ENGINE
```
stweil, line 771: `osd_tesseract_` could be disabled, too.

amitdo: Well, you can't just disable it here, because it is needed in other places, so more ifdefs are needed.

I did disable **some** more osd parts because after I managed the `equ_detect_ ` delete issue, clang will probably have the same issue with osd. Note that with the disable legacy mode, `osd_tesseract_` is always `nullptr` so there is noting to delete.

amitdo: BTW, I tested this PR with clang++ on Ubuntu 20.04 with `--diable-legacy` and the compilation completed with no crash or errors.

src/api/baseapi.cpp
amitdo: Done.

stweil:
Thank you.

stweil:
While thinking more about these modifications, I now noticed that it might cause problems because of a conditional API change in class `TessBaseAPI`: now the size and memory layout of that class depends on `DISABLED_LEGACY_ENGINE`.

Users of the API must use the same setting for `DISABLED_LEGACY_ENGINE` which was used when building the Tesseract library. If there is a mismatch, this won't be detected by the compiler but result in runtime issues.

I wonder whether we should revert that part of the changes.

amitdo:
It's not meant to be compatible in this way. This occur in many other places with the disable legacy build.

stweil:
Before that commit it was possible to use the Tesseract API without legacy functions, no matter whether `DISABLED_LEGACY_ENGINE` was defined or not and with any kind of Tesseract library.

This is now no longer possible.

amitdo:
As I said, it's not the only place. So just reverting this PR will give you nothing. I don't know how much work is needed to make it compatible as you want to achieve.

amitdo:
If you decide to revert, cosider to do a partial revert. You can still keep the part that disable the delete of `equ_detect_` and the part that disable the header you re-enabled in a previous commit. This should still prevent clang to crash (I hope).

stweil:
> As I said, it's not the only place.

Could you give an example?

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/2a66694754e37537bacc7773623d1a2fa8463c61/include/tesseract/renderer.h#L295


https://github.com/tesseract-ocr/tesseract/blob/e9b8b840bf40d8809b409578bd16a53f0239f752/include/tesseract/baseapi.h#L159

https://github.com/tesseract-ocr/tesseract/blob/e9b8b840bf40d8809b409578bd16a53f0239f752/include/tesseract/baseapi.h#L645

https://github.com/tesseract-ocr/tesseract/blob/48c5d426cabadfc8e9f912db38896076fe347e8f/include/tesseract/capi.h#L336

https://github.com/tesseract-ocr/tesseract/blob/48c5d426cabadfc8e9f912db38896076fe347e8f/include/tesseract/capi.h#L356

So there are only a few places.

stweil:
Those conditionals are not problematic, because they only disable the declarations of some functions, and none of those functions is a virtual one. Compiling code with additional headers and additional function declarations does not change the compilation result (so those conditionals could also be removed without much effect). Removing a class variable like `equ_detect_` is different because it changes the compilation result: `TessBaseAPI` is now smaller, so any runtime object with that class no longer fits to other code which still has the larger `TessBaseAPI`.

It is not necessary to revert the whole commit. Making `equ_detect_` unconditional in the class definition and setting it to `nullptr` is sufficient to restore the compatibility.

amitdo:
> It is not necessary to revert the whole commit. Making `equ_detect_` unconditional in the class definition and setting it to `nullptr` is sufficient to restore the compatibility.

That was what I did at first, like what I suggested [here](https://github.com/tesseract-ocr/tesseract/issues/3620#issuecomment-956400243):

>IMO, the right solution is to disable the delete in disable legacy mode (it will stay null all the time anyway in this mode).

But then I change it because what you wrote [here](https://github.com/tesseract-ocr/tesseract/issues/3620#issuecomment-956426921):

(My emphasis) 
>Disabling the delete is fine if **all other code** with `equ_detect_` **is disabled, too**.

amitdo:
#3627

PR #3625, opened 2021-11-03
README: Remove the reference to version 3.05.02

amitdo:
Versions 4.1.1 and 5.0.0 still support the legacy engine with the same functionality as 3.05.02, so there is no reason to mention 3.05.02.


stweil (approved):
Thanks!

Issue #3560, opened 2021-09-14
output true CER for checkpoints (at least the final one)


bertsky:
AFAICS, `lstmtraining` produces two types of figures for measuring the error:
1. **bag-of-character training error** (on `list.train`): this is shown as
   - `char train=%.3f%%` every 100 iterations
   - `Finished! Error rate = %.2f` at the end
   - first figure in the file name of the checkpoint files
2. **bag-of-character testing error** (on `list.eval`): this is shown as
   - `Eval Char error rate=%f` (but also in percent) whenever (IIUC) training error reached a new minimum

<details><summary>Call-tree for 1:</summary>
<p>

- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtrainer.cpp#L945-L947
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtrainer.cpp#L1261-L1286
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtrainer.cpp#L1201-L1221
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtrainer.cpp#L1325-L1339
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtrainer.cpp#L369-L378

</p></details>
<details><summary>Call-tree for 2:</summary>
<p>

- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtester.cpp#L48-L134
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/unicharset/lstmtrainer.h#L151-L154
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/lstmtraining.cpp#L188-L197
- https://github.com/tesseract-ocr/tesseract/blob/7a308edcb1fc7455008b531bc2a49de583d7b171/src/training/lstmtraining.cpp#L217

</p></details>

[...]

PR #3644, opened 2021-11-15
Improve training messages (issue #3560)

stweil:
The old messages could wrongly be interpreted as CER / WER values,
but Tesseract training currently uses simple bag of characters /
bag of words error rates (see LSTMTrainer::ComputeCharError,
LSTMTrainer::ComputeWordError).


stweil:
@bertsky, I used the message texts which you suggested in issue #3560.

src/training/unicharset/lstmtester.cpp
```diff
@@ -118,8 +118,8 @@ std::string LSTMTester::RunEvalSync(int iteration, const double *training_errors
   std::string result;
```
bertsky, line 118: There's one more: one line 109 (`Line Char error rate` → `Line BCER`) – cannot use GH suggestions.

bertsky (approved):
Yes, that's the relevant places I know.

src/training/unicharset/lstmtester.cpp
stweil: That's fixed now, too.

PR #3650, opened 2021-11-19
Training: Fix compiler warnings

amitdo:
warning: format ‘%c’ expects argument of type ‘int’, but argument 2 has type ‘tesseract::Validator::CharClass’ [-Wformat=]

### DIFF
--- a/src/training/unicharset/validate_indic.cpp
+++ b/src/training/unicharset/validate_indic.cpp
@@ -35,7 +35,8 @@ bool ValidateIndic::ConsumeGraphemeIfValid() {
       return true;
     default:
       if (report_errors_) {
-        tprintf("Invalid start of grapheme sequence:%c=0x%x\n", codes_[codes_used_].first,
+        tprintf("Invalid start of grapheme sequence:%c=0x%x\n",
+                static_cast<int>(codes_[codes_used_].first),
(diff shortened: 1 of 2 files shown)


stweil:
Technically the current code works fine, because the enum `CharClass` is an `int`, so adding type casts only fixes the compiler warnings. Maybe it would also be better to use an `int` type cast instead of `char`, because that is expected by `%c`. The best solution is replacing `tprintf` by C++ code. There are also lots of similar warnings with `%p` where the compiler would like to get `void *`. I don't want to fix those.

amitdo:
There are two instances of this warning when doing `make training`.

These are the only warning I see during compiling (lib+training), so I want to eliminate them.

amitdo:
https://en.cppreference.com/w/c/io/fprintf

>Conversion
Specifier **c**:
The argument is first converted to unsigned char.

stweil:
Yes, internally it is handled as an `unsigned char`. But the same documentation also states that the expected argument is of type `int`.

amitdo:
I can change it to int. Do you want to eliminate the warning with a different code?

stweil:
A type cast to `int` is fine. Using C++ streams cannot be done for 5.0.0.

PR #3651, opened 2021-11-19
Add RowAttributes getter to PageIterator and format code

stweil:
Note that this is an API change for 5.0.0, but fixes compatibility with the 4.1 branch. See pull request #2971.


stweil:
@poke1024, does this change work for you?

PR #3657, opened 2021-11-25
Limit BCER to interval [0,1]

stweil:


stweil:
With the current code `lstmtraining` reports character error rates larger than 100% when a new training starts. This is fixed here.

Issue #3255, opened 2021-01-22
g++ 6.3 build failure


amitdo:
### Environment

* **Tesseract Version**: 5.0 alpha
* **Commit Number**: e3fd938bca0e
* **Platform**: Linux debian 4.9.0-14-amd64 #1 SMP Debian 4.9.246-2 (2020-12-17) x86_64 GNU/Linux

`gcc version 6.3.0 20170516 (Debian 6.3.0-18+deb9u1) `

g++ 6.3 fails to compile commit e3fd938bca0e.

[config.log](https://github.com/tesseract-ocr/tesseract/files/5858583/config.log)

[make.log](https://github.com/tesseract-ocr/tesseract/files/5858571/make.log)

stweil:
Building with g++-6 fails since commit 04d3cfcf2fb63baee8bf5a13c511c4c4b6791b71 according to my tests.

stweil:
The problems are indeed related to C++11. Those related to the `STRING` class can easily be fixed. As far as I see the only remaining code which needs a newer compiler is this:

    src/ccutil/unicharset.cpp: In member function ‘void tesseract::UNICHARSET::unichar_insert(const char*, tesseract::OldUncleanUnichars)’:
    ../../../src/ccutil/unicharset.cpp:618:37: error: forming reference to void
         auto &u = unichars.emplace_back();

In C++-11 `emplace_back` is a void function, so that needs different code if we want to support C++-11.

stweil:
The modifications which I mentioned above are not sufficient, because training has more dependencies on newer compilers. In addition, I get unresolved symbols when linking.

So I am afraid that supporting g++-6 would require more work. There are currently other, more important, problems.

stweil:
Builds of git master currently work with `-std=c++14`, but 3 code locations like the following one then show warnings because they use C++17 code:

    src/training/mftraining.cpp: In function ‘int main(int, char**)’:
    src/training/mftraining.cpp:207:8: warning: structured bindings only available with ‘-std=c++17’ or ‘-std=gnu++17’
      207 |   auto [trainer,shape_table] = tesseract::LoadTrainingData(argc, argv,
          |        ^

amitdo:
It's not important to me to have gcc-6 support, but I do want to know what are the minimum GCC, Clang and MSVC versions that can compile the current 5.0 code.

amitdo:
Travis CI seems to use gcc 6 and was able to build (with cmake) commit e3fd938.

[...]

PR #3663, opened 2021-11-28
Allow compilation with clang-7

stweil:

### DIFF
--- a/src/ccutil/helpers.h
+++ b/src/ccutil/helpers.h
@@ -252,7 +252,7 @@ bool Serialize(FILE *fp, const std::vector<T> &data) {
   uint32_t size = data.size();
   if (fwrite(&size, sizeof(size), 1, fp) != 1) {
     return false;
-  } else if constexpr (std::is_class_v<T>) {
+  } else if constexpr (std::is_class<T>::value) {
     // Serialize a tesseract class.
     for (auto &item : data) {
       if (!item.Serialize(fp)) {
--- a/src/ccutil/serialis.h
+++ b/src/ccutil/serialis.h
@@ -104,7 +104,7 @@ public:
     } else if (size > 50000000) {
       // Arbitrarily limit the number of elements to protect against bad data.
       return false;
-    } else if constexpr (std::is_same_v<T, std::string>) {
+    } else if constexpr (std::is_same<T, std::string>::value) {
       // Deserialize a string.
       // TODO: optimize.
       data.resize(size);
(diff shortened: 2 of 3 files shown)


stweil:
These small modifications allow building Tesseract 5 with clang-7 on Debian stretch. They are not sufficient for g++ 6.3 (see issue #3255).

egorpugin:
This happens when no C++17 available in stdlib.
Do we require C++17 support still?

(Compilers port some core language (not std) features into older standards. Like such `if constexpr` into c++14, c++11 and probably even c++98.)

stweil:
`configure` accepts g++ and clang++ on Debian stretch, so both compilers should formally support C++17. Nevertheless both fail to compile the current code because their C++17 support has deficits.

Therefore I think that it's good to check for C++17 to get at least a compiler which nearly supports it. If we would lower the requirements and only check C++14, we might risk getting more reports for old compilers which are nearly compatible with C++14.

egorpugin:
5.x.x releases won't get too much attention and code changes, so they could live with current "semi-C++17" standard.
---
As always I propose to move forward with the next 6,7... main branch (C++20/23).

egorpugin:
By the way, do we have v5 branch already? Such fixes could land exactly there.
But main will keep moving forward.

amitdo:
Before applying this patch, did you try to install libc++-7-dev, and tell clang to use it with `-stdlib=libc++7`(or `=libc++-7`, I don't remember...) with `-std=c++17` ?

amitdo:
Also note that Debiain Stretch is not maintained by the Debian project anymore.

https://wiki.debian.org/LTS

>Debian Long Term Support (LTS) is a project to extend the lifetime of all Debian stable releases to (at least) 5 years. Debian LTS is not handled by the Debian security team, but by a separate group of volunteers and companies interested in making it a success.
 
EOL of Debian Stretch LTS: June 30, 2022.

Personally, I don't think we should support it in 5.0.0.

amitdo:
>Compilers port some core language (not std) features into older standards. Like such `if constexpr` into c++14, c++11 and probably even c++98.

https://gcc.gnu.org/gcc-12/changes.html

>New warning options for C++ language mismatches: -Wc++11-extensions, -Wc++14-extensions, -Wc++17-extensions, -Wc++20-extensions, and -Wc++23-extensions. They are enabled by default and can be used to control existing pedwarns about occurences of new C++ constructs in code using an old C++ standard dialect.

amitdo:
These warnings are also available in Clang:

https://clang.llvm.org/docs/DiagnosticsReference.html#wc-17-extensions

PR #3669, opened 2021-12-02
Correctly detect amd64 and i386 on FreeBSD

pkubaj:
FreeBSD uses amd64 and i386 names.


amitdo:
Here is what is done in autoconf:

https://github.com/tesseract-ocr/tesseract/blob/924b746e6f8e02e634188de1f15c8a95ebe9680b/configure.ac#lL136-L144

PR #3671, opened 2021-12-02
Fix build on non-x86 non-ARM with clang

pkubaj:
Clang doesn't exit with error on powerpc64* after adding -mfpu=neon, but merely ignores the argument and prints a warning, which incorrectly makes it defined as ARM with NEON.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,7 +220,7 @@ set(HAVE_SSE4_1 FALSE)
 add_definitions("-DHAVE_NEON")
 set(HAVE_NEON TRUE)
 
-else()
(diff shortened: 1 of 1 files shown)


stweil:
@pkubaj, did you remove your tesseract repository? Meanwhile I could reproduce the issue and would like to merge your contribution.

stweil:
As soon as this issue is fixed, there remain more obstacles. See issue #3702 for more details.

pkubaj:
Right, I removed it by mistake. It's recreated now.

stweil:
Thank you. I could not merge it here but applied it in commit 790f3bd1041778dd8ba258f3a0320a59303aaff3.

PR #3681, opened 2021-12-16
Support up to 8 redirections when running OCR on a URL

stweil:


stweil:
See https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html for details.

I wonder whether this change should be handled as a bug fix (=> for release 5.0.1) or a new feature (=> for release 5.1.0). Redirects are needed very often because many existing image URLs still use HTTP, but the image server redirects to HTTPS.

Issue #3680, opened 2021-12-16
build for Android/i686,x86-64 failed due to 'asm/hwcap.h' file not found


leleliu008:
### Environment

* **Tesseract Version**:  https://github.com/tesseract-ocr/tesseract/archive/refs/tags/5.0.0.tar.gz
* **Commit Number**:   
* **Platform**: Android/i686

### Current Behavior:
```
/tmp/tmp.wrf96PRFHZ/src/arch/simddetect.cpp:60:14: fatal error: 'asm/hwcap.h' file not found
#    include <asm/hwcap.h>
             ^~~~~~~~~~~~~
1 error generated.
```
Android NDK not provide `asm/hwcap.h`  for i686,x86-64
### Expected Behavior:
successfully build for Android/i686

### Suggested Fix:
Android NDK not provide `asm/hwcap.h`  for i686,x86-64, but provide `asm/hwcap2.h`, I don't known if can replace `asm/hwcap.h` with `asm/hwcap2.h`

leleliu008:
```
CHECK_CXX_COMPILER_FLAG("-mfpu=neon" HAVE_NEON)
if(HAVE_NEON)
    set(NEON_COMPILE_FLAGS "-mfpu=neon")
    add_definitions("-DHAVE_NEON")
endif()
```

stweil:
That part should fail for Android/i686 and only be used for Android/arm*.

leleliu008:
no, `HAVE_NEON` always  `TRUE`, I'm sure.

leleliu008:
```
CHECK_CXX_COMPILER_FLAG("-mfpu=neon" HAVE_NEON)
```
`HAVE_NEON=FALSE` when CXX=gcc
`HAVE_NEON=TRUE`  when CXX=clang

Android NDK using clang

leleliu008:
CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.0.0)

project(test CXX)

include(CheckCXXCompilerFlag)
CHECK_CXX_COMPILER_FLAG("-mfpu=neon" HAVE_NEON)
if(HAVE_NEON)
    message("HAVE_NEON")
endif()
```
build with following command:
```
cmake -S . -B build.d -DCMAKE_CXX_COMPILER=gcc
cmake -S . -B build.d -DCMAKE_CXX_COMPILER=clang
```

`uname -a` :
```
[...]

PR #3682, opened 2021-12-18
fix for that motioned in https://github.com/tesseract-ocr/tesseract/issues/3680

leleliu008:
#3680 

Reference:
https://github.com/Kitware/CMake/blob/master/Modules/Platform/Android-Determine.cmake
https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake
https://developer.android.com/ndk/guides/cpu-features#the_google_cpu_features_library
https://github.com/google/cpu_features#android-ndks-drop-in-replacement


CMakeLists.txt
```diff
@@ -721,6 +721,22 @@ if (WIN32 AND CLANG AND OPENMP_BUILD)
...
 
+# fix for that motioned in https://github.com/tesseract-ocr/tesseract/issues/3680
+#
+# 2021-12-18 14:56:11 UTC  leleliu008@gmail.com
+#
+# Reference:
+# https://github.com/Kitware/CMake/blob/master/Modules/Platform/Android-Determine.cmake
+# https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake
+# https://developer.android.com/ndk/guides/cpu-features#the_google_cpu_features_library
+# https://github.com/google/cpu_features#android-ndks-drop-in-replacement
```
stweil, line 732:
```suggestion
```

stweil: Usually we don't add such details to the code, only short comments where needed. All that information is also in the commit message, so it does not get lost.

zdenop:
thanks

amitdo:
I don't think we should add `-DANDROID`.

https://stackoverflow.com/questions/15328751/android-macro-suddenly-not-defined

https://github.com/neoml-lib/neoml/issues/30

Issue #3683, opened 2021-12-20
Build with clang-cl on Windows fails 


zdenop:
Current clang build on windows fails with error:

```
167/179] Building CXX object CMakeFiles\libtesseract.dir\src\arch\simddetect.cpp.obj
FAILED: CMakeFiles/libtesseract.dir/src/arch/simddetect.cpp.obj
C:\PROGRA~1\LLVM\bin\clang-cl.exe  /nologo -TP -DCMAKE_BUILD -DHAVE_AVX -DHAVE_AVX2 -DHAVE_CONFIG_H -DHAVE_FMA -DHAVE_SSE4_1 -DTESS_EXPORTS -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_SECURE_NO_WARNINGS -D_SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS=1 -Dlibtesseract_EXPORTS -IF:\win64\include\leptonica -IF:\Project-Personal\tesseract\build.clang -IF:\Project-Personal\tesseract\build.clang\include -IF:\Project-Personal\tesseract\src -IF:\Project-Personal\tesseract\include -IF:\Project-Personal\tesseract\src\arch -IF:\Project-Personal\tesseract\src\ccmain -IF:\Project-Personal\tesseract\src\ccstruct -IF:\Project-Personal\tesseract\src\ccutil -IF:\Project-Personal\tesseract\src\classify -IF:\Project-Personal\tesseract\src\cutil -IF:\Project-Personal\tesseract\src\dict -IF:\Project-Personal\tesseract\src\lstm -IF:\Project-Personal\tesseract\src\opencl -IF:\Project-Personal\tesseract\src\textord -IF:\Project-Personal\tesseract\src\viewer -IF:\Project-Personal\tesseract\src\wordrec -IF:\Project-Personal\tesseract\src\training -imsvc F:\win64\include /DWIN32 /D_WINDOWS /W3 /GR /EHsc -Wno-microsoft-unqualified-friend /utf-8 /O2 /Ob2 /DNDEBUG /wd4244 /wd4305 /wd4267 /wd4251 /wd4275 -Wno-unused-command-line-argument -MD -flto=thin -std:c++17 /showIncludes /FoCMakeFiles\libtesseract.dir\src\arch\simddetect.cpp.obj /FdCMakeFiles\libtesseract.dir\ -c -- F:\Project-Personal\tesseract\src\arch\simddetect.cpp
[...]

PR #3684, opened 2021-12-22
Do not #define __attribute__(attr)

amitdo:
This should fix the broken clang-cl build.

### DIFF
--- a/src/ccutil/errcode.h
+++ b/src/ccutil/errcode.h
@@ -37,18 +37,17 @@ enum TessErrorLogCode {
 #define MEMORY_ABORT 2
 #define FILE_ABORT 3
 
-#if !defined(__GNUC__) && !defined(__attribute__)
-# define __attribute__(attr) // compiler without support for __attribute__
-#endif
-
 class TESS_API ERRCODE { // error handler class
   const char *message;   // error message
 public:
+#if defined(__GNUC__) && defined(__attribute__)
+  __attribute__((format(printf, 4, 5)))
+#endif
   void error(                  // error print function
       const char *caller,      // function location
       TessErrorLogCode action, // action to take
       const char *format, ...  // fprintf format
-  ) const __attribute__((format(printf, 4, 5)));
+  ) const;
   constexpr ERRCODE(const char *string) : message(string) {} // initialize with string
 };
 
--- a/src/ccutil/tprintf.h
+++ b/src/ccutil/tprintf.h
@@ -24,17 +24,15 @@
 
 namespace tesseract {
 
-#if !defined(__GNUC__) && !defined(__attribute__)
-# define __attribute__(attr) // compiler without support for __attribute__
-#endif
-
 // Disable some log messages by setting log_level > 0.
 extern TESS_API INT_VAR_H(log_level);
 
 // Main logging function.
+#if defined(__GNUC__) && defined(__attribute__)
+__attribute__((format(printf, 1, 2)))
+#endif
 extern TESS_API void tprintf( // Trace printf
-    const char *format, ...)  // Message
-    __attribute__((format(printf, 1, 2)));
+    const char *format, ...);  // Message
 
 } // namespace tesseract
 
--- a/src/viewer/scrollview.h
+++ b/src/viewer/scrollview.h
@@ -40,10 +40,6 @@
 
 namespace tesseract {
 
-#if !defined(__GNUC__) && !defined(__attribute__)
-# define __attribute__(attr) // compiler without support for __attribute__
-#endif
-
 class ScrollView;
 class SVNetwork;
 class SVSemaphore;
@@ -297,7 +293,10 @@ public:
   // ...which can be added by this command.
   // This is intended as an "debug" output window.
   void AddMessage(const char *message);
-  void AddMessageF(const char *format, ...) __attribute__((format(printf, 2, 3)));
+#if defined(__GNUC__) && defined(__attribute__)
+  __attribute__((format(printf, 2, 3)))
+#endif
+  void AddMessageF(const char *format, ...);
 
   // Zoom the window to the rectangle given upper left corner and
   // lower right corner.
@@ -310,7 +309,10 @@ public:
   // this just for fun will likely break your application!
   // It is public so you can actually take use of the LUA functionalities, but
   // be careful!
-  void SendMsg(const char* msg, ...) __attribute__((format(printf, 2, 3)));
+#if defined(__GNUC__) && defined(__attribute__)
+  __attribute__((format(printf, 2, 3)))
+#endif
+  void SendMsg(const char* msg, ...);
 
   // Custom messages (manipulating java code directly) can be send through this.
   // Send a message to the server without adding the


egorpugin:
What about templating our *printfs* and *formats* instead?

egorpugin:
It will remove any of these ifdefs, so profit is obvious.

src/ccutil/errcode.h
```diff
@@ -37,18 +37,17 @@ enum TessErrorLogCode {
...
 #define FILE_ABORT 3
 
-#if !defined(__GNUC__) && !defined(__attribute__)
-# define __attribute__(attr) // compiler without support for __attribute__
-#endif
-
 class TESS_API ERRCODE { // error handler class
   const char *message;   // error message
 public:
+#if defined(__GNUC__) && defined(__attribute__)
```
stweil, line 43: Is `__attribute__`defined anywhere? If not, then the new code simply disables all attributes.

stweil:
We could also use a new macro `GCC_FMT_ATTR` like it is done for [QEMU](https://gitlab.com/qemu-project/qemu/-/blob/master/include/qemu/compiler.h#L82). Then conditional code is only required in a single header file.

egorpugin:
> qemu

They are C. We can do better.

stweil:
Yes, the C++ solution is using streams.

stweil:
@amitdo, isn't it sufficient to add something like `/D__attribute__(n)` to fix a build with `clang_cl`?

src/ccutil/errcode.h
amitdo: For gcc and regular clang? Are you sure?

`__ attribute__() `can be uses before the function declaration instead of in the end.

stweil: Yes, but it is compiled conditionally, and the condition `&& defined(__attribute__)` is now always `false`.

amitdo:
I don't know.

Note that I didn't tested this PR with clang-cl, but I believe it will fix #3683.

If you have something better, you can close this PR.

amitdo:
>Yes, the C++ solution is using streams.

C++20 has std::format(), which is base on [fmtlib/fmt)](https://github.com/fmtlib/fmt). fmt also supports C++17.

src/ccutil/errcode.h
amitdo: I thought `_attribute_` is auto defined by the compiler. Any way, the second condition can be dropped.

amitdo: >then the new code simply disables all attributes.

I don't think you are right.

egorpugin:
> @amitdo, isn't it sufficient to add something like `/D__attribute__(n)` to fix a build with `clang_cl`?

No, build must has as less flags as possible.
It is better when code handles this.

amitdo:
If we want to use a logging library, there are at least two which use the fmt library: [spdlog](https://github.com/gabime/spdlog) and [fmtlog](https://github.com/MengRao/fmtlog).

stweil:
I have added a new branch [libfmt](https://github.com/tesseract-ocr/tesseract/tree/libfmt) which can be used to test that library and started by replacing `ScrollView::AddMsg` and `ScrollView::SendMsg`.

egorpugin:
Not sure whether it is needed.

We have several ways:
1. Just use variadic templates - the easiest solution and fast to implement.
2. Use fmt from C++20 - not ready yet, since libstdc++ does not have it. Probably only MS STL has fmt changes today.
3. Use fmt as dependency - additional dependency, bloating binaries

stweil:
I have added some more commits to the `libfmt` branch. It should now work with `clang_cl` and uses variadic templates for several functions with variable argument lists. As there is a huge number of `tprintf` calls, there still remain many of those calls which still need updates (replacing classical format strings by the new ones). Nevertheless those changes already show that the new template code requires more RAM (currently +89276 bytes in the Tesseract code and 96253 bytes for `libfmt`).

That means that variadic templates increase the initial memory requirements by about 6 %. Is that acceptable? Changing the code in 5.x is risky because of the large number of affected code lines which can only be tested partially.

egorpugin:
We need to compare this (libfmt) to raw variadics with C-printf style functions to check.
Comparison to existing code is not useful.

stweil:
> We need to compare this (libfmt) to raw variadics with C-printf style functions.

Can you provide such an implementation for `tprintf` as an example?

egorpugin:
Yes, I'll prepare an example.
By the way, list of warnings from main branch.
```
====================[ Build | tesseract | Debug ]===============================
/var/lib/snapd/snap/clion/177/bin/cmake/linux/bin/cmake --build /home/egor/dev/tesseract/cmake-build-debug --target tesseract
[33/223] Building CXX object CMakeFiles/libtesseract.dir/src/ccmain/resultiterator.cpp.o
/home/egor/dev/tesseract/src/ccmain/resultiterator.cpp: In member function ‘void tesseract::ResultIterator::IterateAndAppendUTF8TextlineText(std::string*)’:
/home/egor/dev/tesseract/src/ccmain/resultiterator.cpp:734:39: warning: format ‘%p’ expects argument of type ‘void*’, but argument 2 has type ‘tesseract::ROW_RES*’ [-Wformat=]
  734 |     tprintf("Strong Script dirs     [%p/P=%s]: ", it_->row(),
      |                                      ~^           ~~~~~~~~~~
      |                                       |                   |
      |                                       void*               tesseract::ROW_RES*
/home/egor/dev/tesseract/src/ccmain/resultiterator.cpp:737:39: warning: format ‘%p’ expects argument of type ‘void*’, but argument 2 has type ‘tesseract::ROW_RES*’ [-Wformat=]
  737 |     tprintf("Logical textline order [%p/P=%s]: ", it_->row(),
      |                                      ~^           ~~~~~~~~~~
      |                                       |                   |
      |                                       void*               tesseract::ROW_RES*
[50/223] Building CXX object CMakeFiles/libtesseract.dir/src/ccutil/elst.cpp.o
/home/egor/dev/tesseract/src/ccutil/elst.cpp: In member function ‘tesseract::ELIST_LINK* tesseract::ELIST_ITERATOR::forward()’:
/home/egor/dev/tesseract/src/ccutil/elst.cpp:198:66: warning: format ‘%p’ expects argument of type ‘void*’, but argument 5 has type ‘tesseract::ELIST_ITERATOR*’ [-Wformat=]
  198 |     NULL_NEXT.error("ELIST_ITERATOR::forward", ABORT, "This is: %p  Current is: %p", this, current);
      |                                                                 ~^                   ~~~~
      |                                                                  |                   |
      |                                                                  void*               tesseract::ELIST_ITERATOR*
/home/egor/dev/tesseract/src/ccutil/elst.cpp:198:82: warning: format ‘%p’ expects argument of type ‘void*’, but argument 6 has type ‘tesseract::ELIST_LINK*’ [-Wformat=]
  198 |     NULL_NEXT.error("ELIST_ITERATOR::forward", ABORT, "This is: %p  Current is: %p", this, current);
      |                                                                                 ~^         ~~~~~~~
      |                                                                                  |         |
      |                                                                                  void*     tesseract::ELIST_LINK*
[54/223] Building CXX object CMakeFiles/libtesseract.dir/src/ccutil/elst2.cpp.o
/home/egor/dev/tesseract/src/ccutil/elst2.cpp: In member function ‘tesseract::ELIST2_LINK* tesseract::ELIST2_ITERATOR::forward()’:
/home/egor/dev/tesseract/src/ccutil/elst2.cpp:193:67: warning: format ‘%p’ expects argument of type ‘void*’, but argument 5 has type ‘tesseract::ELIST2_ITERATOR*’ [-Wformat=]
  193 |     NULL_NEXT.error("ELIST2_ITERATOR::forward", ABORT, "This is: %p  Current is: %p", this,
      |                                                                  ~^                   ~~~~
      |                                                                   |                   |
      |                                                                   void*               tesseract::ELIST2_ITERATOR*
/home/egor/dev/tesseract/src/ccutil/elst2.cpp:193:83: warning: format ‘%p’ expects argument of type ‘void*’, but argument 6 has type ‘tesseract::ELIST2_LINK*’ [-Wformat=]
  193 |     NULL_NEXT.error("ELIST2_ITERATOR::forward", ABORT, "This is: %p  Current is: %p", this,
      |                                                                                  ~^
      |                                                                                   |
      |                                                                                   void*
  194 |                     current);
      |                     ~~~~~~~                                                        
      |                     |
      |                     tesseract::ELIST2_LINK*
/home/egor/dev/tesseract/src/ccutil/elst2.cpp: In member function ‘tesseract::ELIST2_LINK* tesseract::ELIST2_ITERATOR::backward()’:
/home/egor/dev/tesseract/src/ccutil/elst2.cpp:233:68: warning: format ‘%p’ expects argument of type ‘void*’, but argument 5 has type ‘tesseract::ELIST2_ITERATOR*’ [-Wformat=]
  233 |     NULL_PREV.error("ELIST2_ITERATOR::backward", ABORT, "This is: %p  Current is: %p", this,
      |                                                                   ~^                   ~~~~
      |                                                                    |                   |
      |                                                                    void*               tesseract::ELIST2_ITERATOR*
/home/egor/dev/tesseract/src/ccutil/elst2.cpp:233:84: warning: format ‘%p’ expects argument of type ‘void*’, but argument 6 has type ‘tesseract::ELIST2_LINK*’ [-Wformat=]
  233 |     NULL_PREV.error("ELIST2_ITERATOR::backward", ABORT, "This is: %p  Current is: %p", this,
      |                                                                                   ~^
      |                                                                                    |
      |                                                                                    void*
  234 |                     current);
      |                     ~~~~~~~                                                         
      |                     |
      |                     tesseract::ELIST2_LINK*
[142/223] Building CXX object CMakeFiles/libtesseract.dir/src/dict/dict.cpp.o
In file included from /home/egor/dev/tesseract/src/dict/dawg_cache.h:24,
                 from /home/egor/dev/tesseract/src/dict/dict.h:30,
                 from /home/egor/dev/tesseract/src/dict/dict.cpp:19:
/home/egor/dev/tesseract/src/ccutil/object_cache.h: In instantiation of ‘tesseract::ObjectCache<T>::~ObjectCache() [with T = tesseract::Dawg]’:
/home/egor/dev/tesseract/src/dict/dawg_cache.h:29:7:   required from here
/home/egor/dev/tesseract/src/ccutil/object_cache.h:44:27: warning: format ‘%p’ expects argument of type ‘void*’, but argument 2 has type ‘tesseract::ObjectCache<tesseract::Dawg>*’ [-Wformat=]
   44 |             "ObjectCache(%p)::~ObjectCache(): WARNING! LEAK! object %p "
      |                          ~^
      |                           |
      |                           void*
/home/egor/dev/tesseract/src/ccutil/object_cache.h:44:70: warning: format ‘%p’ expects argument of type ‘void*’, but argument 3 has type ‘tesseract::Dawg*’ [-Wformat=]
   44 |             "ObjectCache(%p)::~ObjectCache(): WARNING! LEAK! object %p "
      |                                                                     ~^
      |                                                                      |
      |                                                                      void*
   45 |             "still has count %d (id %s)\n",
   46 |             this, it.object, it.count, it.id.c_str());
      |                   ~~~~~~~~~                                           
      |                      |
      |                      tesseract::Dawg*
[167/223] Building CXX object CMakeFiles/libtesseract.dir/src/textord/colfind.cpp.o
/home/egor/dev/tesseract/src/textord/colfind.cpp: In member function ‘bool tesseract::ColumnFinder::AssignColumns(const PartSetVector&)’:
/home/egor/dev/tesseract/src/textord/colfind.cpp:711:63: warning: format ‘%p’ expects argument of type ‘void*’, but argument 4 has type ‘tesseract::ColPartitionSet*’ [-Wformat=]
  711 |           tprintf("Set id %d did not match at y=%d, lineset =%p\n", col_i, part_i, line_set);
      |                                                              ~^                    ~~~~~~~~
      |                                                               |                    |
      |                                                               void*                tesseract::ColPartitionSet*
[171/223] Building CXX object CMakeFiles/libtesseract.dir/src/textord/colpartition.cpp.o
/home/egor/dev/tesseract/src/textord/colpartition.cpp: In member function ‘void tesseract::ColPartition::ComputeLimits()’:
/home/egor/dev/tesseract/src/textord/colpartition.cpp:1004:44: warning: format ‘%p’ expects argument of type ‘void*’, but argument 2 has type ‘tesseract::ColPartition*’ [-Wformat=]
 1004 |     tprintf("Recomputed box for partition %p\n", this);
      |                                           ~^     ~~~~
      |                                            |     |
      |                                            void* tesseract::ColPartition*
[216/223] Building CXX object CMakeFiles/libtesseract.dir/src/wordrec/language_model.cpp.o
/home/egor/dev/tesseract/src/wordrec/language_model.cpp: In member function ‘bool tesseract::LanguageModel::UpdateState(bool, int, int, tesseract::BLOB_CHOICE_LIST*, tesseract::LanguageModelState*, tesseract::LMPainPoints*, tesseract::WERD_RES*, tesseract::BestChoiceBundle*, tesseract::BlamerBundle*)’:
/home/egor/dev/tesseract/src/wordrec/language_model.cpp:257:25: warning: format ‘%p’ expects argument of type ‘void*’, but argument 2 has type ‘tesseract::LanguageModelState*’ [-Wformat=]
  257 |       tprintf("(parent=%p)\n", parent_node);
      |                        ~^      ~~~~~~~~~~~
      |                         |      |
      |                         void*  tesseract::LanguageModelState*
/home/egor/dev/tesseract/src/wordrec/language_model.cpp: In member function ‘bool tesseract::LanguageModel::AddViterbiStateEntry(tesseract::LanguageModelFlagsType, float, bool, int, int, tesseract::BLOB_CHOICE*, tesseract::LanguageModelState*, tesseract::ViterbiStateEntry*, tesseract::LMPainPoints*, tesseract::WERD_RES*, tesseract::BestChoiceBundle*, tesseract::BlamerBundle*)’:
/home/egor/dev/tesseract/src/wordrec/language_model.cpp:591:29: warning: format ‘%p’ expects argument of type ‘void*’, but argument 2 has type ‘tesseract::ViterbiStateEntry*’ [-Wformat=]
  591 |       tprintf(" parent_vse=%p\n", parent_vse);
      |                            ~^     ~~~~~~~~~~
      |                             |     |
      |                             void* tesseract::ViterbiStateEntry*
[223/223] Linking CXX executable bin/tesseract

Build finished
```

egorpugin:
See https://pastebin.com/7k1W2AzG
This patch fixes all warnings above.

stweil:
> This patch fixes all warnings above.

Does the new code still check whether the format string and the arguments match? It should report an error if someone tries code like `tprintf("num=%d");`.

egorpugin:
As I said in some other ticket, all `__attribute__` checks are now responsibility of libc code. They are delegated to `fprintf` automatically.

stweil:
I tried your code. It adds +113369 bytes to the `tesseract` executable.

egorpugin:
> currently +89276 bytes in the Tesseract code and 96253 bytes for libfmt
and
> +113369 

So, is it 90kB+96kB vs. 113kB or 90 vs 113?

egorpugin:
Ok.

At the end I do not recommend to evaluate binary sizes a lot.
Disk and memory space are cheap enough. And tesseract should not target embedded usages as primary direction.
Embedded should use statically linked binaries with very agressive optimizations, LTO/WHOPR with ipa passes etc.

stweil:
The new `tprintf` code with variadic template does not detect wrong format strings. Similar errors in direct calls of `fprintf` are reported. I tested that with `clang` on Linux.

Issue #3674, opened 2021-12-09
FreeBSD tesseract installation - missing configs dir


StefCGN:
Hi,
when installing tesseract on FreeBSD, the configs dir, which should be in "/usr/local/share/tessdata/" is not created.

It was necessary for me, to download the complete package from here, extract the configs folder and place it on my server.

After that, tesseract was working fine.

Please fix that!

stweil:
Which version of Tesseract did you try to install?
How did you build / install Tesseract?

Additional information is needed to understand, reproduce and finally fix the issue.

StefCGN:
HI,
i installed the folllowing:
`pkg install tesseract-5.0.0 tesseract-data-4.1.0`

After that, i ran the following command:
`ocrmypdf -v --redo-ocr 2020_11_30_Kellerdecke_neu.pdf Keller_neu.pdf`

This told me, that tesseract was not able to open "params_file" txt and pdf. So, ocrmypdf didnt create such a file.

I downloaded the package from here, extracted the configs dir and placed it into the tessdata folder.

After that, it was working.

stweil:
Did you report the problem to the maintainer of the FreeBSD package?

I am afraid here is the wrong place for your report.

stweil:
Cc'ing @pkubaj.

StefCGN:
Where is that package located?

stweil:
It looks like FreeBSD uses the cmake build. Debian bases distributions use the autotools build (which I normally use, too).

stweil:
> Where is that package located?

https://www.freebsd.org/cgi/ports.cgi?query=tesseract-5.0.0
https://cgit.freebsd.org/ports/tree/graphics/tesseract

StefCGN:
Ok.
And who is maintaining this package? Because i installed it via pkg.

stweil:
@egorpugin, I just tried `make install` with a cmake build. It did not install `/usr/local/share/tessdata`. Is that a regression compared to 4.1.3?

egorpugin:
Does it work on 4.1.3?

amitdo:
https://github.com/marketplace/actions/freebsd-vm

Maybe someone here wants to use this and write an action to test Tesseract on FreeBSD.

egorpugin:
I did not implement our cmake install.
Probably it was contributed via PRs, so not every expected functionality might present.

stweil:
[...]

Issue #3673, opened 2021-12-05
Plans for tesseract 5.x.y


amitdo:
I suggest to focus on 5.x for 2022 at least.

That means we should not break the API (and ABI?). Use C++17, not C++20/C++23.

stweil:
What about releasing a 5.0.1 after Christmas at the end of December? I think there are several fixes since 5.0.0 which would be good for a new release.

amitdo:
Mind reader :-)
I was about to suggest to release 5.0.1 before year end. But it would be nice if we can fix #3683 before releasing 5.0.1.

amitdo:
Right before tagging 5.0.1. you can update this sentence from the README:

>The latest stable version is 5.0.0, released on November 30, 2021.

egorpugin:
What should be added into v5?
5.x changes could be merged into branch and cherry picked into v6 main.

stweil:
We already have a wish list for improved training, a lot of issues with layout detection, want improved logging, and much more. Maintaining two branches did not work good with 4.x, and I am afraid it would not work better with 5.x.

egorpugin:
Maybe keep 5.0 as is? It is a good release with a number of changes.
Everything else will go straight into 6?

amitdo:
https://github.com/tesseract-ocr/tessdoc/blob/main/ReleaseNotes.md#changes-made-since-last-release

amitdo:
> What about releasing a 5.0.1 after Christmas at the end of December? I think there are several fixes since 5.0.0 which would be good for a new release.

Do you plan to release 5.0.1 next week?

stweil:
Yes, unless we discover that something very important is still missing.

stweil:
> It would be nice if we can fix #3683 before releasing 5.0.1.

There is still no fix, and I still have no `clang-cl`, so I cannot look for a fix myself. Should we release 5.0.1 without a fix? Are other things missing for 5.0.1 (besides updating of the documentation)?

egorpugin:
clang-cl is not worth it currently.

amitdo:
You can release 5.0.1 without the clang-cl fix.

stweil:
Release 5.0.1 is now online.

stweil:
[...]

PR #3685, opened 2021-12-22
Cmake install

zdenop:
This is aim to sync with autotools make install to satisfied requests like #3674
Question is if to install also [pdf.tff](https://github.com/tesseract-ocr/tesseract/blob/main/tessdata/pdf.ttf) - autotools install it, but it is not needed as there is [pdf_ttf.h](https://github.com/tesseract-ocr/tesseract/blob/main/src/api/pdf_ttf.h).

Relevant code is here: 
https://github.com/tesseract-ocr/tesseract/blob/60fd2b4abaa9c5c5c42d32db57576bc95d28a78a/src/api/pdfrenderer.cpp#L619-L633


egorpugin:
There is some `x` file in PR.

amitdo:
>Question is if to install also pdf.tff - autotools install it, but it is not needed as there is pdf_ttf.h.

@stweil, should we do the same for autotools, not install pdf.tff?

stweil:
That might be an incompatible change and require a new major version, even if the file was unused.

egorpugin:
@stweil 
Please consider making 5.0 branch, so we can push modern c++ or big (and possibly breaking) changes into main.

amitdo:
Egor, your request conflicts with mine from #3673 - to focus on 5.x in 2022.

zdenop:
> That might be an incompatible change and require a new major version, even if the file was unused.

Any guidelines for this? I had a look at [https://semver.org/](https://semver.org/) and IMO such change has no effect on API and the new version would be backward compatible.

stweil:
I think that the C / C++ API is only one aspect of compatibility.

Removing a file would break distributions which still have that file in their list of installed files. Users might have used that file for special applications.

Note that it is not necessary to add `pdf.tff` to the cmake install rules. They never installed it, so nothing breaks if they now install additional files, but not that one.

Issue #3688, opened 2021-12-23
RFC: Use SPDX license identifier instead of long license plate.


stweil:
// SPDX-License-Identifier: Apache-2.0

https://spdx.org/licenses/
Linux uses them already.
https://github.com/torvalds/linux/blob/master/arch/x86/kernel/cpuid.c#L1

stweil:
I created this issue from @egorpugin's  project list item to allow broader discussion.

Using SPDX license identifiers would reduce the source code lines by removing redundant information from the comments in the file headers (1 line instead of 10 lines).

We could start with the public header files in `include/tesseract`, use it for new files and then extend the changes to the remaining files as well.

egorpugin:
I think it can be automated in case if we remove other info in headers (date, author etc.)

stweil:
Note that the planning ["Tesseract next"](https://github.com/tesseract-ocr/tesseract/projects/1) now includes additional columns for bug fixes and new features for release 5. If needed we could also create a separate project planning "Tesseract future" for release 6.

stweil:
> I think it can be automated in case if we remove other info in headers

Should such additional information be removed? Do we have to preserve it? I don't know the answer and would therefore start with removing the license text only.

egorpugin:
Author probably should present in git history but it can't be true in all cases.
Copyright is used usually.
For example, I have it https://github.com/SoftwareNetwork/sw/blob/b0.4.3/src/sw/manager/database.cpp#L1

amitdo:
>Author: ???

Didn't you remove such lines in the past?

IMO, Author and Copyright should be kept for legal reason.

stweil:
Yes, I remove `History` and `Created` in files where I make other changes.

amitdo:
My point was that you removed developers names AFAIR. I hope all of them are at least mentioned in AUTHORS.

I think most if not all the removed named were HP employees. Maybe also some Googlers.

stweil:
You are right, I did not remember that I removed authors in a few cases. I now found the 3 commits b5498c70fa0b171cb952e04c5d9176a09c70963b (new file content), d960a50c12c4b991d0a86ff4c1fb9b05fd580aae and 6d170a15ec7ca0950fc69734ed586ffe6465f9ca (uncertain author). So all of them where very special cases.

stweil:
So do we agree to

- replace the license text by SPDX license identifier
- remove `File:`, `History:` and `Created:`
[...]

PR #3689, opened 2021-12-25
Add SPDX-License-Identifier to public include files

stweil:


stweil:
See issue #3688. Adding SPDX license identifiers to the public API allows automated parsing of license information.

egorpugin (approved):

egorpugin:
I don't like existing ///////////////////////////// first line.
Can we remove it later too?

stweil:
> Can we remove it later too?

Yes, I can add a commit which removes it. But maybe the `////` line at the end of the header comments should then be removed, too.

stweil:
Is PR #3690 fine?

Issue #3694, opened 2021-12-29
potential FPE 


junsik-kim0:
Describe the bug
A bug was found within the tesseract. Though it might not be an intended use of the relevant API, the bug can still produce critical issues within a program using tesseract. It would be best if the affected logic is checked beforehand.
The bug was found with a fuzzer based on the function "STATSTest.BasicStats"
This may cause problems in the use of libraries
If you give the Range value as the same value in the following code
Problems occur in Pile_Count Function
Array Index becomes -1

  stats_.set_range(150995072, 150995072);

  int32_t pile_count(int32_t value) const {
    if (value <= rangemin_) {
      return buckets_[0];
    }
    if (value >= rangemax_ - 1) {
      return buckets_[rangemax_ - rangemin_ - 1]; //Array Index becomes -1
    }
    return buckets_[value - rangemin_];
  }

How To Reproduce
git clone --recursive https://github.com/tesseract-ocr/tesseract.git
git clone https://github.com/tesseract-ocr/langdata_lstm.git
git clone https://github.com/tesseract-ocr/tessdata.git 
Download the attached file
cp stats_test.cc tesseract/unittest/.  (attached file)
cd tesseract 
./autogen.sh
./configure --prefix=/root/fuzz-test-generation/tmp/tesseract/tesseract/local --enable-shared=no CC="clang" CXX="clang++" CFLAGS="-g" CXXFLAGS="-g" LDFLAGS="-g"
make -j2 clean
make -j2 training
make -j2 check
./stats_test --gtest_filter="STATSTest.BasicStats"

Expected behavior
It should be terminated normally without crash.

Platform (please complete the following information):
OS: Ubuntu 18.04
[tesseract_bug1.tar.gz](https://github.com/tesseract-ocr/tesseract/files/7787272/tesseract_bug1.tar.gz)

stweil:
Could you please check whether pull request #3695 fixes the issue?

stweil:
The crash is not caused by a negative index, but because `buckets_ == nullptr`.

stweil:
This issue is now fixed in Git main.

PR #3695, opened 2021-12-29
Catch nullptr in STATS::pile_count (fix isse #3694)

stweil:


stweil:
This change is incomplete (it does not fix the issue #3694). There is also a related TODO from Ray to fix the definition of `rangemax_`.

stweil:
Issue #3694 is fixed now by the updated commit.

Issue #3702, opened 2022-01-01
cmake build fails on Ubuntu bionic because of old cmake version


stweil:
cmake version 3.10.2 aborts the build:

```
CMake Error at src/training/CMakeLists.txt:4 (cmake_policy):
  Policy "CMP0074" is not known to this version of CMake.
```

Can this policy be set conditionally only for newer versions of cmake?

After removing the policy line, the build sets wrong linker options for Curl (caused by commit c4ad9b7bbf1a390872402cf52cc52e9e9f4901e2):

```
CMake Error at CMakeLists.txt:812 (add_executable):
  Target "tesseract" links to target "CURL::libcurl" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
[...]
```

After fixing that locally, building `tesseract` works, but the training tools give linker errors:

```
[ 90%] Linking CXX executable ../../bin/text2image
cd /home/ubuntu/src/github/tesseract-ocr/tesseract/build/src/training && /usr/bin/cmake -E cmake_link_script CMakeFiles/text2image.dir/link.txt --verbose=1
/usr/bin/clang++-10  -O3 -DNDEBUG -Wno-unused-command-line-argument   CMakeFiles/text2image.dir/text2image.cpp.o CMakeFiles/text2image.dir/degradeimage.cpp.o  -o ../../bin/text2image libpango_training.a libunicharset_training.a libcommon_training.a ../../libtesseract.a -lpthread /usr/lib/powerpc64le-linux-gnu/libarchive.so /usr/lib/powerpc64le-linux-gnu/libcurl.so -llept 
/usr/bin/ld: libpango_training.a(ligature_table.cpp.o): undefined reference to symbol '_ZN6icu_608ByteSink5FlushEv'
//usr/lib/powerpc64le-linux-gnu/libicuuc.so.60: error adding symbols: DSO missing from command line
clang: error: linker command failed with exit code 1 (use -v to see invocation)
src/training/CMakeFiles/text2image.dir/build.make:126: recipe for target 'bin/text2image' failed
make[2]: *** [bin/text2image] Error 1
make[2]: Leaving directory '/home/ubuntu/src/github/tesseract-ocr/tesseract/build'
CMakeFiles/Makefile2:228: recipe for target 'src/training/CMakeFiles/text2image.dir/all' failed
make[1]: *** [src/training/CMakeFiles/text2image.dir/all] Error 2
[...]

PR #3704, opened 2022-01-02
Two fixes for cmake builds

stweil:


stweil:
Related issue: #3702.

PR #3705, opened 2022-01-03
Fix build with cmake 3.10.2 (Ubuntu bionic)

stweil:
This fixes issue #3702.


stweil:
Issue #3671 is still open as the pull request here only fixes the cmake related problems. Therefore builds on Ubuntu bionic ppc64le still fail.

PR #3708, opened 2022-01-03
(Draft) Add support for Unicode paths on Windows

danpla:
Tesseract doesn't work on Windows if it resides in a path that contains characters outside the system ANSI code page: it cannot list or load languages in this case. This is a proof-of-concept pull request (not intended to be merged) with minimal set of changes that fix TessBaseAPI::GetAvailableLanguagesAsVector() and TessBaseAPI::Init().

The final PR will probably need:
* Replacing the remaining fopen() calls with the UTF-8 version
* archive_read_open_filename_w()


stweil:
Does it help if `setlocale(LC_ALL, ".UTF8")` is executed right at the start of the program (for example in function `main`)? According to the documentation this should accept UTF-8 strings for all C runtime functions.

And what about C++ file functions? Do they accept UTF-8 on Windows?

danpla:
I can't test `setlocale(LC_ALL, ".UTF8")` right now since I'm using Windows 7, while the documentation says it was added in Windows 10. Anyway, docs say it only affects functions from C runtime, so I suppose ANSI variants of WinAPI routines (`FindFirstFile`, `GetModuleFileName`, etc.) will work as usual.

Wide-char routines seem to be more appropriate as they don't interfere with the global locale, which may be important when libtesseract is used in a third-part app. They are also available in Windows versions older than 10, if that matters.

> And what about C++ file functions? Do they accept UTF-8 on Windows?

If you mean `std::basic_*fstream`, then MSVC, GCC, and clang provide `wchar_t` overloads for the constructor and `open()`:
* https://docs.microsoft.com/en-us/cpp/standard-library/basic-fstream-class?view=msvc-170&redirectedfrom=MSDN#basic_fstream
* https://github.com/gcc-mirror/gcc/blob/16e2427f50c208dfe07d07f18009969502c25dc8/libstdc%2B%2B-v3/include/std/fstream#L1067
* https://github.com/llvm/llvm-project/blob/eb8650a75793b2bd079d0c8901ff066f129061da/libcxx/include/fstream#L1545

src/api/baseapi.cpp
```diff
@@ -156,26 +157,29 @@ static void addAvailableLanguages(const std::string &datadir, const std::string
   }
   const size_t extlen = sizeof(kTrainedDataSuffix);
 #ifdef _WIN32
-  WIN32_FIND_DATA data;
-  HANDLE handle = FindFirstFile((datadir + base2 + "*").c_str(), &data);
+  const auto kTrainedDataSuffixUtf16 = winutils::Utf8ToUtf16(kTrainedDataSuffix);
+
+  WIN32_FIND_DATAW data;
+  HANDLE handle = FindFirstFileW(
```
egorpugin, line 163: C++17 filesystem iterators should be used instead.

egorpugin (commented):
`fopenUtf8` with `_wfopen` is fine.
But improvement can be done using C++ std::*fstream + fs::path in binary mode. Here you don't need any utf16<->utf8 conversions at all.

But FindFirstFileW should be replaced with fs iterators.

In general I'd stuck to c++-style approach, not C.

If you say *fstreams are slow, it is mostly for binary mode.
In case you need maximum speed, you should use OS specific mmap family.

danpla:
> But improvement can be done using C++ std::*fstream + fs::path in binary mode. Here you don't need any utf16<->utf8 conversions at all.

It seems like constructing fs::path from UTF-8 still requires using either `fs::u8path()` or (since C++20) `char8_t` constructor. On Windows, passing UTF-8 `const char*` directly to fs::path literally transforms the data to `wchar_t` without UTF-8 -> UTF-16 conversion.

egorpugin:
There is https://en.cppreference.com/w/cpp/filesystem/path/u8path

danpla:
There's a small catch with `fs::path::u8string()`: it returns `std::u8string` instead of `std::string` since C++20. As I understand, making the code compile both in 17 and 20 requires either constructing strings like `std::string(u8string.begin(), u8string.end())` or writing helper functions like http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r2.html#conversion_fns

egorpugin:
Yes, we'll jump on C++20.
Don't bother with C++17 in 5.0 branch.

stweil:
> I can't test `setlocale(LC_ALL, ".UTF8")` right now since I'm using Windows 7, while the documentation says it was added in Windows 10. Anyway, docs say it only affects functions from C runtime, so I suppose ANSI variants of WinAPI routines (`FindFirstFile`, `GetModuleFileName`, etc.) will work as usual.

Recent versions of Windows 10 and Windows 11 support applications with UTF-8 codepage. That should support paths with any characters and is my preferred solution. It would not fix the issue for older versions of Window, but I think that is acceptable as those versions are unsupported by Microsoft and because it's easy to avoid the issue by renaming critical paths.

danpla:
Ok, I'll close the PR then.

I tested [the app manifest approach](https://docs.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page) on Windows 10, and it seems to work fine: Tesseract is able to both list (`addAvailableLanguages`) and load (`LoadDataFromFile`) language files.

PR #3712, opened 2022-01-06
cmake: TIFF is only used for WIN32

stweil:


stweil:
This avoids another cmake build issue on MacOS with Homebrew.

PR #3716, opened 2022-01-11
Check if platform supports feenableexcept

ffontaine:
feenableexcept is not supported by uclibc

Signed-off-by: Gilles Talis <gilles.talis@gmail.com>
[Retrieved (and updated to add cmake support and simplify configure.ac)
from
https://git.buildroot.net/buildroot/tree/package/tesseract-ocr/0001-Check-if-platform-supports-feenableexcept.patch]

### DIFF
--- a/cmake/Configure.cmake
+++ b/cmake/Configure.cmake
@@ -104,6 +104,13 @@ set(types_list
 )
 check_types(types_list)
 
+list(APPEND CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
+list(APPEND CMAKE_REQUIRED_LIBRARIES -lm)
+set(functions_list
+    feenableexcept
+)
+check_functions(functions_list)
+
 file(APPEND ${AUTOCONFIG_SRC} "
 /* Version number */
 #cmakedefine PACKAGE_VERSION \"${PACKAGE_VERSION}\"
--- a/configure.ac
+++ b/configure.ac
@@ -188,6 +188,9 @@ case "${host_cpu}" in
 
 esac
 
+# check whether feenableexcept is supported. some C libraries (e.g. uclibc) don't.
+AC_CHECK_FUNCS([feenableexcept])
+
 AX_CHECK_COMPILE_FLAG([-fopenmp-simd], [openmp_simd=true], [openmp_simd=false], [$WERROR])
 AM_CONDITIONAL([OPENMP_SIMD], $openmp_simd)
 
--- a/src/tesseract.cpp
+++ b/src/tesseract.cpp
@@ -629,7 +629,7 @@ static void PreloadRenderers(tesseract::TessBaseAPI &api,
  **********************************************************************/
 
 int main(int argc, char **argv) {
-#if defined(__USE_GNU)
+#if defined(__USE_GNU) && defined(HAVE_FEENABLEEXCEPT)
   // Raise SIGFPE.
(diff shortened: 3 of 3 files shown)


stweil:
These changes should not be needed. The GNU C library has `feenableexcept`, and other C libraries should not define `__USE_GNU`. Where does it come from when uClibc is used?

ffontaine:
uclibc-ng defines `__USE_GNU` (cf. https://github.com/wbx-github/uclibc-ng/blob/98680eec548b921abecc641b70246587ffbbdc0a/include/features.h#L350) as well as `__UCLIBC__`. However, I would be more than happy to just disable tesseract with uclibc-ng. This PR (as well as #3717) is needed because tesseract is raising build failures on buildroot autobuilders since our bump from version 4.1.1 to 5.0.1.

stweil:
I see two acceptable solutions:

1. If nobody uses Tesseract with uClibc I'd disable it. That does not require changes in the Tesseract code and reduces electrical power consumption.
2. If it is used, we could add conditional compilation using `__UCLIBC__`. Normally I prefer the way which you have implemented in your two commits, but this looks like a very special case needed because of an incomplete C library, so using that macro would be acceptable for me.

I always hesitate to add new checks to `configure` because each check increases the execution time and the log file a little bit more.

ffontaine:
After further investigation, [musl](https://musl.libc.org) doesn't support `feenableexcept` either: https://github.com/ifduyue/musl/search?q=feenableexcept. As we can't really say that musl is _incomplete_ (e.g. it is used by [Alpine Linux](https://www.alpinelinux.org)), I would say that `feenableexcept` is a glibc specific extension. 

Guarding the call to `feenableexcept` by `defined(__GLIBC__) && !defined(__UCLIBC__)` should work but this is really ugly compared to having the proper function checks in configure ...

stweil:
You are right, `feenableexcept` is a glibc extension. From the documentation: "Since version 2.2, glibc supports the functions feenableexcept() and fedisableexcept() to set individual floating-point traps". So `#if defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2,2)` should be the right test.

uClibc claims to be glibc 2.2 by setting the relevant macros, but is obviously incompatible to glibc 2.2. I'd call that a bug.

We can use a modified test `#if defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2,3)` which should work for uClibc and musl (which correctly does not claim to be glibc).

ffontaine:
Unfortunately, uclibc-ng also defines `__GLIBC_PREREQ`: https://github.com/wbx-github/uclibc-ng/blob/ab1dd83bec59c9e65c31efd6e887182948f627be/include/features.h#L391.
So the defines will have to be:
 ```
#if defined(__GLIBC_PREREQ) && !defined(__UCLIBC__)
#if __GLIBC_PREREQ(2,3)
```
It should be noted that `defined(__GLIBC_PREREQ)` can't be in the same if `&&` condition than `__GLIBC_PREREQ(2,3)` as it will result in a build failure if `__GLIBC_PREREQ` is not defined.

egorpugin:
This PR's function check is much simpler than ifdefs you are talking about.

amitdo:
libc in FreeBSD, NetBSD and OpenBSD also supports `feenableexcept()`.

ffontaine:
Then I assume that function checks is indeed the best solution.

PR #3717, opened 2022-01-11
Check if platform supports sys/auxv.h

ffontaine:
sys/auxv.h is not supported by uclibc

### DIFF
--- a/cmake/Configure.cmake
+++ b/cmake/Configure.cmake
@@ -86,6 +86,7 @@ set(include_files_list
     stdint.h
     stdlib.h
     string.h
+    sys/auxv.h
     sys/stat.h
     sys/types.h
     unistd.h
--- a/configure.ac
+++ b/configure.ac
@@ -241,8 +241,10 @@ fi
 # check whether to build OpenMP support
 AC_OPENMP
 
-have_tiff=false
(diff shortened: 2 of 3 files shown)


stweil (changes requested):
I am afraid that Tesseract will be very slow on 32 bit ARM and uClibc with this patch. Please try to find a solution which allows NEON detection with uClibc.

stweil:
Ideally uclibc should be fixed to support the missing functionality instead of compiling Tesseract code with restricted functions (this also applies to pull request #3716).

ffontaine:
Unfortunately, [uclibc-ng](https://uclibc-ng.org) won't be patched any time soon. The project is not very active and dedicated to embedded systems. If you think that this PR doesn't make sense, an other option is to disable tesseract with uclibc on buildroot side.

egorpugin (approved):

stweil:
@egorpugin, please don't merge this PR.

ffontaine:
Would it be acceptable to merge this PR?

It seems that the conclusion of https://github.com/tesseract-ocr/tesseract/pull/3716 is that function checks are better than `#ifdef` blocks.

stweil:
I see no use for ARM support without NEON because that would result in a very slow Tesseract and would prefer disabling the Tesseract build for such platforms on the buildroot side.

Issue #3719, opened 2022-01-14
Feature request: Add timeout and retry parameters for curl_easy_perform


stweil:
A batch of more than 40000 `tesseract` runs with images from the net randomly hangs in `curl_easy_perform`. Maybe this is caused by problems of the server which delivers the images.

By default the processes which have that problem will hang without a time limit.

Adding a new parameter could allow a user configurable timeout. See https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html for details.

Another parameter could specify how often `tesseract` should retry the image download.

PR #3721, opened 2022-01-16
Add new parameter curl_timeout for curl_easy_setop

stweil:

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -99,6 +99,9 @@ namespace tesseract {
 
 static BOOL_VAR(stream_filelist, false, "Stream a filelist from stdin");
 static STRING_VAR(document_title, "", "Title of output document (used for hOCR and PDF output)");
+#ifdef HAVE_LIBCURL
+static INT_VAR(curl_timeout, 0, "Timeout for curl in seconds");
+#endif
 
 /** Minimum sensible image size to be worth running tesseract. */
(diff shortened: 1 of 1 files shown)


stweil:
This addresses parts of feature request #3719.

src/api/baseapi.cpp
```diff
@@ -99,6 +99,9 @@ namespace tesseract {
 
 static BOOL_VAR(stream_filelist, false, "Stream a filelist from stdin");
 static STRING_VAR(document_title, "", "Title of output document (used for hOCR and PDF output)");
+#ifdef HAVE_LIBCURL
+static INT_VAR(curl_timeout, 0, "Timeout for curl in seconds");
```
stweil, line 103: Are there better suggestions for the name of the new parameter?

egorpugin (approved):

egorpugin:
I think, we can refactor this code further. In the separate commit?
```
auto setopt = [error, curl](auto && ... args) {
        auto curlcode = curl_easy_setopt(curl, args...);
        if (curlcode != CURLE_OK) {
          return error("curl_easy_setopt");
        }
        return true;
};
if (!(1
&& setopt(CURLOPT_NOSIGNAL, 1L)
&& setopt(opt2, val2)
&& setopt(opt3, val3)
...
)) return false;
```

stweil:
> I think, we can refactor this code further.

Yes, indeed. That can be done any time later.

PR #3723, opened 2022-01-16
Handle image and line regions in output formats ALTO, hOCR and text

stweil:

### DIFF
--- a/src/api/altorenderer.cpp
+++ b/src/api/altorenderer.cpp
@@ -13,9 +13,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "errcode.h" // for ASSERT_HOST
 #ifdef _WIN32
-#  include "host.h" // windows.h for MultiByteToWideChar, ...
+#  include "host.h"  // windows.h for MultiByteToWideChar, ...
 #endif
+#include "tprintf.h" // for tprintf
 
 #include <tesseract/baseapi.h>
 #include <tesseract/renderer.h>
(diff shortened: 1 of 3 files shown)


stweil:
The size of the output files is significantly reduced by these changes (sizes in MiB for 41998 images):

```
# old code
19489	alto
16838	hocr
844	txt

# new code
18205	alto
15264	hocr
827	txt
```

Shreeshrii:
> // Ignore images and lines for text output.

Will adding a placeholder in the text output for images and lines be
helpful?

stweil:
> Will adding a placeholder in the text output for images and lines be helpful?

Personally I don't need such a placeholder. If I want image information, I use ALTO or hOCR.

But if there are good reasons for a placeholder, it could be implemented as an optional parameter which defines the placeholder text. The default would be an empty string which would not be written. Or the implementation could use a boolean parameter which would be set to true to write a fixed placeholder.

MerlijnWajer:
I'd say a placeholder would be confusing, it's too easy to confuse it with actual text.

MerlijnWajer:
I have tried this code on a few books and it seems to work nicely.

In the meantime Aram developed a rudimentary method to create epubs from the hOCR output (and the input images), which can read `ocr_photo` tags and add them to the epub. This is IMHO a good way to validate that the hOCR ocr_photo elements also make sense.

You can find one such an epub here: https://archive.org/download/sim_english-illustrated-magazine_1884-12_2_15/sim_english-illustrated-magazine_1884-12_2_15.epub 
(Other files likes images or the hOCR file can be found in the directory index: https://archive.org/download/sim_english-illustrated-magazine_1884-12_2_15)

The ocr_photo boxes in this particular epub are not that great (still it does a decent job if you disregard the page edges that it finds as photos), but that is because the input images were not cropped to the pages, so we can't really blame Tesseract for that.

MerlijnWajer:
I would be happy if this makes it into a 5.1.0 release, but I suppose you weren't asking me. In any case, the code seems to work well for me, so you could add my sign off or tested by if that's useful / helpful.

stweil:
Thank you. I added your Tested-by and merged the commit now.

As I don't know how the TSV output is used I won't change that for 5.1.0, but it can be changed any time later.

PR #3745, opened 2022-02-06
Remove unused functions in genericvector.h.

egorpugin:


egorpugin:
@stweil 

Are we allowed to remove internal unused functions, classes, methods, variables?

Like ccutil/bitvector.h has also some of those.
```
int NextSetBit(int prev_bit) const;
...
  // Set subtraction *this = v1 - v2.
  void SetSubtract(const BitVector &v1, const BitVector &v2);
```

I have found that clion shows unused code/includes well.
https://www.jetbrains.com/help/clion/unused-include-directive.html

stweil (approved):
Thanks!

stweil:
> Are we allowed to remove internal unused functions, classes, methods, variables?

Yes, that's fine.

Issue #3751, opened 2022-02-09
CMake problem


OgreTransporter:
When building Tesseract, an error message appears in CMake. Depending on the CMake version, the target OpenMP::OpenMP_CXX is apparently already defined. In any case it is already present in 3.22.2. Therefore you should check if the target already exists.

### Environment

* **Commit Number**: acd6f6c93ff7989f9280dfc6b12869873cfb6cfa 
* **Platform**: Windows 10, x64, Visual Studio 2022

### Current Behavior:
```
>> OpenMP_FOUND TRUE 
CMake Error at CMakeLists.txt:295 (add_library):
  add_library cannot create imported target "OpenMP::OpenMP_CXX" because
  another target with the same name already exists.
```

### Expected Behavior:
```
>> OpenMP_FOUND TRUE 
```
### Suggested Fix:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8d229291..3c7b44b0 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -292,7 +292,9 @@ if(OPENMP_BUILD)
   if(OpenMP_FOUND)
     message(">> OpenMP_FOUND ${OpenMP_FOUND} ${OpenMP_VERSION}")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
-    add_library(OpenMP::OpenMP_CXX IMPORTED INTERFACE)
+       if (NOT TARGET OpenMP::OpenMP_CXX)
+      add_library(OpenMP::OpenMP_CXX IMPORTED INTERFACE)
+       endif()
   endif()
   # https://stackoverflow.com/questions/12399422
   # how-to-set-linker-flags-for-openmp-in-cmakes-try-compile-function
```

PR #3752, opened 2022-02-09
Check OpenMP target in CMake (issue #3751)

OgreTransporter:
When building Tesseract, an error message appears in CMake. Depending on the CMake version, the target OpenMP::OpenMP_CXX is apparently already defined. In any case it is already present in 3.22.2. Therefore you should check if the target already exists.

Solves issue #3751


amitdo:
@egorpugin,  


Please review this PR.

PR #3753, opened 2022-02-09
Partially revert changes of list data types (fix compiler warnings)

stweil:
Changing from class to struct causes clang compiler warnings like this one:

In file included from ../../../src/api/baseapi.cpp:63:
../../../include/tesseract/osdetect.h:29:1: warning: class 'BLOB_CHOICE_LIST' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
class BLOB_CHOICE_LIST;
^
../../../src/ccstruct/ratngs.h:228:1: note: previous use is here
ELISTIZEH(BLOB_CHOICE)
^
../../../src/ccutil/elst.h:804:10: note: expanded from macro 'ELISTIZEH'
  struct CLASSNAME##_LIST : X_LIST<ELIST, ELIST_ITERATOR, CLASSNAME> { \
         ^
<scratch space>:458:1: note: expanded from here
BLOB_CHOICE_LIST
^
../../../include/tesseract/osdetect.h:29:1: note: did you mean struct here?
class BLOB_CHOICE_LIST;
^~~~~

As it is not possible to change the API header tesseract/osdetect.h,
some of the changes from class to struct had to be reverted.

Fixes: 968d653f89d68d ("Shorten macros")

### DIFF
--- a/src/ccstruct/matrix.h
+++ b/src/ccstruct/matrix.h
@@ -39,7 +39,7 @@
 
 namespace tesseract {
 
-struct BLOB_CHOICE_LIST;
+class BLOB_CHOICE_LIST;
 class UNICHARSET;
 
 #define NOT_CLASSIFIED static_cast<BLOB_CHOICE_LIST *>(nullptr)
--- a/src/ccutil/clst.h
+++ b/src/ccutil/clst.h
@@ -703,7 +703,7 @@ public:
 };
 
 #define CLISTIZEH(CLASSNAME)                                    \
-  struct CLASSNAME##_CLIST : X_CLIST<CLASSNAME> {               \
+  class CLASSNAME##_CLIST : public X_CLIST<CLASSNAME> {         \
     using X_CLIST<CLASSNAME>::X_CLIST;                          \
   };                                                            \
   struct CLASSNAME##_C_IT : X_ITER<CLIST_ITERATOR, CLASSNAME> { \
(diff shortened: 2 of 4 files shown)


egorpugin:
`grep BLOB_CHOICE_LIST -r include/*`

You will find that this type is only used in that file.
But on the user side we cannot create it with our public api.
Am I wrong?

egorpugin:
If those methods return BLOB_CHOICE_LIST, then how and where we can use this opaque type?

egorpugin:
Both `detect_blob(BLOB_CHOICE_LIST` use BLOB_CHOICE_LIST as input, therefore this public APIs are useless.
We cannot create BLOB_CHOICE_LIST somehow in user code.

egorpugin:
They should be revised including
```
class BLOBNBOX;
class BLOBNBOX_CLIST;
class BLOB_CHOICE_LIST;
class TO_BLOCK_LIST;
class UNICHARSET;
```

My vote is to remove them as broken.

stweil:
> My vote is to remove them as broken/orphaned.

There is a public function `os_detect_blob` which uses `OrientationDetector` and `ScriptDetector` pointers as argument, and the API allows creating such objects. So unless we are sure that this API cannot be used it must remain in release 5.x.

stweil:
Replacing a few `struct` by `class` like it is done in my pull request looks like the safest solution for me. And for a future release 6.x we already agreed that getting rid of the proprietary list classes is desirable.

egorpugin:
> So unless we are sure that this API cannot be used it must remain in release 5.x.

Yes, it's ok.
The good thing is that we know about some bad public APIs now.

PR #3761, opened 2022-02-26
Fix linker flags for MSYS2 clang64 builds

stweil:
MSYS2 clang64 uses the lld linker which does not support --as-needed.
The normal GNU ld uses that linker option with ELF targets but ignores
it for PE targets (.exe, .dll), so it can be removed.

Remove also the -Wl, which is only needed when linker options are
passed to the compiler but not when they are directly passed to the
linker.


stweil:
MSYS2 currently requires a patch for building: https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-tesseract-ocr/tesseract-no-as-needed.patch. This commit makes that patch unnecessary fro future Tesseract releases.

PR #3765, opened 2022-03-10
cmake: speed up configuration

zdenop:


egorpugin:
Do we have fma checks in runtime?
This PR sets FMA enabled unconditionally for MSVC, but what about processors without it?

zdenop:
AFAIK this PR does not change current logic (at least for VS >=2019): when compiler supports FMA - it was compiled with this feature..
My build with msvc 2019 reports:
```
> tesseract -v
tesseract 5.1.0
 leptonica-1.83.0 (Jan 26 2022, 19:15:03) [MSC v.1929 LIB Release x64]
  libgif 5.2.1 : libjpeg 6b (libjpeg-turbo 2.0.91) : libpng 1.6.37 : libtiff 4.3.0 : zlib 1.2.11
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
 Found OpenMP 2019
 Found libarchive 3.5.1 zlib/1.2.11 liblzma/5.2.4 bz2lib/1.0.6 libzstd/1.4.9
 Found libcurl/7.75.0 zlib/1.2.11 libssh2/1.10.1_DEV
```

Shreeshrii:
cmake Actions CI failed on Ubuntu. Please check if it is related. Thanks!
https://github.com/tesseract-ocr/tesseract/actions/runs/1965506399

zdenop:
This is strange: 

- Ubuntu 18 has problem with gcc and clang
- Ubuntu 20 has problem only with clang
- All other cmake configurations are fine (mac, win64, Ubuntu 20 gcc)

image

I tried to revert 2 suspected option changes (LTO and C language for project), but it does not seems to fix it.

Interesting is that failure is different for gcc and clang:

- gcc fails during compiling time of training tools because of `pthread`: 

image

- clang fails during linging `testing/basicapitest.cpp` (e.g. tesseract build and running `tesseract -v` work fine):

image

amitdo:
You can probably omit the checks for avx2, avx, fma, sse4 for other compilers (in Windows and other OSes), at least for clang and gcc.

PR #3768, opened 2022-03-11
Update ccutil.cpp

### LINKED ISSUE #3767: Compiled-in TESSDATA_PREFIX unused on Windows
Hello!

We tried to compile Tesseract including a compiled-in TESSDATA_PREFIX using a respective preprocessor definition.
However, compiling `ccutil.cpp` on Windows with defined `_WIN32` always causes the code
```
#if defined(_WIN32)
  else if (datadir.empty() || _access(datadir.c_str(), 0) != 0)
  {
    ...
  }
#endif /* _WIN32 */
```
to be executed and, if `datadir` is currently empty, it will be initialized as `<Path to binary>/tessdata`.
Thus, the code that would initialize `datadir` according to the compile-time constant `TESSDATA_PREFIX` is not executed. Most likely on Unix systems this bug is not present because here, `_WIN32` is not defined.

------------------------

### Environment

* **Tesseract Version**: 5.1.0
* **Commit Number**: released version
* **Platform**: MSYS_NT-10.0-19043 64-bit

### Current Behavior:
Compiling Tesseract with defined `TESSDATA_PREFIX=C:\Path\to\somewhere` and starting tesseract.exe with an attached debugger as well as only the supplied command line argument `--list-langs` does not find Tesseract's language files, even if they exist in a folder 'tessdata' in the respective compiled-in directory.

### Expected Behavior:
As long as neither an other data directory is explicitly given nor does the directory `tessdata` next to the executing binary exist but the compiled-in `TESSDATA_PREFIX` does, it should be used.

### Suggested Fix:
In my understanding of the code, the `tessdata` directory should be resolved according to the following rules:
1. Explicitly set by `argv0` overrules everything.
2. Thereafter, the `TESSDATA_PREFIX` environment variable is used.
3. Next, a `tessdata` directory next to the executing binary is used.
4. As the last possible solution, the compiled-in `TESSDATA_PREFIX` should be checked.
5. Return an initialization error if all four alternatives fail (redirect to `./`).

I think the easiest option is to move alternative 3 into a function that on Windows does the same as the `#if defined(_WIN32)` part while on Unix, it does a similar check (even though this is not currently implemented). If the `tessdata` directory next to the binary exists, it is used and saved in the `datadir` variable and the function returns true. Then, the second `else if` simply calls this function and if it returns false, the additional else part used the compiled-in `TESSDATA_PREFIX` if this is set.

CSBVision:
Fixes #3767.

### DIFF
--- a/src/ccutil/ccutil.cpp
+++ b/src/ccutil/ccutil.cpp
@@ -57,7 +57,7 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
 #if defined(_WIN32)
   } else if (datadir.empty() || _access(datadir.c_str(), 0) != 0) {
     /* Look for tessdata in directory of executable. */
-    char path[_MAX_PATH] = {0};
+    char path[_MAX_PATH];
     DWORD length = GetModuleFileName(nullptr, path, sizeof(path));
     if (length > 0 && length < sizeof(path)) {
       char *separator = std::strrchr(path, '\\');
--- a/src/ccutil/ccutil.cpp
+++ b/src/ccutil/ccutil.cpp
@@ -57,27 +57,30 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
 #if defined(_WIN32)
   } else if (datadir.empty() || _access(datadir.c_str(), 0) != 0) {
     /* Look for tessdata in directory of executable. */
-    char path[_MAX_PATH];
+    char path[_MAX_PATH] = {0};
     DWORD length = GetModuleFileName(nullptr, path, sizeof(path));
     if (length > 0 && length < sizeof(path)) {
       char *separator = std::strrchr(path, '\\');
       if (separator != nullptr) {
         *separator = '\0';
-        datadir = path;
-        datadir += "/tessdata";
+        std::string subdir = path;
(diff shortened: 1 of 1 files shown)


src/ccutil/ccutil.cpp
```diff
@@ -57,27 +57,30 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
 #if defined(_WIN32)
   } else if (datadir.empty() || _access(datadir.c_str(), 0) != 0) {
     /* Look for tessdata in directory of executable. */
-    char path[_MAX_PATH];
+    char path[_MAX_PATH] = {0};
```
stweil, line 60: Why is this change needed?

CSBVision: It's not strictly necessary, just found it slightly cleaner this way.

stweil, line 60:
```suggestion
    char path[_MAX_PATH];
```

stweil: It's unnecessary, so let's keep the old code.

stweil:
Thank you.

CSBVision:
Thanks for merging.

PR #3774, opened 2022-03-26
Add CI for openmp performance comparison

Shreeshrii:
See https://github.com/Shreeshrii/tesseract/runs/5701503306?check_suite_focus=true for the log for test run


.github/workflows/autotools-openmp.yml
```diff
@@ -0,0 +1,81 @@
+name: autotools-openmp
+# autotools on Ubuntu - run benchmark test. '--enable-openmp' no training tools
+on:
+  #push:
+  schedule:
+    - cron: 0 20 * * *
```
stweil, line 6: Would it be sufficient to run this (and maybe some other actions, too) only on demand?

Most days there are no changes in the code, so running daily costs resources without any need.

.github/workflows/autotools-openmp.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: |
+           for lmt in {1..3}; do
+                time OMP_THREAD_LIMIT=$lmt tesseract benchmarks.list - --tessdata-dir ../tessdata_best > /dev/null 2>&1 && echo "OMP_THREAD_LIMIT=" $lmt "tessdata_best"
+           done
+
+    - name: Run Tesseract using image from issue 263 with tessdata and OpenMP Thread Limit
+      run: |
+           for lmt in {1..3}; do
+                time OMP_THREAD_LIMIT=$lmt tesseract benchmarks.list - --tessdata-dir ../tessdata > /dev/null 2>&1 && echo "OMP_THREAD_LIMIT=" $lmt "tessdata" 
+           done
```
stweil, line 81:
```suggestion
           done
```

stweil: (missing linefeed in last line)

.github/workflows/autotools-openmp.yml
Shreeshrii: >Would it be sufficient to run this (and maybe some other actions, too) only on demand?

Yes, that's perfectly fine.

stweil: See https://github.com/OCR-D/ocrd_all/blob/c28e3296c4b6321a1e86ea5b0be85767515d1256/.github/workflows/makeall.yml#L16 for an example.

stweil, line 6:
```suggestion
  #schedule:
  #  - cron: 0 20 * * *
  workflow_dispatch:
```

Shreeshrii: Thanks @stweil

.github/workflows/autotools-openmp.yml
Shreeshrii: Not able to commit from here

.github/workflows/autotools-openmp.yml
stweil: Note that the resource costs of daily CI actions can become really huge if those actions run also on hundreds (or even thousands) of forks. This can be reduced by limiting the daily actions to the primary repository only. Offering a manual trigger would still allow forks to run CI on their code, too.

Shreeshrii: >This can be reduced by limiting the daily actions to the primary repository only. 

How do I do that?

PR #3775, opened 2022-03-26
Add benchmark test with disable-openmp

Shreeshrii:
Using same image list as in https://github.com/tesseract-ocr/tesseract/pull/3774
This PR adds it to tesseract built with diable-openmp


Shreeshrii:
See https://github.com/Shreeshrii/tesseract/runs/5701698594?check_suite_focus=true for log from a test run

Issue #3283, opened 2021-02-05
libtesseract on MSVC 2019 32 bit requires compiler optimization switched off


BJungmann:
### Environment

* **Tesseract Version**: 5.0.0-alpha-20201231
* **Platform**: W10 64-bit
* compiler MSVC 2019, configuration Win32, optimization flag /O2

### Current Behavior:
If I make tesseract50.lib with above settings and link it into my own Qt Creator project, recognition results are very poor. They do have some relation to the source image, but are unusable.
If I make tesseract50.lib with optimization flag /Od, deactivated, inline functions may still have /Ob2, the recognition results are OK.

### Expected Behavior:
This behaviour should not depend on the optimization flags.

This problem does not show up if I compile everything for x64 (just now succeeded in trying that).

amitdo:
Does this issue occur with Tesseract command line program?

egorpugin:
What about 64-bit build?

BJungmann:
64-bit build is OK, I said it already.

Currently I have not test environment for the command line program. But I guess the problem could be related to borders of rectangles. See an example of a rectangle for TessBaseAPI::SetRectangle(), and the returned ocrx_word rectangles:
![Issue3283_1](https://user-images.githubusercontent.com/16709562/107078220-4a381980-67ee-11eb-9eeb-5b2f2637a084.png)
![Issue3283_2](https://user-images.githubusercontent.com/16709562/107078247-558b4500-67ee-11eb-8ac7-15587d402b3c.png)

egorpugin:
Yes, indeed. Sorry.

---

@stweil @zdenop 
Do we want to support 32-bit builds on windows at all?

zdenop:
If there is no one from contributors able and willing to maintain 32bit build on windows, I would suggest to marked it as unsupported or experimental.

Anyway I would suggest to try [minimal version of tesseract](https://spell.linux.sk/building-minimalistic-tesseract) if problem is really in tesserac&leptonica part. 
Other option is to try build [tesserac&leptonica with clang](https://spell.linux.sk/building-tesseract-and-leptonica-with-cmake-and-clang-on-windows) which is able to build VS library...

BJungmann:
I have now reproduced the problem with Tesseract command line program. With /O2, the recognition result of the first attached image was
Njotker Balbulus/Martin jLuther
[...]

PR #3778, opened 2022-03-30
Set /Os for some 32 bit MS compilers (fixes #3769)

### LINKED ISSUE #3769: Invalid characters with Tesseract 5.1.0 and tessdata_fast data (for German version) when using 32-bit Microsoft compiler
### Environment

* **Tesseract Version**: 5.1.0
* **Platform**: Windows 32-bit 

### Current Behavior:
I have the following problem:
1. I prepared a custom build for Tesseract 5.1.0, so as to generate dlls, which I then use in the project of a 32-bit .exe application.
2. I prepared the following dependencies with CMake 3.23 (without SW build): 
a. tesseract 5.1.0, leptonica 1.82.0, libtiff 4.3.0, libjpeg-turbo 2.1.3, zlib 1.2.11, libpng 1.6.37.
b. Links to src:
- tesseract 5.1.0 (https://github.com/tesseract-ocr/tesseract/releases) – 01.03.2022
- leptonica 1.82.0 (http://www.leptonica.org/download.html) – 22.09.2021
- libtiff 4.3.0 (http://download.osgeo.org/libtiff) – 20.04.2021
- libjpeg-turbo 2.1.3 (https://github.com/libjpeg-turbo/libjpeg-turbo/releases ) – 25.02.2022
- zlib 1.2.11 (https://github.com/madler/zlib/tags) – 15.01.2017
- libpng 1.6.37 (https://github.com/glennrp/libpng/releases/tag/v1.6.37) – 14.04.2019
c. I also fix CMakeList.txt a bit for tesseract to be able to generate .dll files - see: 
[CMakeLists.txt](https://github.com/tesseract-ocr/tesseract/files/8284026/CMakeLists.txt)
3. After generating the dependencies, I used them in a wrapper that uses CAPI and generated a dll file (also 32 bit) that I used in the application. The list of all dependencies is as follows:
image
4. In the next step, I performed an OCR test in the application with tessdata germany data - deu.traineddata model: https://github.com/tesseract-ocr/tessdata_fast. 
5. At this point, I noticed inferior recognition quality compared to the Tesseract 4.1.1 version, which I used earlier. 
image
a. test file: 
test_file
6. I noticed that there is also a problem with slash, for example: It is then changed to "jj" - see:
image
7. I would like to add that I have also prepared a Tesseract 4.1.1 compilation with the dependencies as in point 2b. The quality of OCR did not change then.
8. I use tessdata_best as a temporary workaround (and it work), but the OCR speed for this model is not satisfactory for me.

### Expected Behavior:
I expect Tesseract 5.1.0 to recognize characters correctly, ie not converting "l", "m" to "j" or "i" to "j" for example in the tessdata_fast mode. I would like character recognition to work similar to Tesseract 4.1.1.

### Suggested Fix:
Consideration of an upgrade for deu.traineddata models on the website:
https://github.com/tesseract-ocr/tessdata_fast

stweil:

### DIFF
--- a/src/arch/intsimdmatrixavx2.cpp
+++ b/src/arch/intsimdmatrixavx2.cpp
@@ -27,6 +27,14 @@
 #  include <cstdint>
 #  include <vector>
 
+#  if defined(_MSC_VER) && _MSC_VER >= 1925 && _MSC_VER <= 1929 && \
+      defined(_WIN32) && !defined(_WIN64)
+// Optimize for size (/Os) instead of using the default optimization for some
+// versions of the 32 bit Visual Studio compiler which generate buggy code.
+#    pragma optimize("", off)
+#    pragma optimize("s", on)
+#  endif
+
 namespace tesseract {
 
 // Number of outputs held in each register. 8 x 32 bit ints.


stweil:
@BJungmann, @krzysiekj94, does this patch work for you? I tried to fix the issues reported in #3283 and #3769, but cannot test it myself.

BJungmann:
My suggestion was a patch in simddetect.cpp after line 199. This would execute a little faster than your suggested patch. I did not check the _MSC_VER limits, though.

`      if (max_function_id >= 7) {
        __cpuid(cpuInfo, 7);
        avx2_available_ = (cpuInfo[1] & 0x00000020) != 0;
# if defined(_MSC_VER) && _MSC_VER >= 1925 && _MSC_VER <= 1929
        // forbid usage of AVX2 hardware for some
        // versions of the 32 bit Visual Studio compiler which generate buggy code.
        avx2_available_ = false;
# endif
        avx512F_available_ = (cpuInfo[1] & 0x00010000) != 0;
        avx512BW_available_ = (cpuInfo[1] & 0x40000000) != 0;
      }
`

PR #3780, opened 2022-03-31
Add C API to Init with trained data from memory

### LINKED ISSUE #3691: Feature Request: Add a way to set the traineddata file from memory in the C API
I found a constructor in the C++ API that takes a pair of ptr, size arguments for the training data, but none for the C API. It would be nice to have one in the C API.

nico-abram:
Fixes #3691

### DIFF
--- a/include/tesseract/capi.h
+++ b/include/tesseract/capi.h
@@ -231,6 +231,12 @@ TESS_API int TessBaseAPIInit4(TessBaseAPI *handle, const char *datapath,
                               char **vars_values, size_t vars_vec_size,
                               BOOL set_only_non_debug_params);
 
+TESS_API int TessBaseAPIInit5(TessBaseAPI *handle, const char *data, int data_size,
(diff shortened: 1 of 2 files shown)


nico-abram:
Is the sw harfbuzz package broken? From the appveyor logs:
```
[org.sw.demo.behdad.harfbuzz-4.2.0]/src/harfbuzz.cc
C:\Users\appveyor\.sw\storage\pkg\7d\cb\ea3f\src\sdir\src\hb-ot-layout-gsub-table.hh(32): fatal error C1083: Cannot open include file: 'OT/Layout/GSUB/GSUB.hh': No such file or directory
command failed: exit code = 2
Total errors: 1
```

nico-abram:
Pushed an empty commit to rerun CI, looks like it passed now, so there was probably a problem with it that I assume you've fixed now

amitdo:
@egorpugin, can you fix sw?

egorpugin:
I'll check it later this week.

PR #3782, opened 2022-04-02
src/arch/simddetect.cpp: fix NEON detection on FreeBSD

clausecker:
FreeBSD does not have the `<asm/hwcap>` header, requiring some slightly different code to check for NEON availability.  This code might also apply to other BSD-like operating systems such as NetBSD, OpenBSD, DragonflyBSD, or iOS, but I haven't checked.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -163,7 +163,7 @@ case "${host_cpu}" in
 
     ;;
 
-  aarch64)
+  aarch64|arm64)
 
     # ARMv8 always has NEON and does not need special compiler flags.
     AM_CONDITIONAL([HAVE_NEON], true)
@@ -178,6 +178,7 @@ case "${host_cpu}" in
       AC_DEFINE([HAVE_NEON], [1], [Enable NEON instructions])
       NEON_CXXFLAGS="-mfpu=neon"
       AC_SUBST([NEON_CXXFLAGS])
+      check_for_neon=1
     fi
 
     ;;
@@ -190,6 +191,16 @@ esac
 
 # check whether feenableexcept is supported. some C libraries (e.g. uclibc) don't.
 AC_CHECK_FUNCS([feenableexcept])
+
+# additional checks for NEON targets
+if test x$check_for_neon = x1; then
+  AC_MSG_NOTICE([checking how to detect NEON availability])
+  AC_CHECK_FUNCS([getauxval elf_aux_info android_getCpuFamily])
+
+  if test $ac_cv_func_getauxval = no && test $ac_cv_func_elf_aux_info = no && test $ac_cv_func_android_getCpuFamily = no; then
+      AC_MSG_WARN([NEON is available, but we don't know how to check for it.  Will not be able to use NEON.])
+  fi
+fi
 
 AX_CHECK_COMPILE_FLAG([-fopenmp-simd], [openmp_simd=true], [openmp_simd=false], [$WERROR])
 AM_CONDITIONAL([OPENMP_SIMD], $openmp_simd)
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -53,12 +53,12 @@
 #endif
 
 #if defined(HAVE_NEON) && !defined(__aarch64__)
-#  ifdef ANDROID
+#  if HAVE_ANDROID_GETCPUFAMILY
 #    include <cpu-features.h>
-#  elif defined(__linux__)
+#  elif HAVE_GETAUXVAL
 #    include <asm/hwcap.h>
 #    include <sys/auxv.h>
-#  elif defined(__FreeBSD__)
+#  elif HAVE_ELF_AUX_INFO
 #    include <sys/auxv.h>
 #    include <sys/elf.h>
 #  endif
@@ -212,15 +212,15 @@ SIMDDetect::SIMDDetect() {
 #endif
 
 #if defined(HAVE_NEON) && !defined(__aarch64__)
-#  ifdef ANDROID
+#  if HAVE_ANDROID_GETCPUFAMILY
   {
     AndroidCpuFamily family = android_getCpuFamily();
     if (family == ANDROID_CPU_FAMILY_ARM)
       neon_available_ = (android_getCpuFeatures() & ANDROID_CPU_ARM_FEATURE_NEON);
   }
-#  elif defined(__linux__)
+#  elif HAVE_GETAUXVAL
   neon_available_ = getauxval(AT_HWCAP) & HWCAP_NEON;
-#  elif defined(__FreeBSD__)
+#  elif HAVE_ELF_AUX_INFO
   unsigned long hwcap = 0;
   elf_aux_info(AT_HWCAP, &hwcap, sizeof hwcap);
   neon_available_ = hwcap & HWCAP_NEON;
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -55,7 +55,7 @@
 #if defined(HAVE_NEON) && !defined(__aarch64__)
 #  if defined(HAVE_ANDROID_GETCPUFAMILY)
 #    include <cpu-features.h>
-#  elif HAVE_GETAUXVAL
+#  elif defined(HAVE_GETAUXVAL)
 #    include <asm/hwcap.h>
 #    include <sys/auxv.h>
 #  elif HAVE_ELF_AUX_INFO
--- a/src/arch/simddetect.cpp
+++ b/src/arch/simddetect.cpp
@@ -58,7 +58,7 @@
 #  elif defined(HAVE_GETAUXVAL)
 #    include <asm/hwcap.h>
 #    include <sys/auxv.h>
-#  elif HAVE_ELF_AUX_INFO
+#  elif defined(HAVE_ELF_AUX_INFO)
 #    include <sys/auxv.h>
 #    include <sys/elf.h>
 #  endif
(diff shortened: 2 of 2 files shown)


src/arch/simddetect.cpp
```diff
@@ -85,7 +87,7 @@ SIMDDetect SIMDDetect::detector;
 bool SIMDDetect::neon_available_ = true;
 #elif defined(HAVE_NEON)
 // If true, then Neon has been detected.
-bool SIMDDetect::neon_available_;
+bool SIMDDetect::neon_available_ = false;
```
stweil, line 90: Why this change?

src/arch/simddetect.cpp
```diff
@@ -55,10 +55,12 @@
 #if defined(HAVE_NEON) && !defined(__aarch64__)
 #  ifdef ANDROID
 #    include <cpu-features.h>
-#  else
-/* Assume linux */
+#  elif defined(__linux__)
 #    include <asm/hwcap.h>
 #    include <sys/auxv.h>
+#  elif defined(__FreeBSD__)
```
stweil, line 61: I'd prefer to check for Android first, then FreeBSD, then the rest (no explicit check for Linux). That's closer to the current code and perhaps covers more cases.

src/arch/simddetect.cpp
```diff
@@ -216,9 +218,12 @@ SIMDDetect::SIMDDetect() {
     if (family == ANDROID_CPU_FAMILY_ARM)
       neon_available_ = (android_getCpuFeatures() & ANDROID_CPU_ARM_FEATURE_NEON);
   }
-#  else
-  /* Assume linux */
+#  elif defined(__linux__)
```
stweil, line 221: See comment above.

stweil (changes requested):

src/arch/simddetect.cpp
clausecker: This change makes it so that when compiling for an unknown operating system, `neon_available_` defaults to `false` instead of the build failing.  It is connected to the changed further below where now `neon_available_` is no longer set when compiling for an unknown operating system.

src/arch/simddetect.cpp
clausecker: If you do it that way, the build will fail on non-Linux operating systems.  I have specifically designed this patch to permit compilation on operating systems for which we don't know how to obtain the ELF aux vectors, falling back to the safe assumption that NEON is not supported.

src/arch/simddetect.cpp
stweil: Without your change it is also initialized to `false`.

clausecker: Yes.  This initialisation is overridden in the constructor if running on a platform on which we know how to detect the availability of NEON.

src/arch/simddetect.cpp
stweil: I don't think that the code should silently assume that NEON is unsupported because that might result in code which is slower than necessary. Failing seems better to me, because it allows investigating and fixing the problem.

stweil: As we are talking about 32 bit ARM systems (which typically are not very fast), using NEON is especially important to get at least an acceptable performance.

clausecker: Ideally, a configure test would be used to check which way of finding NEON support is available.  Just failing the build seems nasty to users of other operating systems.  I mean I can change the patch like that if you insist.

stweil: Yes, a configure test (which only runs if necessary) would be the best solution.

clausecker: I've never designed a configure test before.  I'll try to figure it out and report back.

clausecker: I have added configure tests as directed.  Please also check and revise the `AC_CONDITIONAL` usage.  I don't think these are meant to be redefined.

src/arch/simddetect.cpp
stweil, line 90:
```suggestion
bool SIMDDetect::neon_available_;
```

configure.ac
```diff
@@ -163,7 +163,7 @@ case "${host_cpu}" in
 
     ;;
 
-  aarch64)
+  aarch64|arm64)
```
stweil, line 166: Which platforms use `arm64` instead of `aarch64`?

configure.ac
```diff
@@ -178,6 +178,7 @@ case "${host_cpu}" in
       AC_DEFINE([HAVE_NEON], [1], [Enable NEON instructions])
       NEON_CXXFLAGS="-mfpu=neon"
       AC_SUBST([NEON_CXXFLAGS])
+      check_for_neon=1
```
stweil, line 181: I suggest to include the code from below here. Then there is no need for `check_for_neon`.

src/arch/simddetect.cpp
```diff
@@ -210,15 +212,18 @@ SIMDDetect::SIMDDetect() {
 #endif
 
 #if defined(HAVE_NEON) && !defined(__aarch64__)
-#  ifdef ANDROID
+#  if HAVE_ANDROID_GETCPUFAMILY
```
stweil, line 215:
```suggestion
#  if defined(HAVE_ANDROID_GETCPUFAMILY)
```

stweil: `HAVE_ANDROID_GETCPUFAMILY` is not always defined, so a simple `#if HAVE_ANDROID_GETCPUFAMILY` could trigger compiler warnings.

src/arch/simddetect.cpp
```diff
@@ -210,15 +212,18 @@ SIMDDetect::SIMDDetect() {
...
   {
     AndroidCpuFamily family = android_getCpuFamily();
     if (family == ANDROID_CPU_FAMILY_ARM)
       neon_available_ = (android_getCpuFeatures() & ANDROID_CPU_ARM_FEATURE_NEON);
   }
-#  else
-  /* Assume linux */
+#  elif HAVE_GETAUXVAL
```
stweil, line 221:
```suggestion
#  elif defined(HAVE_GETAUXVAL)
```

src/arch/simddetect.cpp
```diff
@@ -210,15 +212,18 @@ SIMDDetect::SIMDDetect() {
...
   neon_available_ = getauxval(AT_HWCAP) & HWCAP_NEON;
+#  elif HAVE_ELF_AUX_INFO
```
stweil, line 223:
```suggestion
#  elif defined(HAVE_ELF_AUX_INFO)
```

configure.ac
clausecker: FreeBSD does for example.  I have not checked other platforms.

src/arch/simddetect.cpp
clausecker: An undefined macro evaluates to `0` in a `#if` preprocessor test.  No warning is emitted in this case.

configure.ac
clausecker: Which code do you mean specifically?

clausecker:
Please consider opening a code review next time so the system doesn't send me one email for each comment you wrote.

src/arch/simddetect.cpp
stweil: The GNU C++ compiler and clang emit warnings (depending on warning level). See https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html (-Wundef).

configure.ac
stweil: Lines 197–202.

clausecker: Line 196 checks for `check_for_neon`.  The point of this variable is to only run the NEON checks on platforms where NEON could possibly be available.  You [specifically requested](https://github.com/tesseract-ocr/tesseract/pull/3782#discussion_r841081250) that the test be implemented that way.

stweil: Right. And now I suggest to move lines 197–202 to replace line 181 in your commit. That also runs the NEON checks only when needed, but simplifies and reduces the code for that.

clausecker: There was some sort of reason why I did it the way I did it.  I can try to check if it works the way you indicate, but it has been almost two months since I last touched this changeset and it will be difficult for me to re-run all the tests and ensure it still works correctly.  It would be a lot easier for me just to commit it as is.

stweil: Then please add at least all other suggested changes.

src/arch/simddetect.cpp
```diff
@@ -53,12 +53,14 @@
 #endif
 
 #if defined(HAVE_NEON) && !defined(__aarch64__)
-#  ifdef ANDROID
+#  if HAVE_ANDROID_GETCPUFAMILY
```
stweil, line 56:
```suggestion
#  if defined(HAVE_ANDROID_GETCPUFAMILY)
```

src/arch/simddetect.cpp
```diff
@@ -53,12 +53,14 @@
...
 #    include <cpu-features.h>
-#  else
-/* Assume linux */
+#  elif HAVE_GETAUXVAL
```
stweil, line 58:
```suggestion
#  elif defined(HAVE_GETAUXVAL)
```

src/arch/simddetect.cpp
```diff
@@ -53,12 +53,14 @@
...
 #    include <asm/hwcap.h>
 #    include <sys/auxv.h>
+#  elif HAVE_ELF_AUX_INFO
```
stweil, line 61:
```suggestion
#  elif defined(HAVE_ELF_AUX_INFO)
```

stweil (approved):
Thanks!

stweil:
While this pull request was an important fix for ARM hosts, we had a similar severe problem for amd64 / x86_64 host. See issue #3964.

PR #3785, opened 2022-04-06
add link for web demo on Hugging Face Spaces

AK391:
added link for web demo on Hugging Face Spaces using Gradio by https://huggingface.co/tomofi

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Table of Contents
 * [Tesseract OCR](#tesseract-ocr)
    * [About](#about)
    * [Brief history](#brief-history)
+   * [Web Demo](#Web-Demo)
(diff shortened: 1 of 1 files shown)


stweil:
I don't think that it is a good idea to add this external link to such a prominent place as the official project README, especially because the Tesseract sample on that web demo uses an inadequate recognition model, so it shows a really bad OCR result.

zdenop:
That is not tesseract-ocr organization project. The right place for promoting your project is https://github.com/tesseract-ocr/tessdoc/blob/main/User-Projects-%E2%80%93-3rdParty.md

stweil:
I'm not even sure whether we want to advertise commercial offers on `tessdoc`. Maybe better not.

PR #3787, opened 2022-04-10
RFC: Improve positioning of symbol bounding boxes

### LINKED ISSUE #1712: Incorrect word coordinates
### Environment

* **Tesseract Version**:

      tesseract --version
      tesseract 4.0.0-beta.1
      leptonica-1.76.0 (Jun 26 2018, 18:21:40) [MSC v.1900 LIB Release x86]
       libgif 5.1.4 : libjpeg 9b : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
      Found AVX
      Found SSE
    (says beta.1, but beta.3 seems to be correct)
* **Commit Number**: AppVeyor: 4.0.0-beta.3.1776
* **Platform**: Windows 10 64bit (but tesseract running as 32bit)
* **Tessdata**: tessdata-fast

We've integrated the engine in our (closed-source) application using the C API, so I cannot share the actual code.

What I do is basically just iterating over the result iterator using `RIL_WORD`, get the text, bounding box and baseline for each word and then creating a PDF out of it, with the recognized text as a red overlay and drawing the bounding boxes in green for better visibility.

But the official PDF output config has the same flaw, it's just more difficult to spot:
```
tesseract andromeda.png andromeda.tess4cli pdf
```
produces the following PDF:
[andromeda.tess4cli.pdf](https://github.com/tesseract-ocr/tesseract/files/2144491/andromeda.tess4cli.pdf)

### Files:
Behavior can be reproduced using the following PNG (which is a part of a bigger file):

andromeda

### Current Behavior:
The coordinates are correct for most words.
But for some words, there seems to be an error in computing the boundaries between the words.
Couriously, the wrong boundary is always before the last character of the previous word.
I almost looks like some kind of off-by-one error.

grafik

grafik

### Expected Behavior:
Result with tesseract 3:

grafik

grafik

p12tic:
This PR improves the positions of symbol bounding boxes in cases when LSTM model is used. Up to 20 times less errors have been observed in sample images. 

This PR still requires potentially significant amount of work, please let me know whether the approach is sensible in principle and if the PR makes sense I will spend time polishing it.

--------

When using LSTM models the accuracy of character bounding boxes is low with many blobs assigned to wrong characters. This is caused by the fact that LSTM model output produces only approximate character positions without boundary data. As a result the input blobs cannot be accurately mapped to characters and which compromises the accuracy of character bounding boxes.

Currently this problem is solved as follows. The character boundaries are computed according to the character positions from the LSTM output by placing the boundaries at the middle between two character positions. The blobs are then assigned according to which character the center of the blob falls to. In other words the blobs are assigned to the nearest characters.

This unfortunately produces a lot of errors because the character positions in the LSTM output have a tendency to drift, thus the nearest character is often not the right one.

Fortunately while the LSTM model produces approximate positions, the blob boundaries produced by the regular segmenter are pretty good. Most of the time a single blob corresponds to a single character and vice-versa.

The above is used to create an optimization algorithm that treats the output of the regular segmenter as a template to which LSTM model output is matched. The selection of best match is done by assigning each unwanted property of the outcome a cost and then minimizing the total cost of the solution.

This reliably solves the most frequent error present in the current solution when blobs are simply assigned to wrong character. As a result the current algorithm produces up to 20 times less errors.

This can be further improved because the root cause of most of the remaining errors is the segmenter producing single blobs for multiple characters. The algorithm could be improved by biasing it to split the blobs in the places where the segmenter makes errors often, such as near character "t". 

Fixes https://github.com/tesseract-ocr/tesseract/issues/1712.

----------

One of the example images I've used:
loremipsum_large

Before this PR, tesseract produced 116 errors in determining character bounding boxes (could be inaccurate due to manual counting)
loremipsum_large-boxes_old

After this PR, tesseract produced only 5 errors in determining character bounding boxes.
loremipsum_large-boxes

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -250,6 +250,7 @@ endif
 # Rules for src/ccstruct.
 
 noinst_HEADERS += src/ccstruct/blamer.h
+noinst_HEADERS += src/ccstruct/blob_bounds_calculator.h
 noinst_HEADERS += src/ccstruct/blobbox.h
 noinst_HEADERS += src/ccstruct/blobs.h
 noinst_HEADERS += src/ccstruct/blread.h
@@ -293,6 +294,7 @@ noinst_HEADERS += src/ccstruct/params_training_featdef.h
 endif
 
 libtesseract_la_SOURCES += src/ccstruct/blamer.cpp
+libtesseract_la_SOURCES += src/ccstruct/blob_bounds_calculator.cpp
 libtesseract_la_SOURCES += src/ccstruct/blobbox.cpp
 libtesseract_la_SOURCES += src/ccstruct/blobs.cpp
 libtesseract_la_SOURCES += src/ccstruct/blread.cpp
@@ -1197,6 +1199,7 @@ if !DISABLED_LEGACY_ENGINE
 check_PROGRAMS += bitvector_test
 endif # !DISABLED_LEGACY_ENGINE
 endif # ENABLE_TRAINING
+check_PROGRAMS += blob_bounds_calculator_test
 check_PROGRAMS += cleanapi_test
 check_PROGRAMS += colpartition_test
 if ENABLE_TRAINING
@@ -1308,6 +1311,10 @@ bitvector_test_SOURCES = unittest/bitvector_test.cc
 bitvector_test_CPPFLAGS = $(unittest_CPPFLAGS)
 bitvector_test_LDADD = $(TRAINING_LIBS)
 endif # !DISABLED_LEGACY_ENGINE
+
+blob_bounds_calculator_test_SOURCES = unittest/blob_bounds_calculator_test.cc
+blob_bounds_calculator_test_CPPFLAGS = $(unittest_CPPFLAGS)
+blob_bounds_calculator_test_LDADD = $(TESS_LIBS)
 
 cleanapi_test_SOURCES = unittest/cleanapi_test.cc
 cleanapi_test_CPPFLAGS = $(unittest_CPPFLAGS)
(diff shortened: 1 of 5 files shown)


p12tic:
@stweil Just a friendly ping :-)

rmast:
I've done some testing with this branch merged to the current main. It's still not perfect.
With oem 0:
image
With LSTM:
image

Original image:
out1496-3078-212-39

Program code of bounding boxes:
```
import tesserocr
import csv
import numpy as np
import cv2
import pytesseract
from PIL import Image
import matplotlib as mpl

import matplotlib.pyplot as plt
img3 = cv2.imread('/home/rmast/plaatjes/out1496-3078-212-39.png')
h, w, _ = img3.shape # assumes color image
pytess_result = pytesseract.image_to_boxes(img3, lang='nld+lat+Latin+eng',
        config="--psm 7 -c tessedit_create_boxfile=1", output_type=pytesseract.Output.DICT)
        #config="--psm 7 --oem 0 -c tessedit_create_boxfile=1", output_type=pytesseract.Output.DICT)
print(pytess_result)
for j in range(0, len(pytess_result["char"])):
    left = pytess_result["left"][j]
    bottom = pytess_result["bottom"][j]
    right = pytess_result["right"][j]
    top = pytess_result["top"][j]
    cv2.rectangle(img3, (left, h - top - 1), (right, h - bottom - 1), (255, 0, 0), 1)
mpl.use('tkAgg')
plt.imshow(img3)
plt.show()
```

lv-saharan:
why still not merged

rmast:
I guess lack of testing capacity for core functionality. Since I discovered that and some other low level segmentation bugs my focus has shifted to EasyOCR.

Verzonden vanaf Outlook voor Android<https://aka.ms/AAb9ysg>
________________________________

PR #3792, opened 2022-04-15
Add initial support for Intel AVX512F

stweil:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,6 +167,10 @@ if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86|x86_64|AMD64|amd64|i386|i686")
     set(AVX2_COMPILE_FLAGS "/arch:AVX2")
     add_definitions("-DHAVE_AVX2")
 
+    set(HAVE_AVX512F ON)
+    set(AVX512_COMPILE_FLAGS "/arch:AVX512")
+    add_definitions("-DHAVE_AVX512F")
+
     set(HAVE_FMA ON)
     set(FMA_COMPILE_FLAGS "-D__FMA__")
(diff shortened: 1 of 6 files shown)


stweil:
This implementation improves the performance for recognition with best models and for training. `intsimdmatrixavx512.cpp` for fast models is still to be implemented.

zdenop:
thanks!

wollmers:
@stweil Just for interest: Can you give a rough measure of of throughput increase (or runtime decrease) in % of a complete run compared to AVX2 (or no SIMD)? Just to get a feeling for the magnitude.

stweil:
I still only have numbers for my standard testcase `lstm_squashed_test`. For that test which only partially uses the float dotproduct the execution time is reduced by about 10 %. But the performance also seems to depend on other factors. For example the test runs slower directly after a build process which exercised all CPU cores, so maybe the Intel Core i3 CPU uses a higher clock rate for the single threaded test as long as the CPU temperature is within some limit.

The throughput is still much better on Apple M1.

Matthew-Bradley:
> The throughput is still much better on Apple M1.

What acceleration, if any, is being used on the M1?

stweil:
M1 currently uses Neon (like other ARM64 platforms, too).

amitdo:
This PR was referenced here:

https://www.phoronix.com/scan.php?page=news_item&px=Tesseract-OCR-5.2-Released

amitdo:
You didn't mentioned the model/spec of the AVX512 machine and the AVX2 machine.

amitdo:
Another thing, did you try to do loop unrolling like is done in AVX2?

stweil:
No, the current code does not do loop unrolling. Nor does AVX code for float do that.

PR #3808, opened 2022-05-02
Pdf offset fix

mcsjosh:
PDF Xref offsets can be up to 10GB.  This fix ensures that TessPDFRenderer uses a data type large enough to store these offsets.


egorpugin:
Please, keep the value signed.

stweil:
> Please, keep the value signed.

Why is a signed offset value better here?

egorpugin:
Unsigned type can break various comparisons/overflows/loops over the codebase.
Since we do not review them in this PR, it's better to keep old behavior.

stweil:
I did have a look on the code before I suggested to use `uint64_t`. The modified value `offset_` is only used in very few places in `pdfrenderer.cpp`:

```
include/tesseract/renderer.h:  std::vector<long int> offsets_; // offset of every PDF object in bytes
src/api/pdfrenderer.cpp:  offsets_.push_back(0);
src/api/pdfrenderer.cpp:  offsets_.push_back(objectsize + offsets_.back());
src/api/pdfrenderer.cpp:  offsets_[kPagesObjectNumber] = offsets_.back(); // manipulation #1
src/api/pdfrenderer.cpp:  offsets_.back() += pages_objsize; // manipulation #2
src/api/pdfrenderer.cpp:    stream << offsets_[i] << " 00000 n \n";
src/api/pdfrenderer.cpp:         << offsets_.back() << "\n%%EOF\n";
```

egorpugin:
Well, 'u' may be added if everything is ok, sorry.

PR #3847, opened 2022-06-21
capi: Fix calling delete[] for memory allocated by malloc

danpla:
`strdup` returns memory allocated via `malloc`, so either `TessDeleteTextArray` should use `free` or `strdup` should be replaced by `new []`.


stweil (approved):
Thank you!

PR #3851, opened 2022-06-24
Use Leptonica API to access internals of Box

stweil:

### DIFF
--- a/src/training/pango/boxchar.cpp
+++ b/src/training/pango/boxchar.cpp
@@ -84,8 +84,10 @@ void BoxChar::TranslateBoxes(int xshift, int yshift, std::vector<BoxChar *> *box
   for (auto &boxe : *boxes) {
     Box *box = boxe->box_;
     if (box != nullptr) {
-      box->x += xshift;
-      box->y += yshift;
+      int32_t box_x;
+      int32_t box_y;
+      boxGetGeometry(box, &box_x, &box_y, nullptr, nullptr);
+      boxSetGeometry(box, box_x + xshift, box_y + yshift, -1, -1);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


stweil:
See also https://github.com/tesseract-ocr/tesseract/issues/3673#issuecomment-1164781686.

Please review carefully (the modified code is not covered by unittests, and we don't want to introduce new bugs).

amitdo:
I reviewed this patch. I looks right, but too verbose :(

I approve it.

egorpugin:
I think we need wrappers here.
```
auto box_get_geometry(Box *b) {
   struct coords {
      int32_t x;
      int32_t y;
      int32_t w;
      int32_t h;
   } c;
   boxGetGeometry(b, &c.x, &c.y, &c.w, &c.h);
   return c;
}
// and usage:
auto &&next = box_get_geometry(b);
auto &&prev = box_get_geometry(b);
auto &&[x,y,w,h] = box_get_geometry(b);
// etc.
```

amitdo:
Another option is to add

```
struct boxGeometry {
      int32_t x;
      int32_t y;
      int32_t w;
      int32_t h;
   };
```

and 

 `struct boxGetGeometryAsStruct()`

to Leptonica's public interface.

In C++17 code we can use structure binding (like Egor's example).

stweil:
So maybe we should postpone this change until a new major release which is using C++17.

egorpugin:
Next major release can jump into C++20 already.

amitdo:
@stweil, I don't understand your comment. We already require C++17 since 5.0.0.

p12tic:
Patch looks good to me.

zdenop:
Can we merge this or are we waiting for something?

PR #3864, opened 2022-07-09
Add CMake option DISABLE_OPTIMIZATION

theirix:
Option allows to build Tesseract without platform-specific optimizations
to run on generic platform without AVX extensions. By default option is OFF
to ensure backward compatibility.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,7 @@ option(USE_SYSTEM_ICU "Use system ICU" OFF)
 option(DISABLE_ARCHIVE "Disable build with libarchive (if available)" OFF)
 option(DISABLE_CURL "Disable build with libcurl (if available)" OFF)
 option(INSTALL_CONFIGS "Install tesseract configs" ON)
+option(DISABLE_OPTIMIZATION "Disable platform-specific optimizations" OFF)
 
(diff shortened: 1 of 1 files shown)


zdenop:
If the platform/CPU does not support specific AVX instructions they are not used.  So what kind of problem does it solve?

theirix:
Autodetecting CPU features in `simddetect.cpp` controls only Tesseract behavior but not compiler's. If we have a recent enough build server with the CPU supporting AVX, compiler will be invoked with `-mavx` flag so the binary will not be able to run at older machine without AVX because compiler can emit AVX instructions. The same story is with `-march=native`. Proposed flag allows to build Tesseract without any avx/fma/sse optimizations.

stweil:
If the cmake build still does that, that's not intentional and should be fixed. autoconf based builds will run on any host.

amitdo:
>Autodetecting CPU features in simddetect.cpp controls only Tesseract behavior but not compiler's.

But the Cmake and autotools build definitions and the code in the dotproduct files do compile time checks.

>If we have a recent enough build server with the CPU supporting AVX, compiler will be invoked with -mavx flag so the binary will not be able to run at older machine without AVX because compiler can emit AVX instructions.

Did you try this and got a crash or you just assume that this scenario won't work?

If you tried and got a crash, please fill a bug report. If not, there is nothing to fix.

Issue #3869, opened 2022-07-16
Regression: 4.1.3 completes run of ScrollView-debugger. Both 5.0.0 and 5.2.0 dump.


rmast:
![175789293-f39ddfdb-6f3e-4598-8d16-80a1f4a88b36](https://user-images.githubusercontent.com/3341558/179373903-ef6cc246-f4e5-4633-a762-ded4dd22708f.jpg)

------------------------

### Environment

* **Tesseract Version**: 5.0.0 or 5.2.0
* **Platform**: Linux Ubuntu20b 5.15.0-1014-azure # 17~20.04.1-Ubuntu SMP Thu Jun 23 20:01:51 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

### Current Behavior:
```
tesseract -c invert_threshold=0.9 --dpi 300 -l Latin -c textord_debug_tabfind=1 -c textord_debug_bugs=1 -c tessedit_reject_block_percent=1 -c tessedit_reject_row_percent=1 -c debug_noise_removal=1 175789293-f39ddfdb-6f3e-4598-8d16-80a1f4a88b36.jpg output81
Vertical skew vector=(7,13180)
Starting sh -c "trap 'kill %1' 0 1 2 ; java -Xms1024m -Xmx2048m -jar /home/rmast/tesseract/java/ScrollView.jar & wait"
ScrollView: Waiting for server...
Socket started on port 8461
Client connected
Click at (2215, 3169)
Click at (2215, 3169)
Click at (2245, 3129)
Click at (2245, 3129)
double free or corruption (fasttop)
sh: 1: kill: No such process

sh: 1: kill: No such process

Aborted (core dumped)

```
The process kills after closing the viewer.
![image](https://user-images.githubusercontent.com/3341558/179374095-2286531f-931a-4883-9f3f-2eaa3b0a05e2.png)

### Expected Behavior:
A complete log of the specified options after closing the viewer, just as in 4.1.3.
### Suggested Fix:
https://github.com/tesseract-ocr/tesseract/pull/3870

rmast:
```
[...]

PR #3870, opened 2022-07-17
Don't use std::unique_ptr for SVEvent

rmast:
https://github.com/tesseract-ocr/tesseract/issues/3869

### DIFF
--- a/src/viewer/scrollview.cpp
+++ b/src/viewer/scrollview.cpp
@@ -98,7 +98,7 @@ void ScrollView::MessageReceiver() {
   // the events accordingly.
   while (true) {
     // The new event we create.
-    std::unique_ptr<SVEvent> cur(new SVEvent);
+    SVEvent *cur(new SVEvent);
     // The ID of the corresponding window.
     int window_id;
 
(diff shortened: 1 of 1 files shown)


stweil:
This reverts commit 37b33749d, but misses some destructor calls which were previously used.

src/viewer/scrollview.cpp
```diff
@@ -836,7 +836,7 @@ char ScrollView::Wait() {
   char ret = '\0';
   SVEventType ev_type = SVET_ANY;
   do {
-    std::unique_ptr<SVEvent> ev(AwaitEvent(SVET_ANY));
+    SVEvent *ev(AwaitEvent(SVET_ANY));
```
stweil, line 839: 4.0.0 already used `std::unique_ptr` for this functionality.

stweil (changes requested):

rmast:
I didn’t test against 4.0.0.

4.1.3:

***@***.***:~/tesseract4.1.3$ grep unique_ptr src/viewer/scrollview.cpp
***@***.***:~/tesseract4.1.3$

rmast:
Conclusion: Also /src/cutil/callcpp.cpp contains a unique_ptr-construction, but that’s probably not getting ‘delete’d. The big log is not falling over it.

I think a clear choice has to be made between using those std::-constructs and make the Java-application use them as well, or just remove them.

stweil:
For 4.x that code was in callcpp.cpp (see link in my previous comment).

rmast:
> For 4.x that code was in callcpp.cpp (see link in my previous comment).

Yes. I saw that on second look. My commit primarily reverses your second commit, but hardly the first one. I wonder whether those other variables are also transferred to and from Java, and whether the std:: -wrapping gives trouble with the Java-application.

PR #3873, opened 2022-07-18
autotools: Add rule for svpaint executable

stweil:
Move also its source code svpaint.cpp from src/viewer/ to src/,
so it is no longer included in libtesseract by the cmake build.


sw.cpp
```diff
@@ -21,7 +21,7 @@ void build(Solution &s)
 
         libtesseract -=
             "src/tesseract.cpp",
```
amitdo, line 23: What about this line?

stweil: Maybe the lines 22...24 are no longer needed at all and can be removed.

Issue #3402, opened 2021-04-21
Accuracy and speed tests


amitdo:
Related to discussion in: https://github.com/tesseract-ocr/tesseract/issues/707#issuecomment-823491996.

@stweil,

Can you contribute these images and their GT and release them under Apache 2.0 license?

Then, can you make a new daily CI test (for Ubuntu at least) that will include these images?

UNLV images could also be used as part of an accuracy test with CI.

stweil:
All images and the ground truth are already available online under the free license [Public Domain Mark 1.0](http://creativecommons.org/publicdomain/mark/1.0/deed.de).

The script for the evaluation is also available online (`do5.sh` [here](https://digi.bib.uni-mannheim.de/~stweil/anciendroit/new/), currently used local filenames), and so is the [list of GT files](https://digi.bib.uni-mannheim.de/~stweil/anciendroit/new/LIST).

The URLs for a GT filename follow this pattern:

41989179X_0036.txt is at https://digi.bib.uni-mannheim.de/fileadmin/digi/41989179X/gt/41989179X_0036.txt, the image is at https://digi.bib.uni-mannheim.de/fileadmin/digi/41989179X/max/41989179X_0036.jpg.

amitdo:
According to this report, there was a huge regression in accuracy and speed in version 3.03 compared to 3.02.02.

We need to evaluate Tesseract weekly using this dataset, in order to make sure there are no regression in performance. The legacy engine and the CRNN engine should be tested separately. This is very importent TODO.

PR #3893, opened 2022-08-06
Create pain points before running associator to resolve #3892

### LINKED ISSUE #3892: Chopper interferes with associator in Legacy engine
### Environment

* **Tesseract Version**: 5.2.0
* **Commit Number**: 15200c6fe7733f71a6cf52fbc1e4d94150f9f168 
* **Platform**: Linux ubuntu 5.15.0-43-generic

### Current Behavior:
The legacy engine often fails to recognize words where characters towards the end of the word are comprised of multiple blobs.  For example, the images below (`and`, `high`, and `which`) are misidentified as `ancl`, `higll`, and `Whicll`. 

mb_words_1_8_0_ancl_and
mb_words_1_9_0_higli_high
mb_words_3_1_0_whicli_which

This issue is caused by the chopper.  When disabling the chopper (`chop_enable=0`) all 3 words are identified correctly (capitalization notwithstanding).  Notably, this is *not* a direct effect of the chopper splitting the last letter (in most cases the chopper does not change the `best_choice`), but rather that running the chopper prevents the associator from working correctly (which should join the last couple blobs into a single letter). 

The issue appears to be that the chopper creates additional blobs, and as this is not properly accounted for, no pain points exist for the last several blobs by the time the associator runs.  In other words, when `chop_enable=1` Tesseract does not even try to combine the last 2 blobs in the examples above. 

### Suggested Fix:
I'm creating a pull request that resolves by making sure `InitialSegSearch` (which creates pain points between blobs) runs directly before the associator step, regardless of whether the chopper is run.  When this change is made the words above are identified correctly whether `clop_enable=0` or `clop_enable=1`.

Balearica:
See #3892

### DIFF
--- a/src/wordrec/segsearch.cpp
+++ b/src/wordrec/segsearch.cpp
@@ -52,6 +52,8 @@ void Wordrec::SegSearch(WERD_RES *word_res, BestChoiceBundle *best_choice_bundle
     if (blamer_bundle != nullptr && !blamer_bundle->ChoiceIsCorrect(word_res->best_choice)) {
       blamer_bundle->SetChopperBlame(word_res, wordrec_debug_blamer);
     }
+
+    InitialSegSearch(word_res, &pain_points, &pending, best_choice_bundle, blamer_bundle);
   }
   // Keep trying to find a better path by fixing the "pain points".
 
--- a/src/wordrec/segsearch.cpp
+++ b/src/wordrec/segsearch.cpp
@@ -44,6 +44,7 @@ void Wordrec::SegSearch(WERD_RES *word_res, BestChoiceBundle *best_choice_bundle
     if (chop_enable && word_res->chopped_word != nullptr) {
       improve_by_chopping(rating_cert_scale, word_res, best_choice_bundle, blamer_bundle,
                           &pain_points, &pending);
+      InitialSegSearch(word_res, &pain_points, &pending, best_choice_bundle, blamer_bundle);
     }
     if (chop_debug) {
       SEAM::PrintSeams("Final seam list:", word_res->seam_array);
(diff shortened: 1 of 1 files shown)


amitdo:
The problem is that we don't know if there might be some cases where this patch will cause worst results.

@stweil, @zdenop, maybe we can accept this addition if it will be optional, using a config variable and will be turn off by default?

Balearica:
@amitdo Do you have a specific scenario in mind where this change would plausibly cause worse results?  Alternatively, is there some corpus of benchmark documents where we could assess the impact of this change empirically? 

The behavior this PR addresses (fully documented in #3892) is clearly a bug, as there is no reason why the associator should randomly skip letters at the end of certain words.  I don't think it makes sense to avoid changing this behavior (using the default settings) in the absence of specific concerns regarding this fix.  As stated above, I am happy to run an accuracy benchmark if one already exists.

amitdo:
>Do you have a specific scenario in mind where this change would plausibly cause worse results? 

No.

Don't assume that the few currently active developers deeply know all the algorithms used in Tesseract.

>Alternatively, is there some corpus of benchmark documents where we could assess the impact of this change empirically?

In the past the UNLV dataset and tools were used for testing Tesseract's accuracy.

See:
* [The test repo](https://github.com/tesseract-ocr/test/blob/main/unlvtests/README.md)
* [UNLV Testing of Tesseract](https://github.com/tesseract-ocr/tessdoc/blob/eb0a25d0ab4676fd1cb28c43ab36d47eb906f892/UNLV-Testing-of-Tesseract.md)
* https://github.com/eddieantonio/ocreval

But the UNLV dataset have just English and Spanish written texts. Do you think your patch is fine for all the scripts that Tesseract supports?

Related issue: #3402.

amitdo:
@stweil,

I think we should trust @Balearica and accept this PR.

stweil:
Yes. @Balearica, do you want to rebase the commits in this pull request and fix the author which is currently called "Your Name"?

PR #3894, opened 2022-08-06
Add framework for implementation of support for AVX512VNNI

stweil:
Using AVX512VNNI instructions is still missing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -170,6 +170,14 @@ libtesseract_la_LIBADD += libtesseract_avx512.la
 noinst_LTLIBRARIES += libtesseract_avx512.la
 endif
 
+if HAVE_AVX512VNNI
+libtesseract_avx512vnni_la_CXXFLAGS = -march=icelake-client
+libtesseract_avx512vnni_la_CXXFLAGS += -I$(top_srcdir)/src/ccutil
+libtesseract_avx512vnni_la_SOURCES = src/arch/intsimdmatrixavx512vnni.cpp
+libtesseract_la_LIBADD += libtesseract_avx512vnni.la
+noinst_LTLIBRARIES += libtesseract_avx512vnni.la
+endif
+
 if HAVE_FMA
 libtesseract_fma_la_CXXFLAGS = -mfma
 libtesseract_fma_la_CXXFLAGS += -I$(top_srcdir)/src/ccutil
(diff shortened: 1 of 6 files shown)


Makefile.am
```diff
@@ -170,6 +170,14 @@ libtesseract_la_LIBADD += libtesseract_avx512.la
 noinst_LTLIBRARIES += libtesseract_avx512.la
 endif
 
+if HAVE_AVX512VNNI
+libtesseract_avx512vnni_la_CXXFLAGS = -mavx512vnni -mavx512vl
```
stweil, line 174: I am curious: what is the use of `-mavx512vl`? Should it be checked in `configure`, too?

amitdo:
Ready for testing.

I suggest to backup the home directory and any other important stuff, in case this patch will trigger a hard crash (reboot).

stweil:
> I suggest to backup the home directory and any other important stuff, in case this patch will trigger a hard crash (reboot).

Will it? I cannot test it before August 21 because I currently don't have access to the required hardware.

Makefile.am
amitdo: It does not compile without the second flag. It is needed becaIuse the code uses ymm registers (256 bit) instead of zmm registers (512 bit).

Yes, we can also add another check in configure.

Note that AFAIK all intel's CPUs that support AVX512VNNI also support AVXVL.

stweil: Thanks for explaining. Maybe we don't need the additional check then.

amitdo:
>Will it?

I don't think so...

amitdo:
Stefan, will you able to test the code today or in in the upcoming days?

stweil:
> Stefan, will you able to test the code today or in in the upcoming days?

I just finished a first run of `intsimdmatrix_test` with the new code. It works, but I was surprised because the performance is identical to the AVX2 implementation.

stweil:
Support for sw build is still missing. Therefore CI currently fails.

amitdo:
>I was surprised because the performance is identical to the AVX2 implementation.

This is disappointing.

We can also remove the 'ones' vector. The vnni version does not need it.

amitdo:
Did you fetched and merged my changes locally?

```
git switch avx512-vnni
git log -1
```

Do you see my commit?

stweil:
> We can also remove the 'ones' vector. The vnni version does not need it.

I removed it locally. But that did not change the generated code. Obviously the compiler also removed it.

amitdo:
@stweil,

Do you have access to anther machine (non-notebook) that supports AVX512VNNI ?

stweil:
No, I have only sometimes access to a notebook which supports it.

PR #3920, opened 2022-09-16
Update LICENSE

markysha:


LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2022
```
stweil, line 190: There is no need to update this text: it is in the appendix "How to apply the Apache License to your work".

stweil (commented):
This change is not needed.

PR #3921, opened 2022-09-20
Fix build with unreleased leptonica 1.83

p12tic:
leptonica moved most of their type definitions to internal headers. Long term usages of internal API should be removed, but for now simply include the internal header to allow tesseract to be built.

### DIFF
--- a/src/textord/devanagari_processing.cpp
+++ b/src/textord/devanagari_processing.cpp
@@ -28,6 +28,9 @@
 #include "tordmain.h"
 
 #include <allheaders.h>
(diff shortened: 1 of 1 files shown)


p12tic:
https://github.com/tesseract-ocr/tesseract/pull/3851 should probably merged instead of this PR (I reviewed that one), but if it looks too risky, then this PR is the simpliest possible fix.

zdenop:
close in favour of 3851

amitdo:
>#3851 should probably merged instead of this PR (I reviewed that one), but if it looks too risky, then this PR is the simpliest possible fix.

This comment is misleading. #3851 will not solve the issue that this PR is trying to solve. They are dealing with the same issue but in different files.
 
However, #3851 could be extended to solve the issue in additional file.

PR #3923, opened 2022-09-21
add link to PPOCRLabel

user3984:
Thank you for your repository. Could you please add a link to our annotation tool PPOCRLabel?


zdenop:
Readme is dedicated to tesseract and not 3rd party tools.

PR #3930, opened 2022-09-23
Fix tesseract.pc from cmake to match autotools

jeroen:
cf: https://github.com/tesseract-ocr/tesseract/blob/HEAD/tesseract.pc.in#L13

We found that bindings fail to compile when tesseract was built with cmake, because its pc file does not properly declare the leptonica dependency in the headers.


stweil (approved):

stweil:
Thanks!

PR #3935, opened 2022-10-02
Update README.md

nicholasz2510:
The `Installation` page has more links to binaries than the `Home` page does, and provides better instructions for installing from pre-built binaries.


stweil (approved):
Yes, thank you.

Issue #3940, opened 2022-10-07
Regression: gif not scraping in 5.2 that was OK in 5.1


jacekkopecky:
### Environment

* **Tesseract Version**: v5.2.0.20220712
* **Platform**: Windows 10, x64

### Current Behavior:

`tesseract.exe` v 5.2 only scrapes top line (window title) from this `gif` image:

<img width="300" alt="image-screenshot-pdf.gif" src="https://user-images.githubusercontent.com/807315/194534960-3c8ab221-910f-42af-a8d9-eb34d8b458f9.gif">

### Expected Behavior:

It should scrape all text.

Tesseract  v5.1.0.20220510 worked as expected, and both 5.1 and 5.2 work as expected with this equivalent `png` image:

<img width="300" alt="image-screenshot-pdf.png" src="https://user-images.githubusercontent.com/807315/194535081-097e4044-47bc-4a70-8f45-7788d69470ec.png">

### Suggested Fix:

It might have something to do with a different DPI estimation: 5.2 estimates resolution 132, while 5.1 estimated 168. However, running 5.2 with `--dpi 168` does not seem to fix anything.

amitdo:
What's the output of tesseract -v for each version?

I want to see if the leptonica and gif library are used are exactly the same.

jacekkopecky:
```
tesseract v5.1.0.20220510
 leptonica-1.78.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.3) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
 Found AVX512BW
 Found AVX512F
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
 Found libarchive 3.5.0 zlib/1.2.11 liblzma/5.2.3 bz2lib/1.0.6 liblz4/1.7.5 libzstd/1.4.5
 Found libcurl/7.77.0-DEV Schannel zlib/1.2.11 zstd/1.4.5 libidn2/2.0.4 nghttp2/1.31.0
```

```
tesseract v5.2.0.20220712
 leptonica-1.78.0
  libgif 5.1.4 : libjpeg 8d (libjpeg-turbo 1.5.3) : libpng 1.6.34 : libtiff 4.0.9 : zlib 1.2.11 : libwebp 0.6.1 : libopenjp2 2.3.0
 Found AVX512BW
 Found AVX512F
[...]

Issue #3092, opened 2020-09-07
PNG images being reencoded unnecessarily in PDFs


ericrechlin:
* **Tesseract Version**: 4.1.1 (also verified with latest 5.0 code)
* **Commit Number**: 5fe1390748a15c0e445a5c57c834edff27ff2f4d and later
* **Platform**: Linux Desktop 4.4.0-19041-Microsoft #1-Microsoft Fri Dec 06 14:06:00 PST 2019 x86_64 x86_64 x86_64 GNU/Linux (Ubuntu 18.04 under WSL)

### Current Behavior:

After upgrading my version of tesseract, I noticed that the PDF files it generates from PNGs are often larger, sometimes much larger, than they used to be with past versions.

For example, consider the following 262 KB PNG file:

![sample](https://user-images.githubusercontent.com/29085101/92346755-6430cd00-f093-11ea-93e8-3ecb7bf76af0.png)

Run it with the following command line to generate an OCR'd PDF:

tesseract sample.png sample pdf

With versions of tesseract-ocr prior to 4.0-rc1, this generates a 269 KB PDF. However, with versions beginning with 4.0-rc1, this results in a 402 KB PDF, because the image is being needlessly reencoded, rather than being placed into the PDF unchanged.

I have narrowed this bug down to the fix applied for issue #1914, applied with commit 5fe1390748a15c0e445a5c57c834edff27ff2f4d.  Specifically, the changes made to src/api/pdfrenderer.cpp in this commit are what are causing the above problem.

### Expected Behavior:

The generated PDF should be similar in size to the original PNG instead of much larger (269 KB in this case instead of 402 KB).

### Suggested Fix:

Changes need to be made to the imageToPDFObj method in src/api/pdfrenderer.cpp, either reverting the changes made in #1914 or modifying them to avoid reencoding the image.

zdenop:
Actually problem in src/ccmain/thresholder.cpp:
    [Image tmp = pixRemoveColormap(src, REMOVE_CMAP_BASED_ON_SRC);](https://github.com/tesseract-ocr/tesseract/blob/395f2315bc9ddd4434e2ad30a178ea3f28226d55/src/ccmain/thresholder.cpp#L174)

Reported image depth (`Image src`) is 8, but after removing colormap is increased to 32 so image has bigger size...
@DanBloomberg: is it correct behavior?

zdenop:
[...]

PR #3942, opened 2022-10-14
fix issue #3940 - remove colormap before thresholding

zdenop:
Maybe code for colormap removal could be simplified to `original = pixConvertTo8(original.copy(), false);` but I decided to stick to original code (in SetImage).

This change works for me for images in issues #3940 and #3092 and for thresholding_method 0,1,2.

### DIFF
--- a/src/ccmain/thresholder.cpp
+++ b/src/ccmain/thresholder.cpp
@@ -175,6 +175,7 @@ void ImageThresholder::SetImage(const Image pix) {
   } else {
     pix_ = src.copy();
   }
+  src.destroy();
   depth = pixGetDepth(pix_);
   pix_channels_ = depth / 8;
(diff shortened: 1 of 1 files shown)


egorpugin (approved):

amitdo:
Unfortunately, this patch causes even worst regression. Many of the CI jobs fails.

amitdo:
commit 14e2517d6d7d64 fixed the issue detected by ASan. There are still a memory leak issue and failed tests.

stweil:
The failed unit tests are hopefully fixed now.

stweil:
Note that restarting the GitHub action uses the same old code as in the previous run, so restarting only helps for temporary failures (network timeouts, ...). So we have to wait for the next regular action trigger.

Issue #3943, opened 2022-10-15
PrintVariables produces config file that ReadConfigFile does not properly read


Balearica:
### Expected Behavior:
The API function `PrintVariables` prints current parameters to a file, and `ReadConfigFile` reads parameters from a file.  Intuitively,  `ReadConfigFile` should be able to read the files that `PrintVariables` writes.  This is explicitly assumed within the `ProcessPage` function, where these functions are used together to "Save current config variables before switching modes" and then "Restore saved config variables".

https://github.com/tesseract-ocr/tesseract/blob/a873553593988407b7a9043c30c3010e0d066205/src/api/baseapi.cpp#L1293-L1306

### Current Behavior:
Unfortunately, this does not currently work properly.  The issue is that `PrintVariables` prints parameter descriptions alongside key/value pairs (e.g. `chs_trailing_punct1	).,;:?!	1st Trailing punctuation`), and `ReadConfigFile` reads the description as a value (for string parameters).  An example showing this is below. 

```
#include <tesseract/baseapi.h>

int main()
{
    tesseract::TessBaseAPI *api = new tesseract::TessBaseAPI();
    if (api->Init(NULL, "eng")) {
        fprintf(stderr, "Could not initialize tesseract.\n");
        exit(1);
    }

    static const char *kOldVarsFile = "failed_vars.txt";

    // Print default value of chs_trailing_punct1
    printf("Initial value: %s\n", api->GetStringVariable("chs_trailing_punct1"));
    FILE *fp = fopen(kOldVarsFile, "wb");
    api->PrintVariables(fp);
    fclose(fp);
    api->ReadConfigFile(kOldVarsFile);
    printf("After PrintVariables/ReadConfigFile: %s\n", api->GetStringVariable("chs_trailing_punct1"));

    api->End();
    delete api;
    return 0;
}
```

This returns the following:

```
Initial value: ).,;:?!
After PrintVariables/ReadConfigFile: ).,;:?!	1st Trailing punctuation
```

The impact of this is:
1. `ProcessPage` does not work correctly when used with `retry_config`
2. There is no simple interface for generating a config file with the user's current settings
   1. This is useful for saving/restoring configurations (as `ProcessPage` attempts to do)

### Suggested Fix:
[...]

PR #3947, opened 2022-10-19
Updated PrintParams to not print info text to config files per #3943

Balearica:
See #3943


stweil (changes requested):
~That's unexpected behaviour for users. Therefore I'd not merge this pull request.~ I'd prefer a different solution, for example an extra command `dump_parameters` or `save_parameters`, or extending `ReadConfigFile` to accept comments / descriptions (see discussion in #3943).

PR #3959, opened 2022-11-07
pdfrenderer.cpp: Ignore non-text blocks

### LINKED ISSUE #3957: PDF renderer: Tesseract inserts spaces for non-text blocks it finds
### Environment

* **Tesseract Version**: 5.2.0 and 4.1.3
* **Platform**:  Windows 10, x64

### Current Behavior:

Spaces are found by Tesseract and inserted in output PDF. It seems to be confused by the horizontal lines in the attached image.
I use this PDF to overlay on top of another PDF which already contains text, and this causes problems in the resulting PDF due to intersections.

### Expected Behavior:

Tesseract finding standalone spaces does not make sense to me. I would expect Tesseract to only find characters in the logo and disregard the lines. At least the lines should be underscores or something - not spaces.

### Suggested Fix:

Unsure if problem should be fixed in Tesseract or if the spaces should be filtered in the PDF renderer. I think Tesseract as it will fix all output formats.

spaces

amitdo:
Fix #3957.

### DIFF
--- a/src/api/pdfrenderer.cpp
+++ b/src/api/pdfrenderer.cpp
@@ -25,6 +25,7 @@
 
 #include <allheaders.h>
 #include <tesseract/baseapi.h>
+#include <tesseract/publictypes.h> // for PTIsTextType()
 #include <tesseract/renderer.h>
 #include <cmath>
 #include <cstring>
@@ -354,6 +355,12 @@ char *TessPDFRenderer::GetPDFTextObjects(TessBaseAPI *api, double width, double 
   const std::unique_ptr</*non-const*/ ResultIterator> res_it(api->GetIterator());
   while (!res_it->Empty(RIL_BLOCK)) {
     if (res_it->IsAtBeginningOf(RIL_BLOCK)) {
+      auto block_type = res_it->BlockType();
+      if (!PTIsTextType(block_type)) {
+        // ignore non-text blocks
+        res_it->Next(RIL_BLOCK);
+        continue;
+      }
       pdf_str << "BT\n3 Tr"; // Begin text object, use invisible ink
       old_fontsize = 0;      // Every block will declare its fontsize
       new_block = true;      // Every block will declare its affine matrix


amitdo:
@egorpugin,

Can you fix sw?

src/api/pdfrenderer.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <allheaders.h>
 #include <tesseract/baseapi.h>
+#include <tesseract/publictypesp.h> // for PTIsTextType()
```
stweil, line 28:
```suggestion
#include <tesseract/publictypes.h> // for PTIsTextType()
```

amitdo: Yes, I also noticed this typo.

amitdo:
I'll reopen this PR after I fix it.

egorpugin:
I'll check sw issues.

amitdo:
I tested it and it hangs forever. Seems the loop never ends. 

I'll see if I can fix it.

amitdo:
Tested using the image from #3957. It works as expected and fixes the issue.

amitdo:
Here is the pdf file that Tesseract produces after applying the patch from this PR.

[3957.pdf](https://github.com/tesseract-ocr/tesseract/files/9958369/3957.pdf)

amitdo:
macOS CI failed almost immediately. The failure is not related to this PR.

amitdo:
Here is the pdf file that Tesseract produces *before* applying the patch from this PR.

[3957-0.pdf](https://github.com/tesseract-ocr/tesseract/files/9967201/3957-0.pdf)

amitdo:
Using 'select all'. 'copy' and then paste to text file, and using pdf2txt there is no spaces, for both 'before' and 'after' PDFs.

Still, there is a difference internally between the files. The 'after' pdf is slightly smaller (in bytes) than the 'before' pdf file.

amitdo:
Waiting for a feedback from @bleze.

amitdo:
https://www.pdf-online.com/osa/validate.aspx




File	3957-0.pdf
Compliance	pdf1.5
Result	Document validated successfully.
Details	

Validating file "3957-0.pdf" for conformance level pdf1.5

The document does conform to the PDF 1.5 standard.



File	3957.pdf
Compliance	pdf1.5
Result	Document validated successfully.
Details	

Validating file "3957.pdf" for conformance level pdf1.5

The document does conform to the PDF 1.5 standard.

amitdo:
@stweil, can I merge this PR?

It removes unneeded stuff from the pdf output in documents with non-text blocks and make the document slightly smaller in byes.

stweil (approved):
Looks good, thank you!

bleze:
> Waiting for a feedback from @bleze.

I already took the code from the patch and applied to my copy. I can confirm that the spaces are no longer included in the output. Thank you for fixing this!

PR #3962, opened 2022-11-12
lstm.train: allow .box from .raw.png too

bertsky:
(typical for ocropy / kraken and GT without binarization)


stweil (approved):
The patch is fine for now, but a more general solution would be good.

PR #3967, opened 2022-11-19
Fix a number of performance issues (reported by Coverity Scan)

stweil:
The latest scan reported a number of performance issues because
several `for` loops used expensive copy operations.

After adding the `const` attribute to compare operators, `const`
could also be used in most `for` loops.

### DIFF
--- a/src/api/hocrrenderer.cpp
+++ b/src/api/hocrrenderer.cpp
@@ -412,7 +412,7 @@ char *TessBaseAPI::GetHOCRText(ETEXT_DESC *monitor, int page_number) {
                    << " id='"
                    << "timestep" << page_id << "_" << wcnt << "_" << tcnt
                    << "'>";
-          for (auto conf : timestep) {
+          for (auto &&conf : timestep) {
             hocr_str << "\n         <span class='ocrx_cinfo'"
                      << " id='"
                      << "choice_" << page_id << "_" << wcnt << "_" << ccnt
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -1191,7 +1191,7 @@ bool Tesseract::SelectGoodDiacriticOutlines(int pass, float certainty_threshold,
     *ok_outlines = best_outlines;
     if (debug_noise_removal) {
       tprintf("%s noise combination ", blob ? "Adding" : "New");
-      for (auto best_outline : best_outlines) {
+      for (auto &&best_outline : best_outlines) {
         tprintf("%c", best_outline ? 'T' : 'F');
       }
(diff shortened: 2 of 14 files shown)


egorpugin:
The fix is not very right.
`for (auto &&` should be used instead.

stweil:
I did not know `auto &&` up to now, and it is currently used only once in Tesseract code. Why would it be better than `const auto &`?

egorpugin:
`auto &&` means `auto` when we have a simple value.
`auto &&` means `auto &` when we have a reference.
`auto &&` means `const auto &` when we have a const reference.
`auto &&` means `auto &&` when we have `&&` reference.

In other words `auto &&` selects right thing for you.
If you set `auto &` and then for some reason you change iteration range to const, you need manually fix all places. With `auto &&` you are already set.

stweil:
> In other words `auto &&` selects right thing for you.

That's indeed a nice feature, but an explicit `const` attribute has also advantages because it helps human readers and detects missing `const` attributes like the ones in the compare operators. Therefore I'd prefer to use `const` where it is possible.

So how should we fix the reported performance issues? Use `auto &&` instead of `auto &` and keep the `const auto &` cases? Or use `auto &&` and `const auto &&`?

egorpugin:
> but an explicit const attribute has also advantages because it helps human readers and detects missing const attributes like the ones in the compare operators

When you will be updating code `const->non const` you'll have to fix all places again.

> So how should we fix the reported performance issues?

`auto &&`

> Use auto && instead of auto &

Almost always `auto &&`. There are some rare cases when you have pointer and want explicitly use `auto` or `auto *`.

> and keep the const auto & cases?

While refreshing codebase all such cases must be changed to `auto &&`.

> const auto &&?

Did not see `const auto &&` usages anywhere. Do not sure exactly what is it and how it works. Do not use it.
It can mean `const auto &`.
But also it may mean copy as in raw `auto` which is bad.

---

So, we should use `auto &&` everywhere (almost).

stweil:
I updated the PR now to use `auto &&` (although I still think that `const auto &` has advantages, too).

egorpugin (approved):

zdenop:
thanks

amitdo:
https://stackoverflow.com/questions/26541920/is-it-a-good-practice-to-use-const-auto-in-a-range-for-to-process-the-element

https://stackoverflow.com/questions/29320575/whats-the-difference-between-and-in-a-range-based-for-loop

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3994.htm

https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n3853.htm

Issue #3925, opened 2022-09-22
Many bugs in training the legacy engine


SpaceView:
I doubt anybody have successfully trained custom data with tesseract 5.2.0 and 5.1.0, **the latest I can succeed is 5.0.0-alpha-20201224**.
Below are some BUGs when I'm running tesseract 5.2.0 for custom data training.  I can say there are TOO MANY BUGS, thus I was not able to finish the whole  training due to limited time at this moment, below are just a few of the found BUGs for reference.
 
```
// Deletes all samples with zero features marked by KillSample.
void TrainingSampleSet::DeleteDeadSamples() {
  using namespace std::placeholders; // for _1
  auto old_it = samples_.begin();
  for (; old_it < samples_.end(); ++old_it) {
    if (*old_it == nullptr || (*old_it)->class_id() < 0) {
      break;
    }
  }
  auto new_it = old_it;
  for (; old_it < samples_.end(); ++old_it) {
    if (*old_it == nullptr || (*old_it)->class_id() < 0) {
      delete *old_it;
    } else {
      *new_it = *old_it;
      ++new_it;
    }
  }
  //samples_.resize(new_it - samples_.begin() + 1);      //<------------crash the program when samples_.size() is 0
  samples_.resize(new_it - samples_.begin());
  num_raw_samples_ = samples_.size();
  // Samples must be re-organized now we have deleted a few.
}



INT_TEMPLATES_STRUCT *Classify::CreateIntTemplates(CLASSES FloatProtos,
                                           const UNICHARSET &target_unicharset) {
  CLASS_TYPE FClass;
  INT_CLASS_STRUCT *IClass;
  int ProtoId;
  int ConfigId;

  auto IntTemplates = new INT_TEMPLATES_STRUCT;

  for (unsigned ClassId = 0; ClassId < target_unicharset.size(); ClassId++) {
    FClass = &(FloatProtos[ClassId]);
    if (FClass->NumProtos == 0 && FClass->NumConfigs == 0 &&
        strcmp(target_unicharset.id_to_unichar(ClassId), " ") != 0) {
      tprintf("Warning: no protos/configs for %s in CreateIntTemplates()\n",
              target_unicharset.id_to_unichar(ClassId));
    }
    assert(UnusedClassIdIn(IntTemplates, ClassId));
    IClass = new INT_CLASS_STRUCT(FClass->NumProtos, FClass->NumConfigs);
[...]

PR #3972, opened 2022-12-03
Fix function tesseract::WriteFeature (issue #3925)

stweil:
Fixes: 3b0759940c16b68e ("Replace more STRING by std::string")


egorpugin:
Dont we need to fix double values parser instead of emitter?

stweil:
Yes, ideally we should do that, too. See [my comment for the issue](https://github.com/tesseract-ocr/tesseract/issues/3925#issuecomment-1336271140).

PR #4015, opened 2023-02-08
Replace 'can not' by 'cannot'

stweil:
Both forms are used in American English, but 'cannot' is more common (also in Tesseract code), so use it always.


stweil:
@egorpugin, can CI for MacOS be fixed (see [details](https://github.com/tesseract-ocr/tesseract/actions/runs/4126311717/jobs/7128070991))?

Issue #3998, opened 2023-01-22
RFC: Info messages to stdout from training utils


amitdo:
https://groups.google.com/g/tesseract-dev/c/OoBUOPZtkrQ

@zdenop, I think your message will be more visible here as an RFC. 

>Zdenko Podobny
Jan 20
>
>I realized that several users did not recognize errors during the training process.
IMO part of the problem is that all messages (error and standard output) from training tools are shown in stderr because of tprinf usage. 
>
>While this make sense in tesseract executable (OCR process output is sent to stdout, all other messages to stderr), in training we should use different approach: only errors (e.g. that should stop further process) should go to stderr and all other info should go to stdout.
>
>Good example is unicharset_extractor:
https://github.com/tesseract-ocr/tesseract/blob/4142b328157fa5acdb5780308ecf308f1c6e2ec7/src/training/unicharset_extractor.cpp#L75-L84
Are you ok with this proposal? This would mean that tprinf will be used for errors, and std::cout/fprintf(stdout for rest...
>
>
>Zdenko

amitdo:
I prefer to keep using tprint in training and use a  prefix (ERROR/WARNING) as suggested by @GerHobbelt.

PR #4020, opened 2023-02-25
follow-up to dev mailing list discuss + issue #3998: RFC:  Consistent tprintf() use, ERROR: & WARNING: prefixes.

GerHobbelt:
follow-up to dev mailing list discuss + issue #3998: RFC:  Info messages to stdout from training utils (Issue #3998)

Here's the extract/backport from my master branch:

- use tprintf() for all error/warning/info/debug-diagnostics output --> fprintf(stderr, msg) --> tprintf(msg). This also anticipates migrating tesseract to using the fmt library (C++20) as done by @stweil in a dev branch.
- all error messages are prefixed with 'ERROR: '. This enables users, f.e., to grep/filter the stderr/log using a regex like 
   `/^ERROR: /` 
   to filter out the error messages. 
   
   Application code which overrides the tprintf() API is another example: there error messages can be easily picked up and sent to different monitor channels / GUI windows / etc. in environments where there's no stdout/stderr (MSWindows GUI) or where end users are not accustomed to watching stdout/stderr.
- ditto for 'WARNING: ' level messages.

CAVEAT: this is a manual transfer from my master branch + code review while I was at it. I hope I got them all... :-)

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -585,7 +585,7 @@ void TessBaseAPI::SetSourceResolution(int ppi) {
   if (thresholder_) {
     thresholder_->SetSourceYResolution(ppi);
   } else {
-    tprintf("Please call SetImage before SetSourceResolution.\n");
+    tprintf("ERROR: Please call SetImage before SetSourceResolution.\n");
   }
 }
 
@@ -1005,7 +1005,7 @@ bool TessBaseAPI::ProcessPagesFileList(FILE *flist, std::string *buf, const char
     chomp_string(pagename);
     Pix *pix = pixRead(pagename);
     if (pix == nullptr) {
-      tprintf("Image file %s cannot be read!\n", pagename);
+      tprintf("ERROR: Image file %s cannot be read!\n", pagename);
       return false;
     }
     tprintf("Page %u : %s\n", page, pagename);
@@ -1074,7 +1074,7 @@ bool TessBaseAPI::ProcessPages(const char *filename, const char *retry_config, i
 #ifndef DISABLED_LEGACY_ENGINE
   if (result) {
     if (tesseract_->tessedit_train_from_boxes && !tesseract_->WriteTRFile(output_file_.c_str())) {
-      tprintf("Write of TR file failed: %s\n", output_file_.c_str());
+      tprintf("ERROR: Write of TR file failed: %s\n", output_file_.c_str());
       return false;
     }
   }
@@ -1106,9 +1106,9 @@ bool TessBaseAPI::ProcessPagesInternal(const char *filename, const char *retry_c
                                        int timeout_millisec, TessResultRenderer *renderer) {
   bool stdInput = !strcmp(filename, "stdin") || !strcmp(filename, "-");
   if (stdInput) {
-#ifdef WIN32
+#if defined(WIN32) || defined(_WIN32) || defined(_WIN64)
     if (_setmode(_fileno(stdin), _O_BINARY) == -1)
-      tprintf("ERROR: cin to binary: %s", strerror(errno));
+      tprintf("ERROR: Cannot set STDIN to binary: %s", strerror(errno));
 #endif // WIN32
   }
 
@@ -1135,7 +1135,7 @@ bool TessBaseAPI::ProcessPagesInternal(const char *filename, const char *retry_c
     } else {
       CURLcode curlcode;
       auto error = [curl, &curlcode](const char *function) {
-        fprintf(stderr, "Error, %s failed with error %s\n", function, curl_easy_strerror(curlcode));
+        tprintf("ERROR: %s failed with error %s\n", function, curl_easy_strerror(curlcode));
         curl_easy_cleanup(curl);
         return false;
       };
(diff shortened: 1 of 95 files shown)


GerHobbelt:
Notes:

- adds an error message when UZN file loading fails -- rather than being utterly silent about that.
- Edits a few error messages minimally to make them easier to understand -- at least for me. 😉 
- As a rule of thumb: version info goes to stdout, but anything else that's not direct OCR results goes to stderr, via tprintf. Thus one can easily redirect stdout to file and not be bothered with a having to pull apart a mix of data and diagnostics messages.

egorpugin:
This is not right.
Proper way is to add fun/macro `error()` or `warning()` or `tprintf_error()` ... that will wrap text with prefix `ERROR: ` or `MESSAGE: `.

GerHobbelt:
Ok. Do note however, that a *few* of those prefixes aren't really prefixes as they are preceded by a "\n" that's part of the same message string, so a function will not cover 100% of cases. 😉 

Given that, I suggest I do a macro as then the result remains the same at compile time, e.g. `tprintf("\n" TPRNT_ERROR("blah blah"), ....)` --> expands to a single constant string. Also nice because `fmt` depends on the format string to be a const_expr, in case tesseract mainline decides to migrate to using `fmt` as was done by @stweeil in a dev branch of his (and which I blatantly ripped for my own).

BTW: definitely *not* an ERROR() or error() macro as that has a huge chance of clashing with foreign code. ERROR, for example, is a 100% collision with a system-level ERROR define in Windows (been there, seen that, t-shirt, elsewhere)

Therefore I propose `TPRNT_ERROR(msg)` and `TPRNT_WARNING(msg)` macros. Deal? 😉

GerHobbelt:
BTW: there are two important levels for the tprintf() messages:

- Error
- Warning

and the rest is *not* prefixed (ok, a rare 'INFO:' may be observed; I'm only consistent 95% of the time 😄 ) so all info/debug messages come *without* a prefix, like before. 
This keeps the general diagnostics output (which is *loud* and *needed* if you want to find out what tesseract is doing at all, I find) clean from useless prefixes that only clutter console/logfile space.

Meanwhile the important stuff (errors and warnings) can be easily filtered/routed in applications, where those are relevant to report/log without being obscured by the diag output noise.

----------

If you are of a different opinion re the above levels & prefixes 'rule', I'm all ears and I'll see how this patch can be tweaked.

egorpugin:
If current `tprint` means logger, we can name new entities `log_error()`, `log_warning()`, `log_info()` etc.

GerHobbelt:
@egorpugin : before I tackle this, there's also `tlog()` and `ERROR::error()` which invokes `fprintf(stderr, ...)` directly, so the question is: what would you like?

I'm myself of two minds about this: I could do it as a `tprint_error`/`tprint_warning`/`tprint_info` set[^1] (do we want the `tlog()` to be merged/mixed in with that, as we're talking _code refactoring_ anyway), which has **my minor preference**, or we could do it as tprintf-the-whole-darn-lot-with-ERROR+WARNING-macros, or, syslog-like levels, e.g. `tprintf(ERROR, msg...)` 🤔 

If we can get some kind of a management decision on this refactor action, that would be great. 😄  Then I can do it in my own fork and backport it, like the stuff done so far, but I'd rather do the refactor **once** as it's... _some work_. And finding out I did it in a direction that's not liked, well... that'll cost me later for a feature I don't mind otherwise while I keep tracking your mainline repo: increased expected cost of git merge collisions. 😭  Anyway, so's you know why I ask this.  

Thanks! 👍 

-------------

Side note: `ERROR::error()` also has an 'abort'/fatal mode a la glog (google log library, see github). I tackle that sort of thing in my backend by pumping everything through the supposed channels and in case of a fatal, doing a hard flush. That means we'll have something like `log_flush()` and `log_abort()` too: I like to keep the flush and actual abort logic separate -- own backend has debugger detection built-in so the fail-on-null-pointer-dereference is then exchanged for a jump-into-debugger call. By providing log_flush() + log_abort() anyone can override those actions easily, like I did for my own backend. 

OK or do you want it done differently?

--------------

[^1]: which meshes nicely with my back-end where I have different error/warning/info/debug calls, so log_error(), log_warning(), log_info() is you suggested earlier makes that a cheaper tactic at run-time. 👍  Hence the light preference for this approach.

egorpugin:
Just leave it as is with `"ERROR: `.
Further refactorings will improve it.

In the second part you are talking about more like full featured logger rather than some one-function-printer.
If tesseract really needs this with sources & sinks & MT support etc, it can be handled with switching to boost log or spd log or small but consistent own implementation. It is not this ticket topic I guess.

GerHobbelt:
Ok, if I understood that correctly, you want to keep this mostly as-is.

I'll do the tprintf() et al -> log_error() et al migration as a separate pullreq later. BTW: the key argument there is to NOT have spdlog/glog/whatever in tesseract itself, but to keep tesseract mostly as-is, BUT provide a somewhat easy way for OTHERS to modify or hook tesseract up to such systems if THEIR use of the tesseract library requires such measures.

In other words: 
- tesseract-the-application should simply keep writing to stderr as before -- only through a small set of functions, such as tprintf.
- tesseract-the-library can do the same, or...
- tesseract-the-library can have its tprintf(),etc. output hooked into and rerouted to whatever the application programmer needs, e.g. glog or completely different means.

Hope I made myself clear on the goal there -- which is ultimately what I need / use.
Being consistent in the format (prefix, rather) of error and warning messages is a first step towards all that.
Any refactoring in that direction can be done in a separate pullreq as I understand you suggest; it would nicely separate the changes intended for a different goal.

Apologies for any confusion I might've caused.

egorpugin:
How do you imagine hooking?

PR #4040, opened 2023-03-24
Remove set but not used variable

zdenop:
reported by clang.

And If I understand the code correctly: this old code (3.01) calculates  gap, but never use it...
https://github.com/tesseract-ocr/tesseract/blob/1569e5080810f4652b720bcd344026a9b236ec50/src/ccmain/control.cpp#L1687-L1698

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -1414,7 +1414,6 @@ char *TessBaseAPI::GetTSVText(int page_number) {
     return nullptr;
   }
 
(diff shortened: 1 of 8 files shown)


stweil:
I wonder whether it would be better to use some variables in (configurable) debug output. Some of them can help to get a better understanding of the code.

zdenop:
No problem for me - I just wanted:

-  to highlight these variables as they indicate not used/unfinished/not cleaned code
-  to have a clean build without warnings

amitdo:
https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

Issue #4010, opened 2023-01-30
HOCR output always sets `textangle 180` and omits baseline info if Tesseract is compiled with `--disable-legacy` 


robertknight:
### Basic Information

tesseract 5.3.0-19-ga3b9ac, compiled with `--disable-legacy`

### Operating System

macOS 13 Ventura

### Compiler

clang 14.0

### Current Behavior

When Tesseract is compiled with `--disable-legacy`, hOCR output reports each line as being upside-down (`textangle 180`) and omits baseline information.

Steps to reproduce:

```
./configure --disable-legacy
./tesseract some-image.jpg output hocr
```

In the generated `output.hocr` file, `ocr_line` entries look like this:

```html
<span class='ocr_line' id='line_1_142' title="bbox 1334 3054 2119 3088; textangle 180; x_size 34; x_descenders 8; x_ascenders 9">
```

### Expected Behavior

If orientation information isn't available I'd expect the image to always be treated as if it were page-up. So entries should look like this:

```html
<span class='ocr_line' id='line_1_142' title="bbox 1334 3054 2119 3088; baseline 0 -8; x_size 34; x_descenders 8; x_ascenders 9">
```

### Suggested Fix

Internally, it looks like the issue is that:

1. `ColumnFinder::text_rotation_` is initialized to a null vector. When the legacy engine is disabled, the `ColumnFinder::CorrectOrientation` function does not get called, and so this vector remains null.
2. This null vector gets propagated to `PageIterator::Orientation`, which does not handle this case correctly, as it converts this null vector to `ORIENTATION_PAGE_DOWN` - https://github.com/tesseract-ocr/tesseract/blob/a3b9acfa4a2f28a9956e830c7354875ebb7213b4/src/ccmain/pageiterator.cpp#L585
3. The HOCR renderer then maps this orientation value to `textangle 180` and omits baseline info

Some fixes I tested locally were to change the initialization of `ColumnFinder::text_rotation_` to be the same as the `norotation` value in `ColumnFinder::CorrectOrientation`, or to change the logic in `PageIterator::Orientation` to handle null rotation vectors by mapping them to `ORIENTATION_PAGE_UP`. I'm happy to submit a PR but I'm not sure the preferred way to go.

amitdo:
#3997 seems related to this issue.

amitdo:
>When the legacy engine is disabled, the `ColumnFinder::CorrectOrientation` function does not get called


[...]

Issue #3997, opened 2023-01-21
PSM_AUTO with jpn_vert gives garbage if the legacy engine is disabled at compile time


danpla:
### Basic Information

Tesseract 5.3.0

### Other Operating System

Windows 7

### uname -a

MINGW32_NT-6.1-7601 PC 3.3.6-341.x86_64 2022-11-20 15:12 UTC x86_64 Msys

### Compiler

GCC 12.2.0

### CPU

Intel Core i7 Q720

### Current Behavior

If tesseact is built without the legacy engine (`--disable-legacy`), recognizing vertical Japanese text with `jpn_vert` (from tessdata_fast) and `PSM_AUTO` (`--psm 3`) gives garbage. Here is an example image:
![1](https://user-images.githubusercontent.com/10910745/213862333-aa41ebee-469e-47a1-bb99-7e5c6d834b76.png)
Here is the output of `tesseract 1.png stdout -l jpn_vert --psm 3`:
```
…4

/
\

09$2pY
コ べ
ほり メー14
```
Here is the correct (albeit with some OCR errors) output that I get either without `--disable-legacy`, or when using `PSM_SINGLE_BLOCK_VERT_TEXT` (`--psm 5`) explicitly regardless of whether the legacy engine is enabled:
```
ケー タイ は
カバ パ バン に
入れ て ある し
```

### Suggested Fix

If this is a kind of feature rather than a bug, it should probably be documented somewhere.

amitdo:
My guess it is caused by the lack of the OSD feature when the legacy engine is disabled.

amitdo:
So it's not a bug but a feature...

You already found the solution  — using `--psm 5`.

danpla:
> My guess it is caused by the lack of the OSD feature when the legacy engine is disabled.

I thought that PSM_AUTO has nothing to do with OSD since there is PSM_AUTO_OSD for that.

amitdo:
>Here is the correct (albeit with some OCR errors) output that I get either without --disable-legacy

With `--psm 3`?

danpla:
> > Here is the correct (albeit with some OCR errors) output that I get either without --disable-legacy
> 
> With `--psm 3`?

Yes.

amitdo:
>I thought that PSM_AUTO has nothing to do with OSD since there is PSM_AUTO_OSD for that.

You are right.

Do you have traineddata.osd in your tessdata dir?

[...]

PR #4041, opened 2023-03-28
Fix issue #4010

amitdo:
Enable some code blocks that were wrongly disabled when the legacy engine is disabled at compile time.


amitdo:
This patch also fixed #3997.

PR #4043, opened 2023-03-30
fix: Fix snap package building

brlin-tw:
This patch fixes the outdated snap package recipe to make the snap buildable with the current Snapcraft release(7.3.1).


stweil (approved):
Thanks.

PR #4066, opened 2023-05-05
Improve format of logging from lstmtraining

stweil:
- always use C ("classic") locale
- limit output of floating point values to 3 digits
- remove unneeded linefeed after log message "wrote checkpoint"


stweil:
I added another commit which limits the floating point values in the names of the checkpoint files to 3 digits, too.

Issue #4062, opened 2023-04-27
Uninitialized variable causes DetectOrientationScript() to fail randomly


todd-richmond:
### Current Behavior

Latest Tesseract 5.3.1 has a bug in DetectOrientationScript() that causes it to randomly fail when attempting to detect image orientation for 90 degree image scans. The problem is that getting a script id can return -1 which then indexes into a "done" array
On my system this fails about 20% of the time on the exact same image input
Error found with valgrind

### Expected Behavior

Trivial fix to not write to done[-1] which then allows DetectOrientationScript() to always return the same results

### Suggested Fix

```
--- src/ccmain/osdetect.cpp.org	2023-04-26 16:26:18.618274574 -0700
+++ src/ccmain/osdetect.cpp	2023-04-26 16:54:37.041375058 -0700
@@ -488,10 +488,12 @@
         }
       }
       // Script already processed before.
-      if (done[id]) {
-        continue;
+      if (id != -1) {
+        if (done[id]) {
+          continue;
+        }
+        done[id] = true;
       }
-      done[id] = true;
 
       unichar = tess_->unicharset.id_to_unichar(choice->unichar_id());

### tesseract -v

_No response_

### Operating System

RHEL 8

### Other Operating System

_No response_

### uname -a

_No response_

### Compiler

gcc 12.2

### CPU

_No response_

### Virtualization / Containers

_No response_

### Other Information

_No response_

stweil:
Is this a regression, or do you see the same issue also with older releases?

todd-richmond:
I'm using my own code that just calls SetImage and then DetectOrientationScript, but if you look at the diff, it is quite trivial so inspection should be enough. "id" can be -1 from choice->script_id() and when it is there is no check before using it as an array index. GitHub "blame" shows it back from 3.0
https://github.com/tesseract-ocr/tesseract/commit/96e8b51feb5b58888f820a7f672097734bc6d4b5

on a side note, this method is much, must faster than using Recognize and iterator calls to determine image rotation as shown in the tesseract samples page

zdenop:
Can you please also provide some example image to test the issue?

todd-richmond:
[scan-90.zip](https://github.com/tesseract-ocr/tesseract/files/11345939/scan-90.zip)
Here is a scanned costco receipt, but most any image should work

stweil:
I tried `tesseract scan-90.tiff - --psm 0` and never get a negative script id. Please paste your test code.

[...]

PR #4067, opened 2023-05-05
Replace bool array by more compact vector

stweil:


stweil:
Related issue: #4062.

todd-richmond:
This fails with a -1 index as noted in #4062. Also, a vector is not any more compact than a stack array because a vector is just an array with more c++ isms around it

stweil:
The old code caused an illegal memory r/w operation with index -1, while the new code causes an exception. That's a clear improvement, even if it does not fix your issue.

And `std::vector<bool>` "is a possibly space-efficient specialization of [std::vector](https://en.cppreference.com/w/cpp/container/vector) for the type bool", see [documentation](https://en.cppreference.com/w/cpp/container/vector_bool).

todd-richmond:
it only throws an exception if you use c++ exceptions which are often disabled due to perf and general usability problems (which is why we do it). A bool vector could be a bitmask underneath, but if it is, you take a perf hit instead - probably not a great tradeoff for a small stack var - especially since the compiler can optimize bool[] any way it sees fit

PR #4068, opened 2023-05-08
Replace deprecated sprintf

stweil:
Either use snprintf or std::stringstream.


stweil:
This fixes related warnings from clang on macOS:

```
src/ccmain/paramsd.cpp:326:5: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
    sprintf(msg_str,
    ^
[...]
```

PR #4070, opened 2023-05-09
Allow for text angle/gradient to be retrieved

### LINKED ISSUE #3836: Detect text rotation without running recognition
As noted in the [documentation ](https://tesseract-ocr.github.io/tessdoc/ImproveQuality.html#rotation--deskewing), Tesseract performs poorly when the page is at an angle (not a multiple of 90 degrees).  This limitation is not problematic from an accuracy standpoint, as Tesseract accurately reports the angle of text lines, so my existing pipeline rotates and re-runs recognition on any image where the angle is significant.  However, this is computationally inefficient as there does not appear to be any way to get the page angle without also running recognition (despite estimating page angle/gradient being [one of the first things calculated](https://github.com/tesseract-ocr/tesseract/blob/60fd2b4abaa9c5c5c42d32db57576bc95d28a78a/src/textord/textord.cpp#L224-L231)). 

Therefore, it would be of significant benefit to be able to get the page angle without running the entire recognition process.  I'll work on a build that does this myself--my initial thought is to add a config option that tells Tesseract to report the page angle and quit early (before recognition) if median line angle is above a user-defined threshold, however let me know if others have thoughts on implementation.

Balearica:
Tesseract already calculates the average gradient (angle) of text lines within `Textord::TextordPage` at present.  The gradient of the text is useful information (as Tesseract performs poorly when the gradient is not [almost] zero), however the average gradient only exists within the [`Textord::TextordPage` function](https://github.com/tesseract-ocr/tesseract/blob/c9895dbad439e5ff42f45ef1d03fe5e8572f2e89/src/textord/textord.cpp#LL222C18-L222C18) at present, with no way for users to access it.  Using the API, getting the gradient currently requires running `Recognize` or `AnalyseLayout` and using the results to manually re-calculate the gradient. 

This PR allows for users to directly retrieve the existing average gradient value calculated in `Textord::TextordPage` using a function named `GetGradient`.  This function can be called any time after `FindLines` has been run.  ~~I also made `FindLines` public so it can be run directly without running `Recognize` or `AnalyseLayout` first (running `AnalyseLayout` would result in paragraph recognition being run twice).~~

I've already used this branch to implement an auto-rotate feature in the latest version of [Tesseract.js](https://github.com/naptha/tesseract.js/) that (unlike adding an auto-rotate pre-processing step) does not negatively impact performance for images without problematic rotation.  A basic script using `GetGradient` is below for demonstrative purposes, along with a test image (named `rotate_image.png` in the code).  Resolves #3836. 

```

int main()
{
    // Initialize tesseract-ocr with English, without specifying tessdata path
    if (api->Init(NULL, "eng")) {
        exit(1);
    }

    // Open input image with leptonica library
    Pix *image = pixRead("rotate_image.png");
    api->SetImage(image);

    // Find lines, get average gradient
    api->AnalyseLayout();
    float gradient = api->GetGradient();

    printf("Average Gradient: %f\n", gradient);

    // Destroy used object and release memory
    api->End();
    delete api;
    pixDestroy(&image);

    return 0;
}

```

```
Average Gradient: 0.085301
```

### DIFF
--- a/src/ccmain/tesseractclass.h
+++ b/src/ccmain/tesseractclass.h
@@ -200,6 +200,9 @@ public:
   const FCOORD &reskew() const {
     return reskew_;
   }
+  float gradient() const {
+    return gradient_;
+  }
   // Destroy any existing pix and return a pointer to the pointer.
   Image *mutable_pix_binary() {
     pix_binary_.destroy();
@@ -1001,6 +1004,7 @@ private:
   int scaled_factor_;
   FCOORD deskew_;
   FCOORD reskew_;
+  float gradient_;
   TesseractStats stats_;
   // Sub-languages to be tried in addition to this.
   std::vector<Tesseract *> sub_langs_;
--- a/src/ccmain/tesseractclass.h
+++ b/src/ccmain/tesseractclass.h
@@ -1004,7 +1004,7 @@ private:
   int scaled_factor_;
   FCOORD deskew_;
   FCOORD reskew_;
-  float gradient_ = 0.0f;
+  float gradient_;
   TesseractStats stats_;
   // Sub-languages to be tried in addition to this.
   std::vector<Tesseract *> sub_langs_;
--- a/src/ccmain/tesseractclass.h
+++ b/src/ccmain/tesseractclass.h
@@ -1004,7 +1004,7 @@ private:
   int scaled_factor_;
   FCOORD deskew_;
   FCOORD reskew_;
-  float gradient_;
+  float gradient_ = 0.0f;
   TesseractStats stats_;
   // Sub-languages to be tried in addition to this.
   std::vector<Tesseract *> sub_langs_;
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -347,6 +347,11 @@ public:
   Pix *GetThresholdedImage();
 
   /**
+   * Return average gradient of lines on page.
+   */
+  float GetGradient();
+
+  /**
    * Get the result of page layout analysis as a leptonica-style
    * Boxa, Pixa pair, in reading order.
    * Can be called before or after Recognize.
@@ -719,6 +724,12 @@ public:
   void set_min_orientation_margin(double margin);
   /* @} */
 
+  /**
+   * Find lines from the image making the BLOCK_LIST.
+   * @return 0 on success.
+   */
+  int FindLines();
+
 protected:
   /** Common code for setting the image. Returns true if Init has been called.
    */
@@ -729,12 +740,6 @@ protected:
    * the source is thresholded to pix instead of the internal IMAGE.
    */
   virtual bool Threshold(Pix **pix);
-
-  /**
-   * Find lines from the image making the BLOCK_LIST.
-   * @return 0 on success.
-   */
-  int FindLines();
 
   /** Delete the pageres and block list ready for a new page. */
   void ClearResults();
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -724,12 +724,6 @@ public:
   void set_min_orientation_margin(double margin);
   /* @} */
 
-  /**
-   * Find lines from the image making the BLOCK_LIST.
-   * @return 0 on success.
-   */
-  int FindLines();
-
 protected:
   /** Common code for setting the image. Returns true if Init has been called.
    */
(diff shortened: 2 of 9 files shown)


amitdo:
>I also made `FindLines` public so it can be run directly without running `Recognize` or `AnalyseLayout` first (running `AnalyseLayout` would result in paragraph recognition being run twice).

So paragraphs detection will run twice if `AnalyseLayout` is followed by `Recognize`. This looks like a bug.

It can be solved by using this condition:

https://github.com/tesseract-ocr/tesseract/blob/424b17f997363670d187f42c43408c472fe55053/src/api/baseapi.cpp#L903-L907

also in `AnalyseLayout`.

Then we won't need to expose `FindLines`. 

What do you think?

egorpugin (approved):

zdenop:
Why not use leptonica solution?
 E.g.

```c++

int main() {
  PIX *pix2;
  l_float32 angle, conf;
  Pix *image = pixRead("rotate_image.png");
  pix2 = pixFindSkewAndDeskew(image, 2, &angle, &conf);
  printf("Skew angle: %7.2f degrees; %6.2f conf\n", angle, conf);
  pixWrite("fixed_rotate_image.png", pix2, IFF_PNG);

  pixDestroy(&image);
  pixDestroy(&pix2);
  return 0;
}
```

amitdo:
>Why not use leptonica solution?

Because it is not necessary.  Tesseract does it anyway.

zdenop:
My understanding is that users want to fix rotation before running OCR. 
This feature will require to use SetImage twice (first to get the angle and then for the corrected image). I guess my proposal will be faster and without the need to touch tesseract API ;-)

Balearica:
> > I also made `FindLines` public so it can be run directly without running `Recognize` or `AnalyseLayout` first (running `AnalyseLayout` would result in paragraph recognition being run twice).
> 
> So paragraph detection will run twice if `AnalyseLayout` is followed by `Recognize`. This looks like a bug.
> 
> It can be solved by using this condition:
> 
> https://github.com/tesseract-ocr/tesseract/blob/424b17f997363670d187f42c43408c472fe55053/src/api/baseapi.cpp#L903-L907
> 
> also in `AnalyseLayout`.
> 
> Then we won't need to expose `FindLines`.
> 
> What do you think?

I think this makes sense conceptually, however if I understand correctly, such a change would impact the results returned by the `AnalyseLayout` API function when run using default settings (`paragraph_text_based` is `true` by default).  I'm always hesitant to advocate for any change that could impact existing code.  If there is opposition to making `FindLines` public I think it would be fine to leave `FindLines` protected and use `AnalyseLayout` instead even if paragraph detection runs multiple times.  I timed these functions using sample image I posted above, and `FindLines` took ~75ms to run while `DetectParagraphs` took ~2ms to run, so the overall impact of this inefficiency appears to be fairly minor.

amitdo:
You are right that my suggestion changes the current behavior,  so it's not a good idea.

I still think we should expose `FindLines` just to make this use case work.

My new suggestion is to choose one of these option:
* Take my previous suggestion, but instead of reusing  `paragraph_text_based`", add a new user configurable variable and use it inside `AnalyseLayout`.
* Use  `AnalyseLayout` without changes. As you said, speed wise the impact of the two calls to `DetectParagraphs` is quite small.

Apart from this small issue, I like the new feature.

Balearica:
@amitdo I changed `FindLines` back to a protected function.  I'm fine with running `AnalyseLayout` as written.

amitdo:
@zdenop,

After reading @Balearica's answer to your question, do you object to merging this PR?

@stweil, can we merge it?

src/ccmain/tesseractclass.h
```diff
@@ -1001,6 +1004,7 @@ class TESS_API Tesseract : public Wordrec {
   int scaled_factor_;
   FCOORD deskew_;
   FCOORD reskew_;
+  float gradient_;
```
stweil, line 1007: The new member variable should be initialized in the constructor, too.

stweil (changes requested):

src/ccmain/tesseractclass.h
Balearica: Edited.

Balearica:
> _First of all_: changing/extending C++-API should be reflected in C-API too.
Good point, I edited the C API to reflect this change.

amitdo:
Regarding semver, the current revision of this PR only **adds** one method to the public API, so the next version should be 5.4.0.

wvanrensburg:
Anyone know when this will get merged? Almost a year now

egorpugin (approved):

stweil (approved):
Thank you!

PR #4073, opened 2023-05-09
Abort with error message if OSD is requested with LSTM-only model

stweil:


stweil:
See related discussion for issue #4062.

PR #4074, opened 2023-05-10
Fix code for generating config files

### LINKED ISSUE #3943: PrintVariables produces config file that ReadConfigFile does not properly read

Balearica:
Resolves #3943, see that issue for a full explanation and previous discussion on this issue. 

In short, the function `ParamUtils::PrintParams` previously created a machine-readable config file with parameter names and values.  It was then edited to add human-readable descriptions to each of the parameters.  While this change undoubtedly made the results easier for users to interpret, the resulting files no longer functioned properly as config files.  This (1) broke the parts of the code that assumed `ParamUtils::PrintParams` created machine-readable config files and (2) meant there was no longer an API function for creating machine-readable config files using the current settings. 

This PR fixes both issues:
1. A boolean parameter has been added to `ParamUtils::PrintParams` to specify whether descriptions of parameters should be printed
   1. This parameter is true by default, so printing to the console (for example) still includes the descriptions
   2. When `ParamUtils::PrintParams` is inarguably being used with the intent of creating a machine-readable config file, the parameter is set to `true`
1. An API function named `DumpVariables` has been added, which creates config files (the old behavior of `PrintVariables`)

### DIFF
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -156,10 +156,18 @@ public:
 
 #endif
 
-  /**
-   * Print Tesseract parameters to the given file.
+  /** 
+   * Print Tesseract parameters to the given file with descriptions of each option. 
+   * Cannot be used as Tesseract configuration file due to descriptions 
+   * (use DumpVariables instead to create config files).
    */
   void PrintVariables(FILE *fp) const;
+
+  /** 
+   * Print Tesseract parameters to the given file without descriptions. 
+   * Can be used as Tesseract configuration file.
+  */
+  void DumpVariables(FILE *fp) const;
 
   /**
    * Get value of named variable as a string, if it exists.
--- a/include/tesseract/capi.h
+++ b/include/tesseract/capi.h
@@ -217,6 +217,10 @@ TESS_API void TessBaseAPIPrintVariables(const TessBaseAPI *handle, FILE *fp);
 TESS_API BOOL TessBaseAPIPrintVariablesToFile(const TessBaseAPI *handle,
                                               const char *filename);
 
+TESS_API void TessBaseAPIDumpVariables(const TessBaseAPI *handle, FILE *fp);
+TESS_API BOOL TessBaseAPIDumpVariablesToFile(const TessBaseAPI *handle,
+                                              const char *filename);
+
 TESS_API int TessBaseAPIInit1(TessBaseAPI *handle, const char *datapath,
                               const char *language, TessOcrEngineMode oem,
                               char **configs, int configs_size);
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -352,9 +352,21 @@ void TessBaseAPI::PrintFontsTable(FILE *fp) const {
 
 #endif
 
-/** Print Tesseract parameters to the given file. */
+/** 
+ * Print Tesseract parameters to the given file with descriptions of each option. 
+ * Cannot be used as Tesseract configuration file due to descriptions 
+ * (use DumpVariables instead to create config files).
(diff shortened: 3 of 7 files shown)


zdenop:
Can you please add it also to  C-API?

Balearica:
> Can you please add it also to C-API?

I added corresponding functions to the C API.

stweil:
> Can you please add it also to C-API?

Do we really need any new API function not only for C++ but also for C?

zdenop:
Why C and C++ API should not be synchronized (providing the same features)?

stweil:
Each additional API creates some costs (implementation, maintenance, code size). And is there still a need for a C API? Maybe we should even consider dropping the whole C API in a future version 6.

If a C API for a new function is really missing it can be added any time later.

stweil:
What about keeping the API unchanged, but fixing the API regression (introduced in commit c51691fdebf186eac9fecb66a9c87d5dcc97f566)?

`tesseract --print-parameters` could still print the parameter information, but use purely private code for that.

That is my preferred solution because I try to minimize API changes.

stweil:
@Balearica, if you want to update your pull request with a new implementation which avoids API changes, it would be nice if you could review your author information. Many projects refuse pull requests without a valid real name for the author, and I think that would be good for Tesseract contributions, too.

PR #4098, opened 2023-07-11
Disable -mfpu=neon for aarch64

hesmar:
This PR removes the -mfpu command line argument on aarch64, because it is not available in gcc on this architecture.
See: https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -258,7 +258,11 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "arm.*")
   set(HAVE_AVX512F FALSE)
   set(HAVE_FMA FALSE)
   set(HAVE_SSE4_1 FALSE)
+
   check_cxx_compiler_flag("-mfpu=neon" HAVE_NEON)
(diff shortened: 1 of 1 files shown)


stweil (changes requested):
I'd prefer to ~add this code at~ move the whole `if` block to line 262 because I think that looks simpler and clearer:

```
  check_cxx_compiler_flag("-mfpu=neon" HAVE_NEON)
  if(HAVE_NEON)
    set(NEON_COMPILE_FLAGS "-mfpu=neon")
    [...]
  endif
```

hesmar:
At the moment, I only moved the NEON_COMPILE_FLAGS up because I am not sure if it is safe to set the ENABLE_LTO only for arm. If you say that this is safe, I can update the PR again.

stweil:
The comment says that link time optimization had a problem on arm/RBPi, therefore I think it was only for 32 bit ARM. Disabling LTO for all aarch64 builds does not look reasonable for me.

PR #4099, opened 2023-07-11
Fix build without git clone in cloned directory

pkubaj:
FreeBSD uses git to manage Ports Tree. Tesseract, when building from the Ports Tree, is built from a tarball that doesn't have .git and then git describe is ran on top of the Ports Tree.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -7,7 +7,7 @@
 # ----------------------------------------
 AC_PREREQ([2.69])
 AC_INIT([tesseract],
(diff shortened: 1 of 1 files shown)


stweil:
So the whole Tesseract source tree is a subdirectory of some directoy structure which is managed with Git? Will your patch also work when configure is running out-of-tree?

stweil:
I nearly never run builds in-tree, but usually have a build directory somewhere below the source root, so `git describe` works fine. Supporting out-of-tree builds is important in my opinion

stweil (approved):
Thank you. I tested the commit. It also works fine with out-of-tree builds.

PR #4118, opened 2023-08-16
Fix Broken URLs in citations.bib

kevinunger:
Context: I noticed that many URLs in the citations.bib file linked to Google papers were returning a 404 error.
Changes Made: Updated these URLs to point to the correct locations.


stweil (approved):
Thank you!

PR #4124, opened 2023-09-12
initDSProfile: correct std::vector usage

stima:
Inside initDSProfile use std::vector::resize instead of std::vector::reserve, due to later usage it by index that triggers debug cheks (reserver does not change size) under MSVC.

### DIFF
--- a/src/opencl/openclwrapper.cpp
+++ b/src/opencl/openclwrapper.cpp
@@ -191,7 +191,6 @@ static ds_status initDSProfile(ds_profile **p, const char *version) {
 
   profile->numDevices = numDevices + 1; // +1 to numDevices to include the native CPU
   profile->devices.resize(profile->numDevices);
-  memset(profile->devices.data(), 0, profile->numDevices * sizeof(ds_device));
 
   next = 0;
   for (i = 0; i < numPlatforms; i++) {
--- a/src/opencl/openclwrapper.cpp
+++ b/src/opencl/openclwrapper.cpp
@@ -174,8 +174,8 @@ static ds_status initDSProfile(ds_profile **p, const char *version) {
   clGetPlatformIDs(0, nullptr, &numPlatforms);
 
   if (numPlatforms > 0) {
-    platforms.reserve(numPlatforms);
-    clGetPlatformIDs(numPlatforms, &platforms[0], nullptr);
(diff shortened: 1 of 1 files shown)


src/opencl/openclwrapper.cpp
```diff
@@ -186,12 +186,12 @@ static ds_status initDSProfile(ds_profile **p, const char *version) {
...
   if (numDevices > 0) {
-    devices.reserve(numDevices);
+    devices.resize(numDevices);
   }
 
   profile->numDevices = numDevices + 1; // +1 to numDevices to include the native CPU
-  profile->devices.reserve(profile->numDevices);
-  memset(&profile->devices[0], 0, profile->numDevices * sizeof(ds_device));
+  profile->devices.resize(profile->numDevices);
+  memset(profile->devices.data(), 0, profile->numDevices * sizeof(ds_device));
```
stweil, line 194: Is the `memset` still needed? The `resize` method should already fill all new elements with zero.

stweil (changes requested):

src/opencl/openclwrapper.cpp
stima: You are right, it initializes it with default. My intent was to minimize changes and follow safe programming (e.g. who said that it wouldn't have any values inserted before). So may adress it here (in that PR) or suggest a small refactoring of that function (as for me it has a mix of c/c++ and unndeded complexity) done by me in next PR.

stweil: Do you plan to use Tesseract with OpenCL or are you already using it? In my tests (done some years ago) it did not work good enough to justify more efforts to improve that code.

stweil: Instead of modifying line 194 I suggest to remove it in the PR here. Additional changes can be done in another PR later.

stweil, line 194:
```suggestion
  profile->devices.resize(profile->numDevices);
```

stima: done

stima: Doing some PoCs to see if I can use it.

stweil (approved):
Thanks!

PR #4132, opened 2023-09-29
Update To README.md

abdoulkkonate:
This commit adds some clarifications about installing and using Tesseract on windows. 
It also adds some ways around some issues I have encountered.

PS: This is for a class assignment.


stweil:
The Tesseract installer for Windows is not part of the official code here, but provided by UB Mannheim.  Therefore I suggest to close this pull request. Maybe you want to add a new Wiki page about Tesseract on Windows to https://github.com/UB-Mannheim/tesseract/wiki.

PR #4141, opened 2023-10-05
Fix loading of sublangs (regression)

stweil:
Model files which where defined with tessedit_load_sublangs in another model file where no longer loaded since Tesseract release 5.0.0-rc2.

Fixes: 9091055783ac8f03 ("Fix loading of additional model files")

### DIFF
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -306,9 +306,10 @@ int Tesseract::init_tesseract(const std::string &arg0, const std::string &textba
   // Add any languages that this language requires
   bool loaded_primary = false;
   // Load the rest into sub_langs_.
-  // A range based for loop does not work here because langs_to_load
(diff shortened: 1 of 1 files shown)


src/ccmain/tessedit.cpp
```diff
@@ -308,7 +308,8 @@ int Tesseract::init_tesseract(const std::string &arg0, const std::string &textba
   // Load the rest into sub_langs_.
   // A range based for loop does not work here because langs_to_load
   // might be changed in the loop when a new submodel is found.
-  for (auto &lang_to_load : langs_to_load) {
```
stweil, line 311: I should have read my own comment which said that using a range based loop is a bad idea here.

amitdo: LOL, but seriously, maybe you should add `WARNING:` in the start of the comment line.

stweil: Good idea. Done now.

PR #4154, opened 2023-10-30
Check if NEON extension are actually available

jpalus:
User may pass own compiler flags to configure which override those provided by project through automake. Therefore it is possible for user on ARM platform to pass CXXFLAGS=-mfpu=vfp which will effectively disable NEON even though used compiler supports -mfpu=neon (since user supplied flags take precedence compiler invocation will use flags: -mfpu=neon -mfpu=vfp). Instead of checking whether compiler supports -mfpu=neon flag, check if NEON extensions are available by checking if __ARM_NEON is defined when compiling with -mfpu=neon and user supplied flags combined.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -178,10 +178,20 @@ case "${host_cpu}" in
     ;;
 
   arm*)
-
-    AX_CHECK_COMPILE_FLAG([-mfpu=neon], [neon=true], [neon=false], [$WERROR])
-    AM_CONDITIONAL([HAVE_NEON], $neon)
-    if $neon; then
+    SAVE_CXXFLAGS="$CXXFLAGS"
+    CXXFLAGS="-mfpu=neon $CXXFLAGS"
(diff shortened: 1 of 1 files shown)


stweil:
Are you sure that this will work on AARCH64?

stweil:
Why should users want to disable NEON instructions?

jpalus:
> Why should users want to disable NEON instructions?

In a distribution I'm targeting ARMv6 CPU with hardware float unit `vfp` so without NEON. That is expressed with compilation flags `-march=armv6 -mfloat-abi=hard -mfpu=vfp` passed to every built package, including `tesseract`, in both `CFLAGS` and `CXXFLAGS`. That's how NEON gets effectively disabled.

Would it hurt if these two sources were compiled with NEON enabled? No since NEON is detected at runtime anyway. That's how automake works unfortunately -- last word is always given to user. The only alternative would be to construct compilation command manually but I doubt it's worth the effort.

PR #4156, opened 2023-10-30
Add new parameter curl_cookiefile for curl_easy_setopt

stweil:
This allows OCR for URLs which require a session cookie.


stweil:
@egorpugin, the CI failures are unrelated to this PR. Please have a look why some sw builds fail.

PR #4169, opened 2023-12-16
Support building and running unit tests with CMake toolchain

phymbert:
**Context:**

- CMake `all` target was actually not building tests with `-DBUILD_TESTS=ON`. Tests were implemented only in make toolchain. Some IDE and developers could not easily work with current cmake/ctest toolchain.

**Changes**: 

1. new `build_tests` and `check` cmake targets which build and run all unit tests (61 as of now). Support of both options `BUILD_TRAINING_TOOLS` and `DISABLED_LEGACY_ENGINE`
2. `BUILD_TESTS` option has been replaced by the standard [CTest](https://cmake.org/cmake/help/latest/module/CTest.html) `BUILD_TESTING`
3. CMake Github workflow has been updated to reflect these changes, and a new check unittest step is added on linux os only
4. unittest/README.md has been updated to document cmake toolchain check target

**Limitations**: 

- [ ] Github workflow cmake-win64.yml is not updated because it is required to install (at least?) icu before test and training compilation.
- [ ] Github workflow cmake.yml does not execute unittest for MacOs because some fonts are missing: `FontUtilsTest.*` are failing. One can probably easy fix it later on.
- [ ] `TENSORFLOW` option yet to be supported in cmake toolchain.
- [X] It is required to enable options `-DBUILD_TRAINING_TOOLS=ON` and `-DBUILD_TESTING=ON` altogether as in the make toolchain, tests cannot compile without it.

**Checks**:
- [X] GitHub workflows [cmake.yml](https://github.com/phymbert/tesseract/actions/runs/7241204284) and [cmake-64.yml](https://github.com/phymbert/tesseract/actions/runs/7241172680/job/19724976343) have been run without  regression.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -64,6 +64,12 @@ jobs:
         with:
              submodules: recursive
 
+      - name: Download fonts, tessdata and langdata required for tests
+        run: |
+          git clone https://github.com/egorpugin/tessdata tessdata_unittest
+          cp tessdata_unittest/fonts/* test/testing/
+          mv tessdata_unittest/* ../
+
       - name: Configure Tesseract (Linux)
         run: |
              mkdir build
@@ -74,6 +80,7 @@ jobs:
                -G Ninja \
                -DCMAKE_BUILD_TYPE=Release \
                -DOPENMP_BUILD=OFF \
+               -DBUILD_TESTING=ON \
                -DCMAKE_CXX_COMPILER=${{ matrix.config.cxx }} \
                -DCMAKE_INSTALL_PREFIX:PATH=inst
         if: runner.os == 'Linux'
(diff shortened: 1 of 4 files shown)


zdenop:
IMO cmake unit tests should be splited into parts based on required dependencies. e.g. ICU for [Windows/VS studio](https://github.com/unicode-org/icu/releases/tag/release-74-2) could be downloaded automatically, but Pango for VS code not (+you can not build it with cmake unless you). 
On the other hand, [mingw/msys2](https://packages.msys2.org/package/mingw-w64-x86_64-pango) should have no problem...

Also, it makes sense for me if I can skip Tesseract library tests if I do modifications only in training tools.

I doubt if it makes sense to run cmake unitest as github actions - there are already running unit tests, that will catch potential problems anyway.

phymbert:
Closing as it requires additional effort mentionned by @zdenop and cannot be merged as is.

Issue #3798, opened 2022-04-26
Add option to use demo.processed.tif image to create the demo.pdf


psychonaute:
Tesseract is doing a fantastic Job at processing the input image! 

I'd love to be able to use the very lightweight processed image ( `tesseract demo.jpg out get.images` -> `demo.processed.tif` ) to create the pdf instead of the substantially heavier original image.

It seems that option isn't implemented, but tesseract project is massive so I might have overlook something...

amitdo:
Related feature request: #210

PR #4171, opened 2023-12-19
PDF Renderer: allow to specify an alternate image or a custom resolution.

### LINKED ISSUE #210: Feature request: Add optional input for alternate image to use when sandwiching OCR data
Hi,
I recently started using tesseract to help unclutter my desk at home, so forgive me if this is a n00b question/request.

I use textcleaner from Fred's ImageMagick Scripts to cleanup my scanned images for better OCR accuracy. However, the images that are optimized for OCR do not necessarily look good from a human standpoint, and I would like the final OCR'd PDF to look visually identical to the original scan.

So here's my feature request: Add an optional argument to take a cleaned image. Example invocation: tesseract -l eng -psm 4 --cleaned-image ${SRC}_cleaned.pnm ${SRC}.pnm out pdf

It will use ${SRC}.pnm to generate the final PDF image but layout detection, character recognition, etc. will be done using the --cleaned-image argument for better accuracy. That way the user will be given a final PDF that looks like the original but searches as well as the cleaned-up image.

I'd be surprised if nobody has already thought of this, so maybe work is already underway or maybe it's not possible. Thoughts?

phymbert:
**Motivation**

Input images passed to OCR are often pre-processed (higher resolution, grayed, etc...).
It can be useful to specify an alternate image or a lower resolution in renderer, especially for a searchable pdf export.

**Proposed changes**

1. Added `TessResultRenderer::SetRenderingImage`  or `TessResultRenderer::SetRenderingResolution` methods allow to programmatically change image or resolution to render before adding image to the renderer
2. New `rendering_dpi` param allows to override the output resolution by scaling the source image
3. Added few pdfrenderer tests
4. Fix missing `pdf.ttf` font in the cmake install target

These changes might resolve #210 and #3798 features request.

***Checks***

- [X] `make check` passed locally on ubuntu 23.10
- [ ] GitHub workflows passed

### DIFF
--- a/include/tesseract/renderer.h
+++ b/include/tesseract/renderer.h
@@ -106,6 +106,23 @@ public:
     return imagenum_;
   }
 
+  /**
+   * Specifies an alternate image to render with the extracted text.
+   * It must be called after BeginDocument and before AddImage.
+   */
+  void SetRenderingImage(Pix *rendering_image) {
+    rendering_image_ = rendering_image;
+  }
+
+  /**
+   * Specifies the expected rendering resolution.
+   * If not set, rendering_dpi api params will be used, else the source image
+   * resolution.
+   */
+  void SetRenderingResolution(int rendering_dpi) {
+    rendering_dpi_ = rendering_dpi;
+  }
+
 protected:
   /**
    * Called by concrete classes.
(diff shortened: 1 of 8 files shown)


zdenop:
cmake does not install a PDF font file. It was the old way, how to handle font in pdf. Now it is automatically [included in library](https://github.com/tesseract-ocr/tesseract/blob/main/src/api/pdf_ttf.h)

zdenop:
@jbreiden @jbreiden2 : Jeff can you have a look at this?

phymbert:
> cmake does not install a PDF font file. It was the old way, how to handle font in pdf. Now it is automatically [included in library](https://github.com/tesseract-ocr/tesseract/blob/main/src/api/pdf_ttf.h)

Thanks, @zdenwas, for the explanation. I was confused with the `tessdata/Makefile.am`, and I will remove it. 
Let me submit pdfrendrerer test fixes, it failed on some platforms. 

@jbreiden @jbreiden2, a better way to check pdf files generated than file minimum size is welcomed

phymbert:
It looks there is a little interest, that happens :) Thanks all

stweil:
Hi, it's not unusual that pull requests take some time before they are merged. That does not necessarily mean that there is little interest, but there is only a small number of people who contribute to pull requests by adding comments or testing them.

phymbert:
No worries at all, I just saw it open on my to-do list for a while, so I preferred to close. Thanks for your feedback, I understand, reopened, no hurry.

zdenop:
Since it extends the API functionality, it should be included in the 5.4.0 release.

stweil:
I rebased this pull request and fixed a merge conflict.

include/tesseract/renderer.h
```diff
@@ -151,6 +183,8 @@ class TESS_API TessResultRenderer {
   const char *file_extension_; // standard extension for generated output
   std::string title_;          // title of document being rendered
   int imagenum_;               // index of last image added
+  Pix *rendering_image_;       // Image to render with the extracted text
+  int rendering_dpi_;          // Resolution of the rendering_image
```
stweil, line 187: Those new member variables and the related methods are currently only used by the PDF renderer.

Could they also be useful for other renderers in the future?

If not: what about moving them to TessPDFRenderer?

zdenop:
What about implementing this feature also to tesseract executable as a command line option?

stweil:
Isn't that already possible with `-c`?

zdenop:
> Isn't that already possible with -c?

With `-c` I can set `rendering_dpi`. How can I set an image for `SetRenderingImage`?

stweil:
Tesseract can create multi-page PDF files when it is called with a list of images. Ideally that should also work with alternate images.

stweil:
> > Isn't that already possible with -c?
> 
> With `-c` I can set `rendering_dpi`. How can I set an image for `SetRenderingImage`?

Would it be possible to implement the desired features by only adding new Tesseract parameters – without any change of the C / C++ API?

PR #4178, opened 2024-01-11
Fix a tiny typo in `publictypes.h`

sadra-barikbin:
To fix a tiny typo in a comment in `publictypes.h`.

### DIFF
--- a/include/tesseract/publictypes.h
+++ b/include/tesseract/publictypes.h
@@ -67,7 +67,7 @@ enum PolyBlockType {
   PT_COUNT
 };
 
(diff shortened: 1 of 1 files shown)


stweil (commented):
Why do you think that this is a typo? The current comment looks good for me.

sadra-barikbin:
The function checks its input to be a line type, whether horizontal or vertical  but the comment says that the function checks if the input is a "horizontal" line type, excluding the vertical line type.

stweil (approved):
Thank you.

PR #4186, opened 2024-02-01
`oss-fuzz-build.sh`: explicitly pass `--disable-openmp`

LebedevRI:
https://github.com/google/oss-fuzz/pull/11548 will add openmp runtime to oss-fuzz, and that breaks the build of this project. Explicitly opt-out to keep the build green.


LebedevRI:
(This may be not needed, i'll reopen if still needed. Sorry for the noise.)

PR #4187, opened 2024-02-04
Make `TessdataManager` able to save archive using LibArchive

sadra-barikbin:
- Made `TessDataManager` able to save archive using LibArchive
 - Added `-t` option to `combine_tessdata` to transform proprietary `.traineddata` to archive file.

### DIFF
--- a/src/ccutil/tessdatamanager.cpp
+++ b/src/ccutil/tessdatamanager.cpp
@@ -85,6 +85,32 @@ bool TessdataManager::LoadArchiveFile(const char *filename) {
   }
   return result;
 }
+
+bool TessdataManager::SaveArchiveFile(const char *filename) const{
+  bool result = false;
+  archive *a = archive_write_new();
+  archive_entry *ae = archive_entry_new();
+  if (a != nullptr) {
+    archive_write_set_format_zip(a);
+    archive_write_open_filename(a, filename);
+    std::string filename_str = filename;
+    filename_str += ".";
+    archive_entry_set_filetype(ae, AE_IFREG);
+    archive_entry_set_perm(ae, 333);
+    for (unsigned i = 0; i < TESSDATA_NUM_ENTRIES; ++i) {
+      if (!entries_[i].empty()) {
+        archive_entry_set_pathname(ae, (filename_str + kTessdataFileSuffixes[i]).c_str());
+        archive_entry_set_size(ae, entries_[i].size());
+        archive_write_header(a, ae);
+        archive_write_data(a, &entries_[i][0], entries_[i].size());
+      }
+    }
+    result = archive_write_close(a) == ARCHIVE_OK;
+    archive_write_free(a);
+    return result;
+  }
+  return result;
+}
 #endif
 
 bool TessdataManager::Init(const char *data_file_name) {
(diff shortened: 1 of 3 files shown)


sadra-barikbin:
@stweil , shall I add a test?

src/ccutil/tessdatamanager.cpp
```diff
@@ -162,12 +188,16 @@ void TessdataManager::OverwriteEntry(TessdataType type, const char *data, int si
...
 // Saves to the given filename.
 bool TessdataManager::SaveFile(const char *filename, FileWriter writer) const {
-  // TODO: This method supports only the proprietary file format.
+// TODO: This method supports only the proprietary file format.
   ASSERT_HOST(is_loaded_);
   std::vector<char> data;
   Serialize(&data);
   if (writer == nullptr) {
+#if defined(HAVE_LIBARCHIVE)
+    return SaveArchiveFile(filename);
```
stweil, line 197: With this change `TessdataManager::SaveFile` will always write traineddata files in ZIP format which are incompatible with Tesseract binaries which were build without LibArchive. I'm afraid that would cause problems for a lot of people.

stweil:
> Added `-t` option to `combine_tessdata` to transform proprietary `.traineddata` to archive file.

I'd prefer a more general solution which allows different target formats. In addition it should allow writing to a different output file and use long options. So the syntax might look like this:

 ```
Usage: combine_tessdata --convert [--format TARGET_FORMAT] INFILE [OUTFILE]
```

`TARGET_FORMAT` would default to `zip`, but should allow any typical file extension of archive files and also `traineddata` for conversions into the proprietary Tesseract format.

egorpugin:
Isn't automatic format simpler?
Libarchive writes files based on their extension.
If we write .zip - it will be .zip.
.tar.gz - tar.gz
.tar.xz or lz - ...

src/ccutil/tessdatamanager.cpp
egorpugin: I thought libarchive can deduce archive types.

stweil:
That's right, and this feature of LibArchive would also be used to implement my suggested solution.

If we implement support for `combine_tessdata --convert eng.traineddata eng.zip`, that would create a filename which is currently unsupported by `tesseract`, so an additional renaming `mv eng.zip eng.traineddata` would be required. Should we extend the Tesseract code to support more extensions than the current `.traineddata`, so `-l eng.zip` would work? Or do you think of another solution? Maybe `eng.traineddata.zip`?

src/ccutil/tessdatamanager.cpp
stweil: Yes, it can, but the current code uses `archive_write_set_format_zip`. And of course libarchive cannot write the proprietary traineddata format.

zdenop:
Is there also intention to read such converted data by tesseract? 

If yes, than please be careful about changing extension: it will break a lot of workflows that looks for available/installed languages (AFAIR also [GetAvailableLanguagesAsVector](https://github.com/tesseract-ocr/tesseract/blob/a7f860a41786358ed27adc3061b857c03e548b2f/src/api/baseapi.cpp#L472) a.k.a `tesseract --list-langs`)

Nowadays it is quite common to use private file extension instated of indication it is archive (e.h. xlsx, odt are zip archives)

On other hand:  if file extension  will not be changed and tesseract will be build without libarchive support, that has to be improved error handling why tesseract is not able read traineddata...

sadra-barikbin:
I thought this feature is decided on to be implemented. Now it seems it's an arguable one. My own requirement i.e. inspecting config file in the `.traineddata` file and possibly overwriting it is fulfilled with the sequence of `combine_tessdata -e` and `combine_tessdata -o`.

PR #4189, opened 2024-02-10
Fix grey result of indexed PNG in pdfrenderer.

sjbronner:
When giving tesseract PNG images with an indexed color map with the "pdf" output, these images come out grayscale. Removing a special case for PNG images in pdfrenderer fixes that.

I tested the patch with PNG images containing indexed color maps and containing rgb content.

### DIFF
--- a/src/api/pdfrenderer.cpp
+++ b/src/api/pdfrenderer.cpp
@@ -664,8 +664,6 @@ bool TessPDFRenderer::BeginDocumentHandler() {
 bool TessPDFRenderer::imageToPDFObj(Pix *pix, const char *filename, long int objnum,
                                     char **pdf_object, long int *pdf_object_size,
                                     const int jpg_quality) {
-  int sad;
-
   if (!pdf_object_size || !pdf_object) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


src/api/pdfrenderer.cpp
```diff
@@ -674,14 +676,7 @@ bool TessPDFRenderer::imageToPDFObj(Pix *pix, const char *filename, long int obj
...
   L_Compressed_Data *cid = nullptr;
-
-  int sad = 0;
-  if (pixGetInputFormat(pix) == IFF_PNG) {
-    sad = pixGenerateCIData(pix, L_FLATE_ENCODE, 0, 0, &cid);
-  }
-  if (!cid) {
-    sad = l_generateCIDataForPdf(filename, pix, jpg_quality, &cid);
-  }
+  sad = l_generateCIDataForPdf(filename, pix, jpg_quality, &cid);
```
egorpugin, line 679: Just make it `auto sad = ...`

sjbronner:
Absolutely.
[test-indexed-before.pdf](https://github.com/tesseract-ocr/tesseract/files/14284420/test-indexed-before.pdf)
[test-indexed-after.pdf](https://github.com/tesseract-ocr/tesseract/files/14284422/test-indexed-after.pdf)
[test-indexed.png](https://github.com/tesseract-ocr/tesseract/assets/3920512/5629b5af-ce99-4975-a56a-b5c3a2540a96)

PR #4195, opened 2024-03-05
normstrngs: add more hyphens and quotes

bertsky:
IIUC, this is used when a language model / dict / dawg is in place to normalise the characters before applying hyphenation and quoting rules (which are not all covered by punctuation dawgs etc).


stweil:
The 2nd commit was not strictly necessary. It's quite common that lists are terminated like that because adding more entries at the end is easier.

amitdo:
>The 2nd commit was not strictly necessary.

3 compilers disagreed with your above statement :-)

bertsky:
sorry, I forgot to increase the constants for the number of initialisers, hence the failing build.

will make a new PR with that fix right now

PR #4196, opened 2024-03-05
normstrngs: fix initialiser length constants

bertsky:
follow-up to #4195


stweil:
Thanks. Instead of fixing the constants, I decided to remove them (see my latest commits in the main branch).

PR #4211, opened 2024-03-13
Improved Diplopia Solution

woodjohndavid:
I previously created a pull request from my branch JDWDIPLOPIA.  This solution, while better than without it as far as diplopia is concerned, was a limited solution.

I am now creating a pull request from my branch JDWDIPLOPIA2.  I believe that this is a more complete solution to the diplopia issue.

### DIFF
--- a/src/lstm/recodebeam.cpp
+++ b/src/lstm/recodebeam.cpp
@@ -34,6 +34,15 @@ const int RecodeBeamSearch::kBeamWidths[RecodedCharID::kMaxCodeLen + 1] = {
 
 static const char *kNodeContNames[] = {"Anything", "OnlyDup", "NoDup"};
 
+// setting to enable diplopia removal functionality
+// needs to be config setting in future
+static const bool kRemoveDiplopia = true;
+// The maximum diplopia gap is the maximum number of timesteps
+// which the peak value of possible diplopia candidates can be apart
+// in order to be considered as genuine diplopia
+// needs to be config setting in future
+static const int kMaxDiplopiaGap = 2;
+
 // Prints debug details of the node.
 void RecodeNode::Print(int null_char, const UNICHARSET &unicharset,
                        int depth) const {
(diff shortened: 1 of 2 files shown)


stweil:
I don't see a commit which changes the Tesseract code. Is something missing?

woodjohndavid:
OK thanks @stweil 

This may leave you somewhat concerned about my code changes, but I can assure you that I am a competent developer.  I just never use GitHub other than with Tesseract, hence incompetent in that regard.

Note by the way, similar to the other pull request I generated, there are some new configuration values that can only be set in code as it stands, but should be made into available settings.  I have not yet figured out the mechanism for doing that.  If the diplopia changes I am proposing turn out to be useful, hopefully someone else familiar with the settings approach could take care of that.  These configuration values are: 

- bool kRemoveDiplopia - if true, enables diplopia removal functionality. If false, my changes have no effect
- int kMaxDiplopiaGap - number of timesteps apart to be considered diplopia, default 2

amitdo:
Apart from testing that this patch has a positive effect on the diplopia issue, people should test if there is no negative effect in other place, like dropping correct characters.

src/lstm/recodebeam.cpp
```diff
@@ -88,15 +97,369 @@ void RecodeBeamSearch::Decode(const NetworkIO &output, double dict_ratio,
...
+      if (top_heap_.size() > top_n) {
+        top_heap_.Pop(&entry);
+      }
+    }
+  }
+
+  std::vector<TopPair> topn_for_step;
+  topn_for_step.resize(top_heap_.size());
+  
+  int i = top_heap_.size();
```
stweil, line 138: We prefer `auto` instead of `int` where this is possible (not only here).

src/lstm/recodebeam.cpp
```diff
@@ -88,15 +97,369 @@ void RecodeBeamSearch::Decode(const NetworkIO &output, double dict_ratio,
...
+  float second_diplopia_max = diplopia_max_;
+  int second_diplopia_max_timestep = diplopia_max_timestep_;
+
+  if (first_diplopia_max >= second_diplopia_max) {
+    chosen_diplopia_char_ = second_diplopia_char_;
+    start_diplopia_ = start_second_diplopia;
+    end_diplopia_ = end_second_diplopia;
+    diplopia_max_ = second_diplopia_max;
+  }
+  else {
```
stweil, line 212: The tesseract coding style is different. Please try to use the same style as in the rest of the tesseract code.

stweil (changes requested):

src/lstm/recodebeam.cpp
woodjohndavid: I have updated the code for the if/else style.   If there are other stylistic changes needed, please let me know what they are and I will attempt to address them.

As far as the int vs auto, I have not made any changes for that as yet.  I do see in the original code in areas that I have not modified there are many examples where auto is not used.

Thanks,

Dave

woodjohndavid:
Please note that this change is likely not appropriate for those using Tesseract for natural language recognition using relevant dawgs.  It is primarily intended for those (like myself) using Tesseract to scan technical data, looking for exact character by character recognition.

Papucs:
Hi,
Is there any chance this will be merged in the near future?

stweil:
We know that better models reduce the number of diplopia cases. That should work for all languages and scripts and is my preferred solution for the problem, instead of very special solutions in the software with unknown side effects.

woodjohndavid:
Hi all:

Just a few comments from myself, the author of this diplopia fix:

1. As mentioned earlier, this change is intended primarily for those who are looking for a character-by-character OCR solution, not so much textual word recognition relying on the dawg functionality.

2. In my testing on my limited set of data, it seems to work well, pretty much entirely eliminating diplopia.  However, it needs much more testing on broader sample sets by others.

3. I see no reason at all that this will not work for any other non-Latin character set.  Of course this would need to be tested, but I don't believe that there is anything in the logic that is character set specific.

4. Indeed it needs a user configurable variable to turn this function on and off.  I have no experience with the config functionality, and it seems pretty arcane in the brief time I have spent looking at that code.  Surely there are available developers with detailed knowledge of that code that can add this config variable easily.  There is already a boolean on/off variable in the fix code, just no way for a user to set it.

For my own purposes, I don't need this fix incorporated into the main Tesseract code, as I am quite happy running my own custom copy.  However, I have taken the time to promote it in this pull request to try to make a modest but hopefully meaningful contribution to Tesseract.  If it is not going to see the light of day for others, that is unfortunate.

Regards,

Dave

PR #4215, opened 2024-03-25
Support training without lstmf files

stweil:
The list of the training or validation files may now contain .png files instead of .lstmf files, so it is no longer necessary to spend time and disc space for creating .box and .lstmf files for a training.

### DIFF
--- a/src/ccstruct/imagedata.cpp
+++ b/src/ccstruct/imagedata.cpp
@@ -33,7 +33,8 @@
 
 #include <allheaders.h> // for pixDestroy, pixGetHeight, pixGetWidth, lept_...
 
-#include <cinttypes> // for PRId64
+#include <cinttypes>    // for PRId64
+#include <fstream>      // for std::ifstream
 
 namespace tesseract {
 
@@ -546,6 +547,31 @@ bool DocumentData::ReCachePages() {
     delete page;
   }
   pages_.clear();
+#if !defined(TESSERACT_IMAGEDATA_AS_PIX)
+  auto name_size = document_name_.size();
+  if (name_size > 4 && document_name_.substr(name_size - 4) == ".png") {
+    // PNG image given instead of LSTMF file.
+    std::string gt_name = document_name_.substr(0, name_size - 3) + "gt.txt";
+    std::ifstream t(gt_name);
+    std::string line;
+    std::getline(t, line);
+    t.close();
+    ImageData *image_data = ImageData::Build(document_name_.c_str(), 0, "", nullptr, 0, line.c_str(), nullptr);
+    Image image = pixRead(document_name_.c_str());
+    image_data->SetPix(image);
+    pages_.push_back(image_data);
+    loaded_pages = 1;
+    pages_offset_ %= loaded_pages;
+    set_total_pages(loaded_pages);
+    set_memory_used(memory_used() + image_data->MemoryUsed());
+#if 0
+    tprintf("Loaded %zu/%d lines (%d-%zu) of document %s\n", pages_.size(),
+            loaded_pages, pages_offset_ + 1, pages_offset_ + pages_.size(),
+            document_name_.c_str());
+#endif
+    return !pages_.empty();
+  }
+#endif
   TFile fp;
   if (!fp.Open(document_name_.c_str(), reader_) ||
       !fp.DeSerializeSize(&loaded_pages) || loaded_pages <= 0) {


src/ccstruct/imagedata.cpp
```diff
@@ -546,6 +547,30 @@ bool DocumentData::ReCachePages() {
     delete page;
   }
   pages_.clear();
+#if !defined(TESSERACT_IMAGEDATA_AS_PIX)
+  if (document_name_.ends_with("png")) {
```
stweil, line 551:
```suggestion
  auto name_size = document_name_.size();
  if (name_size > 4 && document_name_.substr(name_size - 4) == ".png") {
```

src/ccstruct/imagedata.cpp
```diff
@@ -546,6 +547,31 @@ bool DocumentData::ReCachePages() {
     delete page;
   }
   pages_.clear();
+#if !defined(TESSERACT_IMAGEDATA_AS_PIX)
+  auto name_size = document_name_.size();
+  if (name_size > 4 && document_name_.substr(name_size - 4) == ".png") {
+    // PNG image given instead of LSTMF file.
+    std::string gt_name = document_name_.substr(0, document_name_.length() - 3) + "gt.txt";
```
stweil, line 554:
```suggestion
    std::string gt_name = document_name_.substr(0, name_size - 3) + "gt.txt";
```

src/ccstruct/imagedata.cpp
```diff
@@ -546,6 +547,31 @@ bool DocumentData::ReCachePages() {
...
+    pages_.push_back(image_data);
+    loaded_pages = 1;
+    pages_offset_ %= loaded_pages;
+    set_total_pages(loaded_pages);
+    set_memory_used(memory_used() + image_data->MemoryUsed());
+    if (true) {
+      tprintf("Loaded %zu/%d lines (%d-%zu) of document %s\n", pages_.size(),
+              loaded_pages, pages_offset_ + 1, pages_offset_ + pages_.size(),
+              document_name_.c_str());
+    }
```
stweil, line 571: This debug code should be removed before merging.

egorpugin (approved):

src/ccstruct/imagedata.cpp
```diff
@@ -546,6 +547,31 @@ bool DocumentData::ReCachePages() {
     delete page;
   }
   pages_.clear();
+#if !defined(TESSERACT_IMAGEDATA_AS_PIX)
+  auto name_size = document_name_.size();
+  if (name_size > 4 && document_name_.substr(name_size - 4) == ".png") {
```
zdenop, line 552: Why not use [to use std::filesystem::path](https://en.cppreference.com/w/cpp/filesystem/path/extension) as in https://github.com/tesseract-ocr/tesseract/blob/b217cfcfa3b5c41e20414428e5d4700af130b192/src/training/unicharset_extractor.cpp#L74

egorpugin: Path refactoring should be done as more general procedure. It is ok to keep std::string in very local code.

src/ccstruct/imagedata.cpp
```diff
@@ -546,6 +547,31 @@ bool DocumentData::ReCachePages() {
     delete page;
   }
   pages_.clear();
+#if !defined(TESSERACT_IMAGEDATA_AS_PIX)
```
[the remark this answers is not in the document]
stweil: `TESSERACT_IMAGEDATA_AS_PIX` is used at several places in imagedata.cpp, but not defined anywhere. It's possible of course to define it manually for a build process.

My new code is only active if that macro is undefined, so it is active by default.

src/ccstruct/imagedata.cpp
stweil: So would you prefer using `std::filesystem::path::extension()` here or not?

egorpugin: It's ok as is. fs::paths can be refactored for 6.0 wit api changes/breakage.

zdenop: @egorpugin: `document_name_` is not part of tesseract API - it is the private object used in imagedata.cpp only, So I do not think it will break something or change API.

@stweil: I am fine with the current PR code, just I think we should step by step use c++17 features.

yaofuzhou:
> The list of the training or validation files may now contain .png files instead of .lstmf files, so it is no longer necessary to spend time and disc space for creating .box and .lstmf files for a training.

Hi - In my understanding of Tesseract before your merge, `list.train` and `list.eval` list the `.lstmf` files used in training and evaluation, where each `.lstmf` file is the product of an image and its `.box` file. Without the `.lstmf` files, how should I tell the Tesseract training program which files to be used in the training process? Thanks!

stweil:
You still use list.train and list.eval, but they can now include .png files instead of .lstmf files.

yaofuzhou:
> You still use list.train and list.eval, but they can now include .png files instead of .lstmf files.

Sorry for the silly question, but may I assume that the corresponding `.box` files will be automatically fetched?

stweil:
Neither .box nor .lstmf files are used if you provide only .png files in the lists.

amitdo:
I assume the textline within the png image should not have any padding, is that right?

stweil:
Our line images are from real prints and typically have padding.

PR #4220, opened 2024-04-04
Remove unsupported OpenCL code and related API functions

stweil:


egorpugin (approved):
For 6.0?

stweil:
I thought about 5.4.0 (because it only removes two unused API functions). We could even keep those two functions for now and always return a null pointer to avoid any API change.

amitdo:
Thank you!

PR #4223, opened 2024-04-10
facilitate vectorization for generic build

heshpdx:
In the process of developing tesseract for [SPEC CPUv8](https://www.spec.org/cpuv8/) with @zdenop 's help, we used the generic code path to future-proof the benchmark for architectures that do not yet exist. Hence we disable the intrinsics, and rather rely on the C++ code directly. Here, we found an opportunity to assist the compiler in the `IntSimdMatrix::MatrixDotVector` function. We split the loop so it first does all the work in chunks of four, and then clean up with whatever remainder is left over. If there is a guarantee that `w.dim1()` is always a multiple of four, then the remainder code is not required. So, this offers some perf improvement on the generic code path, depending on your compiler and system.

### DIFF
--- a/src/arch/intsimdmatrix.cpp
+++ b/src/arch/intsimdmatrix.cpp
@@ -80,7 +80,32 @@ void IntSimdMatrix::MatrixDotVector(const GENERIC_2D_ARRAY<int8_t> &w,
   int num_out = w.dim1();
   int num_in = w.dim2() - 1;
   // Base implementation.
-  for (int i = 0; i < num_out; ++i) {
+  int i;
+  // Break up into chunks of four to facilitate vectorization
+  for (i = 0; i < (num_out / 4) * 4; i += 4) {
+    const int8_t *wi0 = w[i + 0];
+    const int8_t *wi1 = w[i + 1];
+    const int8_t *wi2 = w[i + 2];
+    const int8_t *wi3 = w[i + 3];
+    int total0 = 0;
+    int total1 = 0;
+    int total2 = 0;
+    int total3 = 0;
+    for (int j = 0; j < num_in; ++j) {
+      total0 += wi0[j] * u[j];
+      total1 += wi1[j] * u[j];
+      total2 += wi2[j] * u[j];
+      total3 += wi3[j] * u[j];
+    }
+    // Add in the bias and correct for integer values.
+    v[i + 0] = (total0 + wi0[num_in] * INT8_MAX) * scales[i + 0];
+    v[i + 1] = (total1 + wi1[num_in] * INT8_MAX) * scales[i + 1];
+    v[i + 2] = (total2 + wi2[num_in] * INT8_MAX) * scales[i + 2];
+    v[i + 3] = (total3 + wi3[num_in] * INT8_MAX) * scales[i + 3];
+  }
+
+  // Capture the remainder mod four
+  for (; i < num_out; i++) {
     const int8_t *wi = w[i];
     int total = 0;
     for (int j = 0; j < num_in; ++j) {
--- a/src/arch/intsimdmatrix.cpp
+++ b/src/arch/intsimdmatrix.cpp
@@ -105,7 +105,7 @@ void IntSimdMatrix::MatrixDotVector(const GENERIC_2D_ARRAY<int8_t> &w,
   }
 
   // Capture the remainder mod four
-  for (; i < num_out; i++) {
+  for (; i < num_out; ++i) {
     const int8_t *wi = w[i];
     int total = 0;
     for (int j = 0; j < num_in; ++j) {


src/arch/intsimdmatrix.cpp
```diff
@@ -80,7 +80,32 @@ void IntSimdMatrix::MatrixDotVector(const GENERIC_2D_ARRAY<int8_t> &w,
...
+    }
+    // Add in the bias and correct for integer values.
+    v[i + 0] = (total0 + wi0[num_in] * INT8_MAX) * scales[i + 0];
+    v[i + 1] = (total1 + wi1[num_in] * INT8_MAX) * scales[i + 1];
+    v[i + 2] = (total2 + wi2[num_in] * INT8_MAX) * scales[i + 2];
+    v[i + 3] = (total3 + wi3[num_in] * INT8_MAX) * scales[i + 3];
+  }
+
+  // Capture the remainder mod four
+  for (; i < num_out; i++) {
```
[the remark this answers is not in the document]
stweil: Let's keep the old variant, even if it does not make a difference (but sometimes people suggest that `++i` is better than `i++`). We also have `++j` below, so this is also more uniform.

stweil:
That looks interesting. Did you already run benchmarks which compare the old and the new code?

heshpdx:
> That looks interesting. Did you already run benchmarks which compare the old and the new code?

There was a couple percentage points improvement on my Ampere Altra with gcc-13 -O3, however your mileage will vary.

src/arch/intsimdmatrix.cpp
```diff
@@ -80,7 +80,32 @@ void IntSimdMatrix::MatrixDotVector(const GENERIC_2D_ARRAY<int8_t> &w,
...
+    }
+    // Add in the bias and correct for integer values.
+    v[i + 0] = (total0 + wi0[num_in] * INT8_MAX) * scales[i + 0];
+    v[i + 1] = (total1 + wi1[num_in] * INT8_MAX) * scales[i + 1];
+    v[i + 2] = (total2 + wi2[num_in] * INT8_MAX) * scales[i + 2];
+    v[i + 3] = (total3 + wi3[num_in] * INT8_MAX) * scales[i + 3];
+  }
+
+  // Capture the remainder mod four
+  for (; i < num_out; ++i) {
```
heshpdx, line 108: Oh wait, I cannot take the `++i` change. Because this "remainder" block relies on the value of `i` coming into the loop. If we pre-increment `i` then we will skip one element and that would be incorrect. How can we undo the commit? (I'm not a github wizard... I could resubmit the PR?)

zdenop: try this: https://gist.github.com/gunjanpatel/18f9e4d1eb609597c50c2118e416e6a6

stweil: Why would post-increment make a difference here?

stweil: You could suggest your original code here and commit it, but it is not necessary to revert anything. The new code will work fine.

zdenop (approved):

src/arch/intsimdmatrix.cpp
heshpdx: Oh yes you are correct, sorry it was a late night 😃

heshpdx:
Looks like the macos-13 runner failed due to timeout. Should we kick it and try again?

egorpugin:
No need. It happens.

stweil (approved):

stweil:
Thanks!

zdenop:
Just to give some hint about this patch:
I made the [speed test](https://github.com/tesseract-ocr/tessdoc/blob/main/Benchmarks.md) on old linux computer with following results:
Average speed before applying this code:  61 s.
Average speed after applying this code:  37 s.

Tesseract was build with default configuration options.
Interestingly - when I tested applying `#pragma omp parallel for`  for this patch, the speed drop down to 88 s. So using OpemMP should be carefully tested on old systems.

CPU info:

```sh
Architecture:           x86_64
  CPU op-mode(s):       32-bit, 64-bit
  Address sizes:        36 bits physical, 48 bits virtual
  Byte Order:           Little Endian
CPU(s):                 1
  On-line CPU(s) list:  0
Vendor ID:              GenuineIntel
  Model name:           Intel(R) Core(TM)2 CPU          4300  @ 1.80GHz
    CPU family:         6
    Model:              15
    Thread(s) per core: 1
    Core(s) per socket: 1
    Socket(s):          1
    Stepping:           2
    CPU max MHz:        1800.0000
    CPU min MHz:        1200.0000
    BogoMIPS:           3591.24
    Flags:              fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse ss
                        e2 ht tm pbe syscall nx lm constant_tsc arch_perfmon pebs bts rep_good nopl cpuid aperfmperf pni dtes64
                        monitor ds_cpl est tm2 ssse3 cx16 xtpr pdcm lahf_lm pti dtherm
Caches (sum of all):
  L1d:                  32 KiB (1 instance)
  L1i:                  32 KiB (1 instance)
  L2:                   2 MiB (1 instance)
NUMA:
  NUMA node(s):         1
  NUMA node0 CPU(s):    0
Vulnerabilities:
  Gather data sampling: Not affected
  Itlb multihit:        KVM: Mitigation: VMX unsupported
  L1tf:                 Mitigation; PTE Inversion
  Mds:                  Vulnerable: Clear CPU buffers attempted, no microcode; SMT disabled
  Meltdown:             Mitigation; PTI
  Mmio stale data:      Unknown: No mitigations
  Retbleed:             Not affected
  Spec rstack overflow: Not affected
  Spec store bypass:    Vulnerable
  Spectre v1:           Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:           Mitigation; Retpolines, STIBP disabled, RSB filling, PBRSB-eIBRS Not affected
  Srbds:                Not affected
  Tsx async abort:      Not affected
```

stweil:
Could you please try `#pragma omp simd reduction` and `#pragma omp simd reduction(+:total)`?

zdenop:
Do you mean this?
```diff
--- a/src/arch/intsimdmatrix.cpp
+++ b/src/arch/intsimdmatrix.cpp
@@ -91,6 +91,7 @@ void IntSimdMatrix::MatrixDotVector(const GENERIC_2D_ARRAY<int8_t> &w,
     int total1 = 0;
     int total2 = 0;
     int total3 = 0;
+#pragma omp simd reduction(+ : total0, total1, total2, total3)
     for (int j = 0; j < num_in; ++j) {
       total0 += wi0[j] * u[j];
       total1 += wi1[j] * u[j];
diff --git a/test b/test
index 27618999..3ea10996 160000
```

Is it slightly slower by 0.2 s.

PR #4225, opened 2024-04-25
A few refactors in some files

sadra-barikbin:
Hi there!

A few tiny refactors in some files.

### DIFF
--- a/src/textord/tordmain.cpp
+++ b/src/textord/tordmain.cpp
@@ -543,7 +543,7 @@ bool Textord::clean_noise_from_row( // remove empties
             norm_count,
             rejected? "REJECTED": "ACCEPTED");
   }
-  return rejected;
+  return super_norm_count < textord_noise_sncount && rejected;
 }
 
 /**********************************************************************
--- a/src/textord/tordmain.cpp
+++ b/src/textord/tordmain.cpp
@@ -533,9 +533,9 @@ bool Textord::clean_noise_from_row( // remove empties
       }
     }
   }
+  // TODO: check whether `&& super_norm_count < textord_noise_sncount`should always be added here.
   bool rejected = dot_count > norm_count * textord_noise_normratio &&
-                  dot_count > 2 &&
-                  super_norm_count < textord_noise_sncount;
+                  dot_count > 2;
   if (textord_noise_debug) {
     tprintf("Row ending at (%d,%g):", blob_box.right(), row->base_line(blob_box.right()));
     tprintf(" R=%g, dc=%d, nc=%d, %s\n",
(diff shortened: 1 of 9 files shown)


src/textord/tordmain.cpp
```diff
@@ -533,16 +533,17 @@ bool Textord::clean_noise_from_row( // remove empties
       }
     }
   }
+  bool rejected = dot_count > norm_count * textord_noise_normratio && \
+                  dot_count > 2 && \
+                  super_norm_count < textord_noise_sncount;
```
stweil, line 538:
```suggestion
  bool rejected = dot_count > norm_count * textord_noise_normratio &&
                  dot_count > 2 &&
                  super_norm_count < textord_noise_sncount;
```

src/textord/tordmain.cpp
```diff
@@ -533,16 +533,17 @@ bool Textord::clean_noise_from_row( // remove empties
...
   if (textord_noise_debug) {
     tprintf("Row ending at (%d,%g):", blob_box.right(), row->base_line(blob_box.right()));
     tprintf(" R=%g, dc=%d, nc=%d, %s\n",
             norm_count > 0 ? static_cast<float>(dot_count) / norm_count : 9999, dot_count,
             norm_count,
-            dot_count > norm_count * textord_noise_normratio && dot_count > 2 ? "REJECTED"
-                                                                              : "ACCEPTED");
+            rejected? "REJECTED": "ACCEPTED");
```
stweil, line 544: The original code did not use `super_norm_count < textord_noise_sncount` here.

sadra-barikbin: But it includes the expression in the return statement.

sadra-barikbin:
```python
return super_norm_count < textord_noise_sncount &&
         dot_count > norm_count * textord_noise_rowratio && dot_count > 2;
```

stweil: Yes, the debug message and the return statement used different expressions in the past. I don't know whether that was intentional or a bug. The new code uses the same expression for both locations.

stweil (changes requested):
With my two suggested changes old and new code should be equivalent and we are safe. Maybe the original patch would be better, but it might change the output of the debug tprintf, and I am not sure about that.

egorpugin:
Just revert and dont touch that condition.

stweil:
Often it is easier if unrelated changes like in this pull request are split in separate pull requests, for example "Remove unused local variable", "Fix typo in comment", "Simplify code by using BlobGridSearch" and so on. Then those contributions are easy to review, and they can be merged quickly.

sadra-barikbin:
I made PRs #4226 , #4227 , #4228 and #4229 for the other changes.

stweil (approved):
Thank you.

PR #4229, opened 2024-04-25
Use `std::min` and `std::max` for min & max operations in `makerow.cpp::most_overlapping_row()`

sadra-barikbin:


egorpugin (approved):

stweil (approved):
Thank you.

PR #4234, opened 2024-05-03
Fix <Unicode> output and issues reported by Coverity Scan for PAGE XML renderer

stweil:


JKamlah (approved):
LGTM. Thanks @stweil

amitdo:
Next time, I suggest to mention the service name "Coverity Scan" also in the commit message/body.

stweil:
Up to now I used "CID" in the commit summary, so you can search for that string to find related commits.

Issue #4230, opened 2024-04-30
Infinite recursion for `tesseract --list-langs` with conda-forge binary


jonashaag:
### Current Behavior

`tesseract --list-langs` goes into infinite loop on macOS if `TESSDATA_PREFIX` is empty.

macOS Instruments shows infinite recursion in `addAvailableLanguages`, and a LOT of `stat64` calls (multiple 10k per second).

### Expected Behavior

Should not go into infinite recursion

### tesseract -v

tesseract 5.3.4
 leptonica-1.83.1
  libgif 5.2.1 : libjpeg 8d (libjpeg-turbo 3.0.0) : libpng 1.6.43 : libtiff 4.6.0 : zlib 1.2.13 : libwebp 1.3.2 : libopenjp2 2.5.2
 Found NEON

### Operating System

macOS 14 Sonoma

### uname -a

Darwin ... 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:12:41 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T8103 arm64

### Compiler

From conda-forge

### Virtualization / Containers

None

zdenop:
Can you please exactly describe how to reproduce the problem?
How did you install tesseract?

I do not have macOS, but on Windows, I got this result:
```sh
> set TESSDATA_PREFIX=""
> tesseract --list-langs
List of available languages in """/" (0):

>set TESSDATA_PREFIX=
>tesseract --list-langs
List of available languages in "./" (1):
Downloads/dotslayer
```

On the Linux I got this result:
```sh
$ export TESSDATA_PREFIX=""
$ tesseract --list-langs
List of available languages in "/usr/share/tesseract-ocr/5/tessdata/" (2):
eng
osd
```

jonashaag:
On macOS, your Linux commands reproduce the infinite loop

zdenop:
@jonashaag : how did you installed tesseract?

@stweil : Can you have a look at this?

stweil:
I don't get an infinite loop on macOS. An empty `TESSDATA_PREFIX` is handled like an unset one:
```
% tesseract --list-langs|head -3
List of available languages in "/opt/homebrew/share/tessdata/" (682):
Fraktur
Greek
% TESSDATA_PREFIX= tesseract --list-langs|head -3
List of available languages in "/opt/homebrew/share/tessdata/" (682):
Fraktur
Greek
```

[...]

PR #4239, opened 2024-05-12
Improve CCUtil::main_setup (fixes issue #4230)

stweil:
Conda installations patch TESSDATA_PREFIX in the binary. That does not work for std::string because the length won't be patched, so use a normal C string which can be patched.

Simplify also the code which checks the last character of datadir.

### DIFF
--- a/src/ccutil/ccutil.cpp
+++ b/src/ccutil/ccutil.cpp
@@ -46,7 +46,7 @@ CCUtil::~CCUtil() = default;
 void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
   imagebasename = basename; /**< name of image */
 
-  char *tessdata_prefix = getenv("TESSDATA_PREFIX");
+  const char *tessdata_prefix = getenv("TESSDATA_PREFIX");
 
   if (!argv0.empty()) {
     /* Use tessdata prefix from the command line. */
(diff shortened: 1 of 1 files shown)


src/ccutil/ccutil.cpp
```diff
@@ -77,17 +77,20 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
   if (datadir.empty()) {
 #if defined(TESSDATA_PREFIX)
     // Use tessdata prefix which was compiled in.
-    datadir = TESSDATA_PREFIX "/tessdata";
+    // Note that some software (for example conda) patch TESSDATA_PREFIX
```
stweil, line 80:
```suggestion
    // Note that some software (for example conda) patches TESSDATA_PREFIX
```

amitdo:
> Conda installations patch TESSDATA_PREFIX in the binary.

Do they really patch the **binary**???

Or maybe they simply patch the configure.ac or a cpp source file?

amitdo:
Do we really want to bend our code just because some distro is doing crazy things?

stweil:
They really patch all binaries which contain a lengthy magic string (which they configure as prefix). In most cases that allows an installation to a different prefix.

stweil:
Example: lib/pkgconfig/tesseract.pc is generated with this first line:
```
prefix=/Users/stweil/src/github/conda-forge/tesseract-feedstock/miniforge3/conda-bld/tesseract_1715530534651/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh
```

The installed file has this first line:
```
prefix=/opt/homebrew/Caskroom/miniconda/base
```

stweil:
> Do we really want to bend our code just because some distro is doing crazy things?

I think we would not bend our code in most cases, but in this case our workaround only requires a single code line with an explaining comment. That is less work for us than handling future issue reports if we don't adapt the Tesseract code.

egorpugin:
In general dont follow it.
We spaw to string, to fs::path? Then what?
It's their thing.

PR #4240, opened 2024-05-13
Fix setup of datadir on installations with Conda (issue #4230)

stweil:


src/ccutil/ccutil.cpp
```diff
@@ -77,11 +77,10 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
   if (datadir.empty()) {
 #if defined(TESSDATA_PREFIX)
     // Use tessdata prefix which was compiled in.
+    datadir = TESSDATA_PREFIX "/tessdata/";
     // Note that some software (for example conda) patches TESSDATA_PREFIX
-    // in the binary, so it should not be used directly with a std::string.
-    tessdata_prefix = TESSDATA_PREFIX;
-    datadir = tessdata_prefix;
-    datadir += "/tessdata/";
+    // in the binary, so it might be shorter. Recalculate its length.
```
danpla, line 82: Since std::string is initialized from "const char*", I wonder how it can be shorter?

stweil: The length is calculated at compile time and not updated at runtime if the `const char *` was patched.

stweil: If you have a locally built `tesseract` installed in /usr/local/bin (= with prefix /usr/local), you can patch the binary for example using `perl -pi -e 's,/usr/l,/tmp/\0,' /usr/local/bin/tesseract`. Then try `tesseract --list-langs` to get the endless recursion.

danpla: Ok, thanks.

danpla: As I understand, if the patched path is longer than the original one, the data in std::string is truncated. If so, it probably makes sense to call strlen() on the string literal rather than on c_str().

stweil: It's not possible to replace the original TESSDATA_PREFIX by a longer one because that would destroy other data. Therefore Conda uses a very long placeholder for TESSDATA_PREFIX.

PR #4244, opened 2024-05-19
Use AM_CPPFLAGS also for compilation of all sources

stweil:
It was not used for all sources before. Therefore some parts of the code (especially the code for training) used different compiler options. For example NDEBUG was not defined, and so the training code was built with debug assertions (resulting in potentially slower execution).


stweil:
This change also reduces the size of the generated binaries.

PR #4245, opened 2024-05-19
Fix some compiler warnings

stweil:


src/dict/dawg.h
```diff
@@ -74,12 +74,12 @@ enum DawgType {
...
-#define MAX_NODE_EDGES_DISPLAY (int64_t)100
-#define MARKER_FLAG (int64_t)1
-#define DIRECTION_FLAG (int64_t)2
-#define WERD_END_FLAG (int64_t)4
+#define FORWARD_EDGE static_cast<int32_t>(0)
+#define BACKWARD_EDGE static_cast<int32_t>(1)
+#define MAX_NODE_EDGES_DISPLAY static_cast<int64_t>(100)
+#define MARKER_FLAG static_cast<int64_t>(1)
+#define DIRECTION_FLAG static_cast<int64_t>(2)
+#define WERD_END_FLAG static_cast<int64_t>(4)
```
egorpugin, line 82: Is it possible to replace with inline vars? constexpr inlines?

egorpugin (approved):

src/dict/dawg.h
stweil: Yes, I think so (it the compiler produces the same efficient code). We could consider such a change for more macros in the future.

PR #4250, opened 2024-05-24
Replace strcpy and strncpy by new inline helper function

stweil:


egorpugin:
We should name it `copy_string_remove_after_transition_to_std_strings()`.

src/ccmain/ltrresultiterator.cpp
```diff
@@ -310,11 +308,7 @@ char *LTRResultIterator::WordTruthUTF8Text() const {
   if (!HasTruthString()) {
```
egorpugin, line 308: What if we change this function return type to std::string?
```
char *LTRResultIterator::WordTruthUTF8Text() const {
```
to
```
std::string LTRResultIterator::WordTruthUTF8Text() const {
```
and also all calling sites?

stweil:
> We should name it `copy_string__remove_after_transition_to_std_strings()`.

That's a little bit long, but a comment which basically says the same thing could be added in helpers.h.

egorpugin (approved):

src/ccutil/helpers.h
```diff
@@ -35,6 +35,15 @@
 
 namespace tesseract {
 
+// Copy a std::string to a newly allocated char *.
```
stweil, line 38:
```suggestion
// Copy a std::string to a newly allocated char *.
// TODO: Remove this function once the related code has been converted to use std::string.
```

PR #4272, opened 2024-06-21
Make regular usage of CMAKE_INSTALL_LIBDIR and GNUInstallDirs

Zopolis4:
Use CMake's logic to define `CMAKE_INSTALL_LIBDIR` rather than relying on a user override, and make use of the additional abstraction that the `GNUInstallDirs` module allows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -471,9 +471,7 @@ else()
   set(DOTPRODUCT_FLAGS "${DOTPRODUCT_FLAGS} -O3 -ffast-math")
 endif()
 
-if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
-  set(CMAKE_INSTALL_LIBDIR lib)
-endif(NOT DEFINED CMAKE_INSTALL_LIBDIR)
+include (GNUInstallDirs)
 
 set(AUTOCONFIG_SRC ${CMAKE_CURRENT_BINARY_DIR}/config_auto.h.in)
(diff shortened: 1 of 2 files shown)


zdenop:
What is a benefit of this? Especially for non GNU installation like Windows?

Zopolis4:
This doesn't affect a windows installation, as it only modifies unix installation code.

zdenop:
But what it the benefit of this PR? What kind of problem are you solving?

Zopolis4:
Avoiding the need to manually define `CMAKE_INSTALL_LIBDIR` as `lib64` on systems where this is the default value.

egorpugin:
Maybe change the default to lib64?
What are systems with `lib` today?

stweil:
> Maybe change the default to lib64?
> What are systems with `lib` today?

Linux installations for i686, older Raspberry PI distributions, other 32 bit embedded systems don't have `lib64` as far as I know.

Zopolis4:
Not all distros use `lib64`-- using `GNUInstallDirs` ensures that tesseract will match whatever is the default for that platform.

zdenop:
I checked openSUSE and it use booth (lib & lib64)

zdenop:
I make basic build&install test on Windows and openSUSE and I do not see problem with this PR.
+ leptonica using GNUInstallDirs already, so I do not expect problems.

zdenop (approved):

egorpugin (approved):

PR #4275, opened 2024-06-27
Run LSTM recognition in multiple threads

jkarthic:
Init time option **lstm_num_threads** should be used to set the number of LSTM threads. This will ensure that word recognition can run independently in multiple threads, thus effectively utilizing multi-core processors.

Following are my test results for a sample screenshot.
CPU : Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
OS : WIndows
Compiler : MSVC 19.38.33130.0 (Installed from Visual Studio 2022)
Model: eng.traineddata from tessfast
PSM: 6

Total time taken for `Recognize` API call, Built without OpenMP
With lstm_num_threads=1, total time taken = 3.95 seconds
**With lstm_num_threads=4, total time taken = 1.4 seconds**

On the other hand, here are the numbers with OpenMP
OMP_THREAD_LIMIT not set, total time taken = 3.59 seconds
OMP_THREAD_LIMIT=4, total time taken = 3.57 seconds
OMP_THREAD_LIMIT=1, total time taken = 4.19 seconds

As we can observe, this branch with `lstm_num_threads` set as 4, performs way better than the openmp multithreading supported currently. Setting `lstm_num_threads` equal to the number of cores in the processor will give the best performance.

### DIFF
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -26,6 +26,7 @@
 #include <cstdint> // for int16_t, int32_t
 #include <cstdio>  // for fclose, fopen, FILE
 #include <ctime>   // for clock
+#include <future>
 #include "control.h"
 #ifndef DISABLED_LEGACY_ENGINE
 #  include "docqual.h"
@@ -194,36 +195,42 @@ void Tesseract::SetupWordPassN(int pass_n, WordData *word) {
   }
 }
 
-// Runs word recognition on all the words.
-bool Tesseract::RecogAllWordsPassN(int pass_n, ETEXT_DESC *monitor, PAGE_RES_IT *pr_it,
-                                   std::vector<WordData> *words) {
-  // TODO(rays) Before this loop can be parallelized (it would yield a massive
-  // speed-up) all remaining member globals need to be converted to local/heap
-  // (eg set_pass1 and set_pass2) and an intermediate adaption pass needs to be
-  // added. The results will be significantly different with adaption on, and
-  // deterioration will need investigation.
-  pr_it->restart_page();
-  for (unsigned w = 0; w < words->size(); ++w) {
-    WordData *word = &(*words)[w];
-    if (w > 0) {
-      word->prev_word = &(*words)[w - 1];
+bool Tesseract::RecogWordsSegment(std::vector<WordData>::iterator start,
+                                  std::vector<WordData>::iterator end,
+                                  int pass_n,
+                                  ETEXT_DESC *monitor,
+                                  PAGE_RES *page_res,
+                                  LSTMRecognizer *lstm_recognizer,
+                                  std::atomic<int>& words_done,
+                                  int total_words,
+                                  std::mutex& monitor_mutex) {
+  PAGE_RES_IT pr_it(page_res);
+  // Process a segment of the words vector
+  pr_it.restart_page();
+
+  for (auto it = start; it != end; ++it, ++words_done) {
+    WordData *word = &(*it);
+    if (it != start) {
+      word->prev_word = &(*(it - 1));
     }
     if (monitor != nullptr) {
+      std::lock_guard<std::mutex> lock(monitor_mutex);
       monitor->ocr_alive = true;
       if (pass_n == 1) {
-        monitor->progress = 70 * w / words->size();
+        monitor->progress = 70 * words_done / total_words;
       } else {
-        monitor->progress = 70 + 30 * w / words->size();
+        monitor->progress = 70 + 30 * words_done / total_words;
       }
+      // Only call the progress callback for the first thread.
       if (monitor->progress_callback2 != nullptr) {
-        TBOX box = pr_it->word()->word->bounding_box();
+        TBOX box = pr_it.word()->word->bounding_box();
         (*monitor->progress_callback2)(monitor, box.left(), box.right(), box.top(), box.bottom());
       }
       if (monitor->deadline_exceeded() ||
-          (monitor->cancel != nullptr && (*monitor->cancel)(monitor->cancel_this, words->size()))) {
+          (monitor->cancel != nullptr && (*monitor->cancel)(monitor->cancel_this, total_words))) {
         // Timeout. Fake out the rest of the words.
-        for (; w < words->size(); ++w) {
-          (*words)[w].word->SetupFake(unicharset);
+        for (; it != end; ++it) {
+          it->word->SetupFake(unicharset);
         }
         return false;
       }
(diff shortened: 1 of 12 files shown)


src/ccmain/tesseractclass.cpp
```diff
@@ -477,7 +481,10 @@ Tesseract::~Tesseract() {
   for (auto *lang : sub_langs_) {
     delete lang;
   }
-  delete lstm_recognizer_;
+  for (int i = 0; i < lstm_recognizers_.size(); ++i) {
```
stweil, line 484: `i` should use the same data type as the return value of `size()`.

egorpugin: `for (auto &&r : lstm_recognizers_) delete r;`

jkarthic: Thanks for your review comments. I have addressed this now.

src/ccmain/tessedit.cpp
```diff
@@ -169,8 +169,11 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
       tessedit_ocr_engine_mode == OEM_TESSERACT_LSTM_COMBINED) {
 #endif // ndef DISABLED_LEGACY_ENGINE
     if (mgr->IsComponentAvailable(TESSDATA_LSTM)) {
-      lstm_recognizer_ = new LSTMRecognizer(language_data_path_prefix.c_str());
-      ASSERT_HOST(lstm_recognizer_->Load(this->params(), lstm_use_matrix ? language : "", mgr));
+      for (int i = 0; i < lstm_num_threads; ++i) {
+        lstm_recognizers_.push_back(new LSTMRecognizer(language_data_path_prefix.c_str()));
```
egorpugin, line 173: Fine for me.
But maybe it is possible to switch to unique_ptrs here?

egorpugin (approved):

src/ccmain/control.cpp
```diff
@@ -238,31 +245,69 @@ bool Tesseract::RecogAllWordsPassN(int pass_n, ETEXT_DESC *monitor, PAGE_RES_IT
...
+  int segment_size = total_words / lstm_num_threads;
+  std::atomic<int> words_done(0);
+  std::mutex monitor_mutex;
+  std::vector<std::future<bool>> futures;
+
+  // Launch multiple threads to recognize the words in parallel
+  auto segment_start = words->begin() + segment_size;
+  for (int i = 1; i < lstm_num_threads; ++i) {
+      auto segment_end = (i == lstm_num_threads - 1) ? words->end() : segment_start + segment_size;
+      futures.push_back(std::async(std::launch::async, &Tesseract::RecogWordsSegment,
```
egorpugin, line 286: What about static thread pool here?

stweil:
Many thanks for this nice contribution.

With this pull request users have the choice of using the new argument `--lstm-num-threads N` or setting the new parameter with `-c lstm-num-threads=N`. Do we need both ways? If a command line argument is desired (like in the case of `--dpi`), I think that there might be more user friendly variants. Although `--lstm-num-thread` describes the technical implementation correctly, it is a lengthy argument which maybe requires too much explanation. Do we expect more `--xxx-num-thread` arguments in the future? Or would `--threads` be sufficient?

Maybe we could also extend the command line syntax to have `--PARAMETER VALUE` as an alternative for `-c PARAMETER=VALUE` for any Tesseract parameter.

stweil:
> Setting `lstm_num_threads` equal to the number of cores in the processor will give the best performance.

Just to clarify this statement: it's only true for the OCR of a single page. For mass production it is still better to run (number of cores) parallel Tesseract processes because then all processing steps use 100 % of the available resources.

src/ccmain/tessedit.cpp
jkarthic: I was just following the existing design pattern of tesseract of using `LSTMRecognizer *` and performing `new` and `delete` for allocation and deallocation.
As per the C++ best practice, we should ideally use shared_ptrs everywhere for accessing LSTMRecognizer, to avoid any dangling ptrs. But that will involve a lot of refactoring, deviating away from the main purpose of this PR. Hence I just followed the existing tesseract practice of using `LSTMRecognizer *`.

src/ccmain/control.cpp
jkarthic: Good question.
As `stweil` has clarified in his latest comment, this multiple lstm threads is not meant for mass production. This feature is probably meant for consumer-end devices running a single page OCR once in a while, but with least possible latency. In such cases thread creation/deletion is not a big overhead. But if we really come across use-cases where thread creation/deletion overhead becomes significant, we could look at replacing this with a thread pool at that point.

stweil: The default case (single thread) must not create a new thread.

jkarthic:
> Many thanks for this nice contribution.

And many thanks to you for reviewing this patiently.
> 
> With this pull request users have the choice of using the new argument `--lstm-num-threads N` or setting the new parameter with `-c lstm-num-threads=N`. Do we need both ways?

This `lstm_num_threads` is a init time parameter. The `LSTMRecognizer ` instances are created during init. Setting this new parameter with  `-c lstm-num-threads=N` will not work, as it is setting the variable after the init is done. 

> Although `--lstm-num-thread` describes the technical implementation correctly, it is a lengthy argument which maybe requires too much explanation. Do we expect more `--xxx-num-thread` arguments in the future? Or would `--threads` be sufficient?

When I tested tesseract with a psm of 3(which is the default for `tesseract.exe`), page segmentation was taking significantly more time than the actual LSTM recognition. For example, in one of my tests, page segmentation was taking ~7 seconds, and lstm was taking ~3 seconds, taking the total to ~10 seconds. Users running with default psm parameter should not expect that the entire 10 seconds will be run in multiple threads. In this case, the major part of ~7 seconds will run single threaded and only a minor part of ~3 seconds will be multi threaded. Hence I thought adding a longer name is setting the user expectation right, that only a portion of tesseract will be running multithreaded.
Also there are other numthreads variables related to OpenMP, inside the code which were named generically such as `kNumThreads`, `__num_threads` and `num_threads`. Naming this as `lstm_num_threads` also differentiates this as a seperate variable, not to be confused with OpenMP num threads.

jkarthic:
> > Setting `lstm_num_threads` equal to the number of cores in the processor will give the best performance.
> 
> Just to clarify this statement: it's only true for the OCR of a single page. For mass production it is still better to run (number of cores) parallel Tesseract processes because then all processing steps use 100 % of the available resources.

Totally agreed. This is meant for latency-sensitive real-time applications, with ocr probably running in the consumer's device itself.

src/ccmain/control.cpp
jkarthic: yes, the default case (lstm_num_threads == 1) does not create a new thread. As you see, the loop starts with `i = 1`, and hence does not execute when `lstm_num_threads == 1`.

jkarthic:
@stweil 
I observed a crash issue in the earlier code due to WERD_RES objects freed by one thread was used by another thread for iterating thru the WERD_RES singly linked list.
To fix the above above issue, I have modified WERD_RES linked list to use shared pointer instead of raw pointers, so that lifetime of the objects are managed automatically. I have also added mutex protections around the PAGE_RES_IT functions that modify this list in order to avoid race conditions.
Please take a look at the modifications whenever you get some time for this.

egorpugin (changes requested):
Now it is much much worse.

Remove std::async.

egorpugin:
I suggest to use previous version as base.

jkarthic:
> Now it is much much worse.

@egorpugin I am not sure, if I understand your comment here. Could you please elaborate what is "much much worse"?

egorpugin:
> @egorpugin I am not sure, if I understand your comment here. Could you please elaborate what is "much much worse"?

* More complex code.
* A lot of sync.
* Much harder to review.

You need to provide a very detailed description of:
1) algorithm. How it works? Is it possible to sync less?
2) changes in files. I see new types, mutex locks in some existing functions. See example how this can be described from gcc commit messages, e.g. https://github.com/gcc-mirror/gcc/commit/5185274c76cc3b68a38713273779ec29ae4fe5d2 (bottom part of the commit message)

LLotme (commented):
gh pr checkout 4275

LLotme (commented):
gh pr checkout 4275

burningfireplace:
I tested this PR on my Mac (M1 chip). I have a few observations to share:

1. The latest commit (d1eed6a) does not compile successfully on my system. I encountered multiple errors related to ELIST_ITERATOR_T, BLOBNBOX_IT, and other classes. It seems some changes may be missing or incomplete in this commit.

2. The initial commit (6a2e239) does compile and produces a working binary. Using the new `--lstm-num-threads` option, I was able to achieve a ~2x speedup (compared to 5.4.1 + OpenMP) by finding the optimal N value for my system.

3. However, I noticed an issue with language/script recognition when using multiple languages. For example:
   - Using `-l "eng+ukr"` tries to recognize Cyrillic chars as Latin chars
   - Using `-l "ukr+eng"` vice-versa: incorrectly recognizes Cyrillic in places with Latin chars
   
This behavior differs from Tesseract 5.4.1, which produces correct output for the same inputs in both cases.

I hope this feedback is helpful.

jkarthic:
@burningfireplace 

Thanks for trying it out and providing a detailed analysis. Here is my reply.
1. I was using windows while developing this. Never tested this on Mac, as mac had apple's inbuilt OCR Vision API, which in my experience performs better than tesseract. But that's no reason for breaking the build on Mac. I will try to fix the compile issues on Mac, when I get some time.
2. Nice to know, this is improving speed on Mac as well.
3. The initial commit had a few race condition bugs that might be causing this. Once compile issues are resolved, I would encourage you to try the latest commit once again. Ideally the latest commit should provide identical output as the single threaded version.

Issue #4277, opened 2024-07-03
After update with Homebrew, version number is included in TESSDATA_PREFIX


freebrowser1:
### Current Behavior

After an update, tesseract cannot find the language files anymore, because the path where TESSDATA_PREFIX changes after every update, so I have to change TESSDATA_PREFIX every time.
Now it is `/opt/homebrew/Cellar/tesseract/5.4.1` and tomorrow `/opt/homebrew/Cellar/tesseract/5.4.2`which is very annoying.

- 

### Expected Behavior

The TESSDATA_PREFIX environmental variable should not contain a version number.

### Suggested Fix

No version number in the TESSDATA_PREFIX path.

### tesseract -v

tesseract 5.4.1
 leptonica-1.84.1
  libgif 5.2.1 : libjpeg 8d (libjpeg-turbo 3.0.0) : libpng 1.6.43 : libtiff 4.6.0 : zlib 1.2.12 : libwebp 1.4.0 : libopenjp2 2.5.2
 Found NEON
 Found libarchive 3.7.4 zlib/1.2.12 liblzma/5.4.6 bz2lib/1.0.8 liblz4/1.9.4 libzstd/1.5.6
 Found libcurl/8.6.0 SecureTransport (LibreSSL/3.3.6) zlib/1.2.12 nghttp2/1.61.0

### Other Operating System

macOS Sonoma 14.5.

### uname -a

Darwin armemac.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64

### Compiler

Pre-compiled by homebrew.

### CPU

Apple Silicon M1 (arm64) but on another Macbook which is an Intel i5 with the same OS, Homebrew  and the same tesseract version the issue also occurs.

### Virtualization / Containers

N/A

stweil:
Why do you set TESSDATA_PREFIX at all? Normally it should not be necessary for users to set it.

And why do you think this is an issue for the tesseract repository although you are using Homebrew?

Please use the [Tesseract user forum](https://groups.google.com/d/forum/tesseract-ocr) for questions.

stweil:
The next release of Tesseract will ignore a TESSDATA_PREFIX which points to a non-existing location. Then running `tesseract` with your Homebrew configution would print a warning, but work otherwise.

freebrowser1:
[...]

PR #4278, opened 2024-07-03
Ignore illegal TESSDATA_PREFIX (not existing filesystem entry, issue #4277)

stweil:
Don't check for a directory, because a symbolic link is also allowed.

### DIFF
--- a/src/ccutil/ccutil.cpp
+++ b/src/ccutil/ccutil.cpp
@@ -17,7 +17,8 @@
 #include "ccutil.h"
 
 #include <cstdlib>
-#include <cstring> // for std::strrchr
+#include <cstring>    // for std::strrchrA
+#include <filesystem> // for std::filesystem
 
 namespace tesseract {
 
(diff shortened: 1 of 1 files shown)


src/ccutil/ccutil.cpp
```diff
@@ -48,6 +49,12 @@ void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
 
   const char *tessdata_prefix = getenv("TESSDATA_PREFIX");
 
+  // Ignore TESSDATA_PREFIX if there is no matching filesystem entry.
+  if (tessdata_prefix != nullptr && ! std::filesystem::exists(tessdata_prefix)) {
```
egorpugin, line 53: If this is not a space typo `! ` it is easier to use fs as
```
namespace fs = std::filesystem;
if (!fs::exists(...))
```

stweil: It was a space typo which is fixed now, thanks for the hint.

egorpugin (approved):

PR #4283, opened 2024-07-12
Fix confidence output for the PAGE XML renderer

JKamlah:
Hi there,
i have a small fix to contribute regarding the PAGE-XML renderer.

Problem: Block- and line-level confidence in the PAGE XML renderer always returned 0.
Fix: Re-arrangement of variable initilization for block- and line-level confidence.


stweil (approved):
Thanks.

PR #4303, opened 2024-08-22
Reduce clock syscalls

heshpdx:
I noticed significant calls to `clock_gettime` so I investigated and found that we could eliminate them when they are not in use.  This was found from the continuing scrutiny of SPEC CPUv8 benchmark development. @zdenop 

Gate the sampling of the clock by the `tessedit_timing_debug` flag, which is the only time it gets used anyway. This eliminates unnecessary clock_gettime() system calls.

### DIFF
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -1312,7 +1312,10 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   PointerVector<WERD_RES> best_words;
   // Points to the best result. May be word or in lang_words.
   const WERD_RES *word = word_data->word;
-  clock_t start_t = clock();
+  clock_t start_t;
+  if (tessedit_timing_debug) {
+    start_t = clock();
+  }
   const bool debug = classify_debug_level > 0 || multilang_debug_level > 0;
   if (debug) {
     tprintf("%s word with lang %s at:", word->done ? "Already done" : "Processing",
@@ -1364,8 +1367,8 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   } else {
     tprintf("no best words!!\n");
   }
-  clock_t ocr_t = clock();
   if (tessedit_timing_debug) {
+    clock_t ocr_t = clock();
     tprintf("%s (ocr took %.2f sec)\n", word_data->word->best_choice->unichar_string().c_str(),
             static_cast<double>(ocr_t - start_t) / CLOCKS_PER_SEC);
   }
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -1312,7 +1312,7 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   PointerVector<WERD_RES> best_words;
   // Points to the best result. May be word or in lang_words.
   const WERD_RES *word = word_data->word;
-  clock_t start_t;
+  clock_t start_t = 0;
   if (tessedit_timing_debug) {
     start_t = clock();
   }


src/ccmain/control.cpp
```diff
@@ -1364,8 +1367,8 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   } else {
     tprintf("no best words!!\n");
   }
-  clock_t ocr_t = clock();
   if (tessedit_timing_debug) {
+    clock_t ocr_t = clock();
```
stweil, line 1371: This part of the commit is a good change.

stweil (commented):
I am not sure whether it would be better to apply only the 2nd of the two modifications.

src/ccmain/control.cpp
```diff
@@ -1312,7 +1312,10 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   PointerVector<WERD_RES> best_words;
   // Points to the best result. May be word or in lang_words.
   const WERD_RES *word = word_data->word;
-  clock_t start_t = clock();
+  clock_t start_t;
+  if (tessedit_timing_debug) {
+    start_t = clock();
+  }
```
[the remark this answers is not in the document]
heshpdx: The call to `clock()` results in a vdso syscall to the operating system. If you are running tesseract in batch mode with multiple copies running on the same machine, this seemingly innocuous call could result in a storm of useless syscalls to the OS. It is definitely more expensive than a single conditional that all branch predictors would get correct.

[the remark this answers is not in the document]
heshpdx: Regardless of whether one believes this is slower or faster, the `start_t` variable has no consumers unless `tessedit_timing_debug` is enabled.

zdenop:
@stweil: what about switching to c++11 `std::chrono::high_resolution_clock::now()` instead of `std::clock()` (see e.g. https://stackoverflow.com/questions/28396014/why-is-clock-considered-bad)?
It should provide higher precision and the "same" result on different platforms comparing to `std::clock()`

src/ccmain/control.cpp
```diff
@@ -1364,8 +1367,8 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   } else {
     tprintf("no best words!!\n");
   }
-  clock_t ocr_t = clock();
   if (tessedit_timing_debug) {
+    clock_t ocr_t = clock();
     tprintf("%s (ocr took %.2f sec)\n", word_data->word->best_choice->unichar_string().c_str(),
             static_cast<double>(ocr_t - start_t) / CLOCKS_PER_SEC);
```
stweil, line 1373: The bad news is that g++ is not clever enough and produces a warning with this PR:
```
src/ccmain/control.cpp:1373:39: warning: 'start_t' may be used uninitialized [-Wmaybe-uninitialized]
```

egorpugin: What g++ version?

egorpugin (approved):

egorpugin:
Apply for now and postpone to timing/debugging related refactoring.

src/ccmain/control.cpp
heshpdx: Sorry about that. I will initialize it.

src/ccmain/control.cpp
```diff
@@ -1312,7 +1312,10 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   PointerVector<WERD_RES> best_words;
   // Points to the best result. May be word or in lang_words.
   const WERD_RES *word = word_data->word;
-  clock_t start_t = clock();
+  clock_t start_t;
```
heshpdx, line 1315:
```suggestion
  clock_t start_t = 0;
```

src/ccmain/control.cpp
egorpugin: But init is exactly wasted cycles again?

egorpugin: You need to use `[[maybe_unused]]` attribute here.

zdenop (approved):

src/ccmain/control.cpp
stweil: > What g++ version?

g++ (Debian 12.2.0-14) 12.2.0

egorpugin: Interesting to check it on gcc-14

stweil: _Apple clang version 15.0.0 (clang-1500.3.9.4)_ and _g++-14 (Homebrew GCC 14.1.0_2) 14.1.0_ show the same warning.

stweil:
I have run a short test to see the effect of less calls of `clock()`. 10 million calls take 2.8 s on my MacBook or 4.5 s on a Linux server. So the effect of this PR is very small. It saves two calls per word or less than a second per one million words.

heshpdx:
The effects will be more egregious when running in batch mode, e.g. with 192 instances executing simultaneously on a modern many-core server.

Issue #4304, opened 2024-08-23
Baselines Incorrect Using Certain PSMs (`PageIterator::Baseline`) 


Balearica:
## Overview
When using certain PSMs with certain inputs, the `PageIterator::Baseline` function produces results that are incorrect due to a bug when getting line bounding boxes.  I noticed this when using `psm` `8` (single word).  This impacts API users trying to get a line's baseline, and also causes incorrect results in CLI output formats that report baseline (`.hocr`).

## Reproducible Example
While this is most noticeable using `it->Baseline` through the API, the phenomenon can be demonstrated using the CLI with the example image below.

![simple_c2](https://github.com/user-attachments/assets/b3b977d7-1117-4ec1-b73c-6c14fc907dd6)

The word in the image is recognized correctly--including having the same bounding box--whether `psm` is set to `6` (single block) or `8` (single word).  However, the latter does not calculate the baseline correctly.

When setting `psm` to `6`, the baseline attribute is set to `-0.036 0`, which is correct.

![simple_c2_line](https://github.com/user-attachments/assets/f1057d6a-00e9-4d5f-b8ba-3de60d5c1784)

```sh
tesseract simple_c2.png stdout --oem 0 --psm 6 hocr
```
```html
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
    "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title></title>
  <meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
  <meta name='ocr-system' content='tesseract 5.1.0-471-gbc490' />
  <meta name='ocr-capabilities' content='ocr_page ocr_carea ocr_par ocr_line ocrx_word ocrp_wconf'/>
 </head>
 <body>
  <div class='ocr_page' id='page_1' title='image "simple_c2.png"; bbox 0 0 328 194; ppageno 0; scan_res 96 96'>
   <div class='ocr_carea' id='block_1_1' title="bbox 85 84 195 104">
    <p class='ocr_par' id='par_1_1' lang='eng' title="bbox 85 84 195 104">
[...]

PR #4305, opened 2024-08-23
Calculate row bounding box in single-word mode per #4304

Balearica:
See #4304 for full context.  

In short, `ROW` objects are expected to have valid/accurate bounding boxes, which are accessed using `row->bounding_box()`.  However, bounding box values are not calculated automatically--the method `row_obj->recalc_bounding_box()` must be run after the row is created and populated to calculate the bounding box.  This happens correctly in 5 of the 6 instances where `new ROW` is called, however does not occur in 1 case (within the `make_single_word` function that runs when `psm` is `SINGLE_WORD` or `SINGLE_CHAR`).  

This PR fixes this by adding the required call to `recalc_bounding_box` within `make_single_word`.  This brings that function in line with all the other functions that create `ROW` objects.  

This change fixes the bug described in #4304, where baselines are not correctly calculated when using certain `psm` settings.  After implementing this fix, the motivating example in that issue is resolved, and the baseline is calculated correctly.

### DIFF
--- a/src/textord/wordseg.cpp
+++ b/src/textord/wordseg.cpp
@@ -87,6 +87,7 @@ void make_single_word(bool one_blob, TO_ROW_LIST *rows, ROW_LIST *real_rows) {
     word->set_flag(W_EOL, true);
     word->set_flag(W_DONT_CHOP, one_blob);
     word_it.add_after_then_move(word);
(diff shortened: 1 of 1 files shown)


stweil:
@Balearica, thanks for this fix. Please use your full name as commit author in future pull requests.

PR #4306, opened 2024-08-25
Update code for tprintf

stweil:
- Remove unused macro and update comment
- Modernize and simplify the code

In a next step a `std::ofstream` which can be used instead of `tprintf` will be added.

### DIFF
--- a/src/ccutil/tprintf.cpp
+++ b/src/ccutil/tprintf.cpp
@@ -35,38 +35,45 @@ INT_VAR(log_level, INT_MAX, "Logging level");
 
 static STRING_VAR(debug_file, "", "File to send tprintf output to");
 
-// Trace printf
-void tprintf(const char *format, ...) {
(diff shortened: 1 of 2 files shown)


src/ccutil/tprintf.cpp
```diff
@@ -31,44 +31,49 @@
...
+// The output is written to stderr if debug_file is empty.
+// Otherwise it is written to debug_file.
+// It is possible to switch between stderr and debug_file output:
+// tprintf("write to configured output\n");
+// debug_file = "";
+// tprintf("write to stderr\n");
+// debug_file = "/tmp/log";
+// tprintf("write to /tmp/log\n");
+// debug_file = "";
+// tprintf("write to stderr\n");
```
stweil, line 51: I wonder whether such use cases exist. We could simplify the code further if we only allow setting the desired output once before the first use of `tprintf`.

PR #4309, opened 2024-08-26
Fix performance and other issues reported by Codacy

stweil:
- replace std::string::substr by std::string::resize
- initialize variables in initialization list
- fix several format strings


egorpugin (approved):

egorpugin:
I think it is possible to use in-class initializers instead of repeating vars in ctor.

```
struct x {
	int *ptr{};
};
```

Instead of `x::x() : ptr{nullptr} {}`

egorpugin:
Also it is 2025 soon, I would propose to allow some c++20/23 features.

What are tesseract compiler requirements?

Support ubuntu20.04 lts gcc?

PR #4313, opened 2024-09-03
Remove unnecessary assignment and assertions

stweil:


src/ccmain/ltrresultiterator.cpp
```diff
@@ -124,19 +121,20 @@ float LTRResultIterator::Confidence(PageIteratorLevel level) const {
...
         ++certainty_count;
         res_it.forward();
       } while (res_it.row() == res_it.prev_row());
       break;
     case RIL_WORD:
-      mean_certainty += best_choice->certainty();
-      ++certainty_count;
+      best_choice = res_it.word()->best_choice;
+      mean_certainty = best_choice->certainty();
+      certainty_count = 1;
```
egorpugin, line 132: Doesn't change behavor?

`++certainty_count;` vs `certainty_count = 1;`

egorpugin (approved):

src/ccmain/ltrresultiterator.cpp
stweil: Initial value 0 and a single increment gives value 1. There is no loop around it.

PR #4314, opened 2024-09-03
Add C++ stream for log messages and use it in two debug messages

stweil:
The two debug message print execution times, previously in units of 0.01 seconds, now in milliseconds.

In function `ErrorCounter::ComputeErrorRate` the new code normally also avoids a call of `clock()` which is only needed for the debug message.

The new C++ stream `tesserr` can be used as an alternative to `tprintf` which is type safe and avoids lots of type casts.

### DIFF
--- a/src/ccmain/control.cpp
+++ b/src/ccmain/control.cpp
@@ -41,6 +41,7 @@
 #endif
 #include "sorthelper.h"
 #include "tesseractclass.h"
+#include "tesserrstream.h"  // for tesserr
 #include "tessvars.h"
 #include "werdit.h"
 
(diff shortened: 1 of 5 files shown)


stweil:
The sw builds failed. This is hopefully now fixed by adding `TESS_API` for `tesserr`. I wonder why autoconf and cmake builds work fine without it.

egorpugin (approved):

src/ccmain/control.cpp
```diff
@@ -1368,10 +1370,10 @@ void Tesseract::classify_word_and_language(int pass_n, PAGE_RES_IT *pr_it, WordD
   } else {
     tprintf("no best words!!\n");
   }
-  if (tessedit_timing_debug) {
-    clock_t ocr_t = clock();
-    tprintf("%s (ocr took %.2f sec)\n", word_data->word->best_choice->unichar_string().c_str(),
-            static_cast<double>(ocr_t - start_t) / CLOCKS_PER_SEC);
+  if (timing_debug) {
+    total_time = clock() - total_time;
```
zdenop, line 1374: Why now to use c++11 `chrono::steady_clock::now()` (https://en.cppreference.com/w/cpp/chrono/steady_clock - *most suitable for measuring intervals*) instead of `std::clock()`? 
AFAIK it should be more platform consistent than `std::clock()'....

stweil: We can use it any time, but in time critical code we'd have to check that it does not cost more resources than `clock()`.

PR #4319, opened 2024-09-18
cmake: Correctly set the soversion based on SemVer properties

Conan-Kudo:
As this project follows Semantic Versioning, the shared object version should match these semantics.

The two options that make sense here are to have the soversion set to the version major (so only breaking changes are tracked) or to set to version major and minor (so breaking and API additions are tracked).

Since the Windows version of the library already uses version major and version minor, let's just do this universally.

Fixes: 832926f5afebd9a5a81343425e733708332f730f ("Update library version handling for cmake")


egorpugin (approved):

zdenop:
thanks.

PR #4330, opened 2024-10-22
Modernize code for renderers and remove filename conversion for Windows

stweil:
Commit db52047420c54bd added the filename conversion for the hOCR renderer, but it was removed later for TSV in commit 6700edd8bcdd2a4.

Tesseract does not use a filename conversion anywhere else, so remove it for the other renderers, too.

### DIFF
--- a/src/api/hocrrenderer.cpp
+++ b/src/api/hocrrenderer.cpp
@@ -21,9 +21,6 @@
 #include <locale>              // for std::locale::classic
 #include <memory>              // for std::unique_ptr
 #include <sstream>             // for std::stringstream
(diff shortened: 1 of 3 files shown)


src/api/hocrrenderer.cpp
stweil, line 1: The code in hocrrenderer.cpp was already modernized in commit db9c7e0312.

egorpugin (approved):

egorpugin:
For the future changes: is it possible to return unique ptr directly from `GetIterator()`?
If we know that caller must free memory, maybe it is better to return uptr?

PR #4341, opened 2024-11-01
Add GitHub action and Makefile target for Windows installer

stweil:
The new GitHub action "installer-for-windows" builds an NSIS installer with Tesseract for Windows.
It is not run automatically, but must be triggered manually.

Autoconf builds also have a new target `winsetup`, so the installer can be built with `make winsetup`.
This requires a build environment with all needed components (NSIS, Windows libraries, ...).


stweil:
Example: https://github.com/stweil/tesseract/actions/runs/11620260190.

nsis/winpath.exe
egorpugin, line 1: an .exe?

stweil: You are right, I'll add a rule to build it from winpath.cpp.

stweil: Egor, could you please have a look on the sw build failures?

nsis/plugins/INetC.dll
egorpugin, line 1: Is this .dll also needed?

nsis/winpath.exe
egorpugin: Yes, something with new python 3.13.

nsis/plugins/INetC.dll
stweil: It's an extension for the NSIS installer and used for the download of Tesseract models.

nsis/winpath.exe
stweil: I updated the rules to build winpath.exe, so this binary is no longer in the git repository.

nsis/plugins/INetC.dll
egorpugin: Is it possible to download via curl? Curl is present in modern windows.

stweil: I updated Makefile.am to download the zip file and unzip the INetC.dll.

egorpugin (approved):

stweil:
Here is the first result: https://github.com/tesseract-ocr/tesseract/actions/runs/11641358702.

Issue #4349, opened 2024-11-07
tenorflow code in Tesseract


amitdo:
### Your Feature Request

All the files under src/lstm that stars with `tf`.

It is useless. There is no way to train a model in TF that Tesseract can use for inference.

Can we finally remove the TF code from Tesseract?

amitdo:
IIRC, the code in Tesseract uses some internal TF 1.x functionality instead of a stable API.

stweil:
I think I can look in old releases how an OCR engine was added, so we can indeed remove the TensorFlow code now.

PR #4350, opened 2024-11-07
Remove Tensorflow support

stweil:
Tensorflow was never used because of missing models.


stweil:
See issue #4349. Note that `unittest` still uses TensorFlow code (partially disabled) for different purposes.

unittest/third_party/utf/utfdef.h
stweil, line 1: It looks like this file was already unused before the removal of the TensorFlow code.

amitdo (approved):

amitdo:
Thank you!

unittest/third_party/utf/utfdef.h
stweil: I was wrong and had to fix the unittest in commit 2a1ce80a42ad5eb1ddb1b770e98b13c1864911d2.

Issue #4146, opened 2023-10-19
Segfault with certain language packs on certain images


cchadowitz:
### Current Behavior

When running the tesseract binary with specific language packs enabled, the binary segfaults (different places for different combinations).

1. Using the `chi_sim` and the `mrz` language packs together for this input image: ![chi_sim-mrz-segfault](https://github.com/tesseract-ocr/tesseract/assets/24324908/0db4be34-fb01-4a22-b2ce-6251a6751ccd)
```
root@78ce6cbf2cc5:/opt/tesseract/bin# gdb ./tesseract 
GNU gdb (Ubuntu 14.0.50.20230907-0ubuntu1) 14.0.50.20230907-git
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./tesseract...
(No debugging symbols found in ./tesseract)
(gdb) run sample_012631.jpg - --tessdata-dir <snip>/tessdata/ -l chi_sim+mrz
Starting program: /opt/tesseract/bin/tesseract sample_012631.jpg - --tessdata-dir <snip>/tessdata/ -l chi_sim+mrz
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Estimating resolution as 548
Detected 108 diacritics

Program received signal SIGSEGV, Segmentation fault.
0x00005619906db4e8 in tesseract::ELIST_ITERATOR::forward() ()
(gdb) bt
#0  0x00005619906db4e8 in tesseract::ELIST_ITERATOR::forward() ()
#1  0x00005619906bc1ff in tesseract::PAGE_RES_IT::ReplaceCurrentWord(tesseract::PointerVector<tesseract::WERD_RES>*) ()
[...]

Issue #4270, opened 2024-06-18
Assert fail in src/ccstruct/pageres.cpp, line 1502 with specific image and language combination - all languages from ubuntu repos


cchadowitz:
### Current Behavior

When running this command line:

```
$ tesseract /opt/sample_013741.jpg -- -l ara+chi_tra
```
The following occurs:
```
Estimating resolution as 303
!w_it.cycled_list():Error:Assert failed:in file src/ccstruct/pageres.cpp, line 1502
Aborted (core dumped)
```

This is reproducible via the following sequence of commands (output is clipped for brevity until the end) to start a clean Ubuntu 24.04 docker container, update existing packages, install `tesseract-ocr` (for command line usage) and the two languages in question, `tesseract-ocr-ara` and `tesseract-ocr-chi-tra`. The test image is the same image in #4148, `wget` is used to download it to test. It is also available in this ticket below.
```
$ docker run --rm -it ubuntu:24.04 /bin/bash

# apt update && apt upgrade -y && apt install -y wget tesseract-ocr tesseract-ocr-ara tesseract-ocr-chi-tra

[...]

Issue #4148, opened 2023-10-23
Assert fail in src/ccstruct/pageres.cpp, line 1502 with specific image and language combination


marcreichman-pfi:
### Current Behavior

When running this command line:
```bash
$ tesseract sample_013741.jpg - --tessdata-dir <snip>/tessdata -l ara+ocrb_int
```

The following occurs:
```bash
Estimating resolution as 303
!w_it.cycled_list():Error:Assert failed:in file src/ccstruct/pageres.cpp, line 1502
Aborted
```

This was first detected with API usage from an internal app, but reproducible in tesseract commandline.
![sample_013741](https://github.com/tesseract-ocr/tesseract/assets/5814160/ac322732-7d8b-45f8-875c-83a04b278165)

Backtrace:
```Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
44	./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f14a8257866 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f14a823b8b7 in __GI_abort () at ./stdlib/abort.c:79
#5  0x000055f2706dac44 in tesseract::ERRCODE::error(char const*, tesseract::TessErrorLogCode, char const*, ...) const [clone .cold] ()
#6  0x000055f270703676 in tesseract::PAGE_RES_IT::DeleteCurrentWord() ()
#7  0x000055f27079d5a8 in tesseract::Tesseract::recog_all_words(tesseract::PAGE_RES*, tesseract::ETEXT_DESC*, tesseract::TBOX const*, char const*, int) ()
#8  0x000055f270763b46 in tesseract::TessBaseAPI::Recognize(tesseract::ETEXT_DESC*) ()
#9  0x000055f2707643d2 in tesseract::TessBaseAPI::ProcessPage(Pix*, int, char const*, char const*, int, tesseract::TessResultRenderer*) ()
[...]

PR #4357, opened 2024-11-22
Fix linear congruential random number generator

stweil:
This reverts commit 2252936fc801aee6c65a0bae9daee0638f56febf.

Fixes: #4146, #4148, #4270

### DIFF
--- a/src/ccutil/helpers.h
+++ b/src/ccutil/helpers.h
@@ -68,7 +68,9 @@ inline const std::vector<std::string> split(const std::string &s, char c) {
   return v;
 }
 
-// A simple linear congruential random number generator.
+// A simple linear congruential random number generator,
+// using Knuth's constants from:
+// http://en.wikipedia.org/wiki/Linear_congruential_generator.
 class TRand {
 public:
   // Sets the seed to the given value.
@@ -95,7 +97,10 @@ public:
   }
 
 private:
-  std::minstd_rand e;
+  std::linear_congruential_engine<std::uint_fast32_t,
+                                  6364136223846793005ULL,
+                                  1442695040888963407ULL,
+                                  UINT64_MAX> e;
 };
 
 // Remove newline (if any) at the end of the string.


egorpugin:
But what is the reason of incorrect work?

If we check this table
https://en.wikipedia.org/wiki/Linear_congruential_generator#Parameters_in_common_use

we see our parameters from original code
`MMIX by Donald Knuth | 264 | 6364136223846793005 | 1442695040888963407`

Here we can find that 
https://en.cppreference.com/w/cpp/numeric/random
std::minstd_rand is `std::linear_congruential_engine<std::uint_fast32_t,                                48271, 0, 2147483647`

Maybe my port was not presicely correct in terms of that `std::minstd_rand` is not exact what it was, I think we should try to make our code use std.

For example, `using our_rand = std::linear_congruential_engine<std::uint_fast32_t, 6364136223846793005ULL, 1442695040888963407ULL, UINT64_MAX>`

egorpugin:
If new random was more correct, then old code has issues - in that place with `delete` after `assert`.

stweil:
That's a good hint, thank you! I updated my pull request as suggested.

egorpugin:
Also I'll put my analysis here.

Tried to reproduce this example on windows - no success (everything works).
https://github.com/tesseract-ocr/tesseract/issues/4148#issuecomment-2486574952

Is it specific STL issue? libstdc++ only? libc++?

---
Analysis

We have no issues with previous random implemenation.

So, cases are:
1. Code around assert + delete is incorrect and old random code is incorrect, so the issue stayed hidden.
2. Code around assert + delete is incorrect and new random is more correct, so it shows the problem around assert.
3. Code around assert + delete is correct and new random is incorrect - fix rand here.

What case do we have?

Mr586:
Fuck

stweil:
> Does it work with updated parameters like with old rand?

Yes, the updated parameters are sufficient to fix the assertion and the heap issue.

egorpugin:
Good, thank you.

I'm not well aware of tesseract testing, but I think it is worth to add some of those assert triggering images into the test suite.
For example, it should be a good test case for my changes with lists.

stweil:
> Yes, the updated parameters are sufficient to fix the assertion and the heap issue.

The fix these issues, but cause failures of recodebeam_test and intsimdmatrix_test.

egorpugin:
The problem is that 2,3 and 4 args to template become `std::uint_fast32_t`, but they should be kept as 64 bit.
`std::linear_congruential_engine<std::uint_fast32_t, 6364136223846793005ULL, 1442695040888963407ULL, UINT64_MAX>`

like this 

`std::linear_congruential_engine<std::uint_fast64_t, 6364136223846793005ULL, 1442695040888963407ULL, UINT64_MAX>`

stweil:
That does not fix the unittests. I hope I can fix it tomorrow.

egorpugin:
It seems we cannot express
`MMIX by Donald Knuth | 2^64 | 6364136223846793005 | 1442695040888963407`
with std::linear_congruential_engine because of 2^64 modulus (we need 128 bit type for it to pass into template).
But 2^64 modulus is just a division by 1.

So the old code did exactly it plus converting a result into 32 bit int (taking first 31 bits).
```
    u *= 6364136223846793005ULL;
    u += 1442695040888963407ULL;
    return u >> 33;
```

egorpugin:
It is even more clear now.

Our tests depends on random generator (default seed). Output is always different - unreliable.

Tesseract execution depends on random generator. Is random even seeded there?
If no, then tesseract code has issue around that `assert + delete`, it is unrelated to random generator.
Random only showed that there is an issue somewhere there.

stweil:
I restored the old random generator (and the related unittest code) now. This fixes the unittests and the reported issues.

egorpugin:
No, it does not fix the problem. Which is the problem around those lists, asserts and delete.
Only hides it behind unseeded rng.

stweil:
I don't understand that. Do you have a test case which still triggers the assertion?

egorpugin:
Code behavior depends on random generator.
We observe crashes on one generator and do not see them on other.

That means that the problem is not in rng, but in that code itself.
Old rng just hides the problem where changes in rng uncover it.

You can try to change initial seed (`1` currently) to something else and try provided cases with it.

stweil:
The initial seed 1 is not used, because the recognition sets a new seed in `SetRandomSeed` (not only once, but many times). And it calls `TRand::SignedRand` several thousand times.

egorpugin:
I cannot even call it 'random'. Not a fair one.
It is just a series of same numbers.

I can understand if they are time based so they change each run, but not like we have it now.

stweil:
It's good that the series is not random, but pseudorandom, because otherwise some results would also be random (which is not desired).

And I think that I now know why commit 32fee1944772ba98f2c181e0a90221c09def78a7 still produced buggy code: the implementation of `IntRand` could return negative numbers which violated the specification. Fixing this also fixed the memory issues and the assertion.

egorpugin:
I see. Indeed, `return e() >> 33;` part was missing.

egorpugin:
> It's good that the series is not random, but pseudorandom

It is not even pseudorandom. It is just a fixed series. Always the same. And it hides other code issues.

Issue #4354, opened 2024-11-17
CLI: Unexpected handling of parameters using -c VAR=VALUE


sdubinsky:
_Note by @amitdo. The original title was:` Inconsistency Between --oem and -c "tessedit_ocr_engine_mode"`_

### Current Behavior

As far as I can tell, these two options should be identical, however the attached image produces different output.  With  --oem it says 11, with the config var it says 1.  I should add that with any --oem value other than 0, it returns '1' not '11'.

Full command: 

```
tesseract -l eng --oem 0 --psm 7  -c "tessedit_char_whitelist=0123456789"  ramdisk/test_0.png  -
tesseract -l eng --psm 7  -c "tessedit_char_whitelist=0123456789" -c "tessedit_ocr_engine_mode=0" ramdisk/test_0.png  -
```

![test_0](https://github.com/user-attachments/assets/a1766c20-42a0-4fa4-8885-be63b8e2b37d)

### Expected Behavior

Both commands should return the same result, ideally 11.

### Suggested Fix

They should either return the same information or the docs should be updated to clarify the difference.

### tesseract -v

tesseract 5.4.1
 leptonica-1.84.1
  libgif 5.2.2 : libjpeg 6b (libjpeg-turbo 3.0.2) : libpng 1.6.40 : libtiff 4.6.0 : zlib 1.3.1.zlib-ng : libwebp 1.4.0
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
 Found libcurl/8.9.1 OpenSSL/3.2.2 zlib/1.3.1.zlib-ng libidn2/2.3.7 nghttp2/1.62.1

### Other Operating System

Fedora 42.

### uname -a

Linux deux-ex 6.11.7-300.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Nov  8 19:23:10 UTC 2024 x86_64 GNU/Linux

### CPU

Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz (12)

### Virtualization / Containers

None.

zdenop:
Why do you think the OEM default value is 0?

https://github.com/tesseract-ocr/tesseract/blob/43b8d743fce4e83cbbe6198b6a917ab5c672a82b/include/tesseract/publictypes.h#L263-L277

https://github.com/tesseract-ocr/tesseract/blob/43b8d743fce4e83cbbe6198b6a917ab5c672a82b/src/tesseract.cpp#L688-L692

[...]

PR #4363, opened 2024-11-27
add cli `-c` parameter(s) to init vectors

zdenop:
see #4354

### DIFF
--- a/src/tesseract.cpp
+++ b/src/tesseract.cpp
@@ -273,32 +273,6 @@ static void PrintHelpMessage(const char *program) {
       program, program, program);
 }
 
-static bool SetVariablesFromCLArgs(tesseract::TessBaseAPI &api, int argc, char **argv) {
-  bool success = true;
-  char opt1[256], opt2[255];
-  for (int i = 0; i < argc; i++) {
-    if (strcmp(argv[i], "-c") == 0 && i + 1 < argc) {
-      strncpy(opt1, argv[i + 1], 255);
-      opt1[255] = '\0';
-      char *p = strchr(opt1, '=');
-      if (!p) {
-        fprintf(stderr, "Missing = in configvar assignment\n");
-        success = false;
-        break;
-      }
-      *p = 0;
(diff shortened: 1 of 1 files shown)


amitdo:
Thanks!

stweil:
What happens now if I want to override parameters which are normally set by the model file?

old: model file parameters could be overridden by command line parameters
new: command line parameters are evaluated before the model file parameters are set?

stweil:
Old error message for wrong usage:
```
Missing = in configvar assignment
```

New error message:
```
libc++abi: terminating due to uncaught exception of type std::invalid_argument: Missing '=' in configvar assignment
zsh: abort      tesseract x - -c x
```

Throwing an error changes the error message which is presented to users. Why not simply print the old error message?

amitdo:
Responding for your first comment:

It not clear if you wonder if there is an issue here or **know** there is an issue.

Regarding your second comment, yes, we should change this to print a simple error message.

amitdo:
stweil,

See the code in:
https://github.com/tesseract-ocr/tesseract/blob/2e5a114bd2aab932b53e103d1391e7f27d2d8cff/src/ccmain/tessedit.cpp#L288

zdenop:
@stweil : Could it be like this?

```diff
diff --git a/src/tesseract.cpp b/src/tesseract.cpp
index 08a65e14..6025af5f 100644
--- a/src/tesseract.cpp
+++ b/src/tesseract.cpp
@@ -462,7 +462,8 @@ static bool ParseArgs(int argc, char **argv, const char **lang, const char **ima
       const std::string argument(argv[i + 1]);
       const auto equal_pos = argument.find('=');
       if (equal_pos == std::string::npos) {
-          throw std::invalid_argument("Missing '=' in configvar assignment");
+        fprintf(stderr, "Missing = in configvar assignment\n");
+        return false;
       }
       // Extract key and value
       const std::string key = argument.substr(0, equal_pos);
```

zdenop:
@stweil : Are you fine with proposal above or does 0a9df94fffffb2182836e69b7a33d4d455a34fc7 solve the problem?

PR #4365, opened 2024-12-02
Check if datadir exists: fixes #4364

### LINKED ISSUE #4364: Unexpected C++ exception in tesseract::addAvailableLanguages
### Current Behavior

$ LANG=en ls /usr/share/tessdata
ls: cannot access '/usr/share/tessdata': No such file or directory

#8  0x00007ffff7aad57b in std::filesystem::__cxx11::recursive_directory_iterator::recursive_directory_iterator
    (this=0x7fffffffc2b0, __p=filesystem::path "/usr/share/tessdata/" = {...}, __options=(std::filesystem::directory_options::follow_directory_symlink | std::filesystem::directory_options::skip_permission_denied))
    at /usr/include/c++/14/bits/fs_dir.h:514
#9  tesseract::addAvailableLanguages (datadir="/usr/share/tessdata/", langs=langs@entry=0x7fffffffc620) at src/api/baseapi.cpp:152
#10 0x00007ffff7aae2d0 in tesseract::TessBaseAPI::GetAvailableLanguagesAsVector (this=this@entry=0x7fffdc0021d8, langs=langs@entry=0x7fffffffc620) at src/api/baseapi.cpp:399

### Expected Behavior

It doesn't crash.

### Suggested Fix

https://github.com/tesseract-ocr/tesseract/blob/main/src/api/baseapi.cpp#L149:
Add check here whether path/file exists and return if not.

### tesseract -v

_No response_

### Operating System

_No response_

### Other Operating System

_No response_

### uname -a

_No response_

### Compiler

_No response_

### CPU

_No response_

_No response_

### Other Information

_No response_

zdenop:
recursive_directory_iterator in addAvailableLanguages crashes if datadir does not exist.

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -146,6 +146,10 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
+  if (!std::filesystem::exists(datadir)) {
+    std::cerr << "Error: The directory '" << datadir << "' does not exist.\n";
+    return;
+  }
   for (const auto& entry :
        std::filesystem::recursive_directory_iterator(datadir,
          std::filesystem::directory_options::follow_directory_symlink |


src/api/baseapi.cpp
```diff
@@ -146,6 +146,10 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
+  if (!std::filesystem::exists(datadir)) {
```
stweil, line 149: This check is not sufficient. It passes if `datadir` exists and is a file (which still raises an exception). I suggest to catch the exception which is thrown by `std::filesystem::recursive_directory_iterator`.

stweil (changes requested):

src/api/baseapi.cpp
zdenop: https://github.com/zdenop/tesseract/blob/2e5a114bd2aab932b53e103d1391e7f27d2d8cff/src/ccutil/ccutil.cpp#L95-L97

This causes that `datadir` is always handled as directory.

stweil: On macOS:
```
% touch a.txt
% tesseract --list-langs --tessdata-dir a.txt                       
libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in recursive_directory_iterator: Not a directory ["a.txt/"]
zsh: abort      tesseract --list-langs --tessdata-dir a.txt
```

egorpugin: We don't have main wrapped into try catch?
Error text is ok, we need only catch std::exception globally.

zdenop: @stweil : I tested this PR on linux  (g++ 11.3.0) and it does not produce "uncaught exception" (without egorpugin commit). Do you still need  to add something like this?
```diff 
diff --git a/src/api/baseapi.cpp b/src/api/baseapi.cpp
index 667ead67..e684c7fe 100644
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -146,6 +146,10 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
+  if (!std::filesystem::is_directory(datadir)) {
+    std::cerr << "Error: '" << datadir << "' is not directory.\n";
+    return;
+  }
   if (!std::filesystem::exists(datadir)) {
     std::cerr << "Error: The directory '" << datadir << "' does not exist.\n";
     return;
```

Or should be make these checks at [ccutils.cpp](https://github.com/zdenop/tesseract/blob/2e5a114bd2aab932b53e103d1391e7f27d2d8cff/src/ccutil/ccutil.cpp#L44) and fail if directory not exists?

stweil: This changes the error messages.

Old
```
libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in recursive_directory_iterator: No such file or directory ["tessdata/"]
libc++abi: terminating due to uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in recursive_directory_iterator: Not a directory ["tessdata/"]
```

New:
```
exception: filesystem error: in recursive_directory_iterator: Not a directory ["tessdata/"]
exception: filesystem error: in recursive_directory_iterator: No such file or directory ["tessdata/"]
```

The new error messages are nicer, but are they user friendly? And do we want to introduce C++ exceptions in the C / C++ API?

egorpugin: 1. C API should be removed later.
2. We want c++ exceptions in api. Nothing wrong here.

stweil: Thanks Egor. Then I think that this pull request is no longer needed, and issue #4364 can be closed.

PR #4369, opened 2024-12-15
handle colormaps correctly

### LINKED ISSUE #4127: Memory access violation when using SetRectangle for 8Bit images with colormap
### Current Behavior

Hi

I got a memory access violation when using SetRectangle and an 8Bit image.
I only tested the simple example from the tesseract documentation (SetRectangle_example from [https://tesseract-ocr.github.io/tessdoc/Examples_C++.html](https://tesseract-ocr.github.io/tessdoc/Examples_C++.html)).
My image is a PNG file. When saving as 8Bit image I got the crash in debug mode. When I save the image in 32bit, then it works properly.

The VS debugger recognized the violation in the method _**HistogramRect**_ in the file **otsuthr.cpp** in line 157: **_int pixel = GET_DATA_BYTE(linedata, (x + left) * num_channels + channel);_**

I tried to look if there is an error in the formula for calculating the address. But I am not familiar with the Tesseract internals (installed it last week for the first time), so I have no idea what the error causes.

But I think the mentioned method is a little weird. The input parameter **_src_pix_** is a pointer to the small rectangle image part, not to the overall (big) image (width and height have the size of the rectangle). Based on this it uses the WPL of the small rectangle image (**_src_wpl_**) and not of the whole original picture with much bigger resolution.
But when calculating the address of the pixels, the algorithm uses the pixel coordinates of the rectangle within the whole big picture (_**left, top, width, height**_), these are "global" pixel coodinates. So it seems, that the image is only a shallow structure pointing to the memory of the actual loaded image.
But if this is true: why does the line **_const l_uint32 *linedata = srcdata + y * src_wpl;_** use the WPL of the small rectangle and not of the actual whole image? If it is the memory of the loaded image, then it should be the stride based in the width of the image and not the width of the small rectangle.
On the other side: it seems to work properly with 32 bit images. So if the address calculation is wrong, then it should also cause problems with all other channel counts.

These are my thoughts. Hope it helps a little bit.

Best regards

### Expected Behavior

_No response_

### Suggested Fix

_No response_

### tesseract -v

Tesseract 5.3.2

### Operating System

Windows 10

### Other Operating System

_No response_

### uname -a

_No response_

### Compiler

MSVC 2022 17.7.4, x64

### CPU

_No response_

_No response_

### Other Information

_No response_

zdenop:
handle colormaps correctly  - fixes #4127

### DIFF
--- a/src/ccmain/thresholder.cpp
+++ b/src/ccmain/thresholder.cpp
@@ -283,30 +283,24 @@ bool ImageThresholder::ThresholdToPix(Image *pix) {
     tprintf("Image too large: (%d, %d)\n", image_width_, image_height_);
     return false;
   }
-  Image original = GetPixRect();
+  // Handle binary image
   if (pix_channels_ == 0) {
     // We have a binary image, but it still has to be copied, as this API
     // allows the caller to modify the output.
+    Image original = GetPixRect();
     *pix = original.copy();
-  } else {
-    if (pixGetColormap(original)) {
-      Image tmp;
-      Image without_cmap =
-          pixRemoveColormap(original, REMOVE_CMAP_BASED_ON_SRC);
-      int depth = pixGetDepth(without_cmap);
-      if (depth > 1 && depth < 8) {
-        tmp = pixConvertTo8(without_cmap, false);
-      } else {
-        tmp = without_cmap.copy();
-      }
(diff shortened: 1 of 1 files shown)


stweil:
Thanks!

PR #4372, opened 2024-12-18
Modernize datadir

zdenop:
- use std::filesystem::path instead of std::string for datadir
- add warning if datadir is not directory or does not exists

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -146,9 +146,6 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
-  if (!std::filesystem::is_directory(datadir))
-    return;
-
   for (const auto& entry :
        std::filesystem::recursive_directory_iterator(datadir,
          std::filesystem::directory_options::follow_directory_symlink |
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -146,10 +146,9 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
-  if (!std::filesystem::exists(datadir) ||
-      !std::filesystem::is_directory(datadir)) {
+  if (!std::filesystem::is_directory(datadir))
     return;
-  }
+
   for (const auto& entry :
        std::filesystem::recursive_directory_iterator(datadir,
          std::filesystem::directory_options::follow_directory_symlink |
@@ -400,7 +399,7 @@ void TessBaseAPI::GetLoadedLanguagesAsVector(std::vector<std::string> *langs) co
 void TessBaseAPI::GetAvailableLanguagesAsVector(std::vector<std::string> *langs) const {
   langs->clear();
   if (tesseract_ != nullptr) {
-    addAvailableLanguages(tesseract_->datadir.string().c_str(), langs);
+    addAvailableLanguages(tesseract_->datadir.string(), langs);
     std::sort(langs->begin(), langs->end());
   }
 }
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -347,7 +347,7 @@ int TessBaseAPI::Init(const char *data, int data_size, const char *language, Ocr
   // Update datapath and language requested for the last valid initialization.
   datapath_ = std::move(datapath);
   if (datapath_.empty() && !tesseract_->datadir.empty()) {
-    datapath_ = tesseract_->datadir;
+    datapath_ = tesseract_->datadir.string();
   }
 
   language_ = language;
@@ -396,7 +396,7 @@ void TessBaseAPI::GetLoadedLanguagesAsVector(std::vector<std::string> *langs) co
 void TessBaseAPI::GetAvailableLanguagesAsVector(std::vector<std::string> *langs) const {
   langs->clear();
   if (tesseract_ != nullptr) {
-    addAvailableLanguages(tesseract_->datadir, langs);
+    addAvailableLanguages(tesseract_->datadir.string().c_str(), langs);
     std::sort(langs->begin(), langs->end());
   }
 }
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -858,7 +858,7 @@ const char *TessBaseAPI::GetInputName() {
 }
 
 const char *TessBaseAPI::GetDatapath() {
-  return tesseract_->datadir.c_str();
+  return datapath_.c_str();
 }
 
 int TessBaseAPI::GetSourceYResolution() {
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -146,6 +146,10 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
+  if (!std::filesystem::exists(datadir) ||
+      !std::filesystem::is_directory(datadir)) {
+    return;
+  }
   for (const auto& entry :
        std::filesystem::recursive_directory_iterator(datadir,
          std::filesystem::directory_options::follow_directory_symlink |
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -83,6 +83,7 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
                                          bool set_only_non_debug_params, TessdataManager *mgr) {
   // Set the language data path prefix
   lang = !language.empty() ? language : "eng";
+  language_data_path_prefix = datadir.string();
   std::filesystem::path tessdata_path = datadir / (lang + "." + kTrainedDataSuffix);
 
   // Initialize TessdataManager.
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -29,6 +29,7 @@
 #include "params.h"
 #include "stopper.h"
 #include "tesseractclass.h"
+#include "tesserrstream.h"  // for tesserr
 #include "tessvars.h"
 #include "tprintf.h"
 #ifndef DISABLED_LEGACY_ENGINE
@@ -86,10 +87,9 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
 
   // Initialize TessdataManager.
   if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string().c_str())) {
-    tprintf("Error opening data file %s\n", tessdata_path.string());
-    tprintf(
+    tesserr << "Error opening data file " << tessdata_path.string() << '\n' <<
         "Please make sure the TESSDATA_PREFIX environment variable is set"
-        " to your \"tessdata\" directory.\n");
+        " to your \"tessdata\" directory.\n";
     return false;
   }
 #ifdef DISABLED_LEGACY_ENGINE
@@ -182,9 +182,8 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
   }
 #ifndef DISABLED_LEGACY_ENGINE
   else if (!mgr->GetComponent(TESSDATA_UNICHARSET, &fp) || !unicharset.load_from_file(&fp, false)) {
-    tprintf(
-        "Error: Tesseract (legacy) engine requested, but components are "
-        "not present in %s!!\n", tessdata_path.string());
+    tesserr << "Error: Tesseract (legacy) engine requested, but components are "
+        "not present in " << tessdata_path.string() << "!!\n";
     return false;
   }
 #endif // ndef DISABLED_LEGACY_ENGINE
--- a/src/ccmain/tessedit.cpp
+++ b/src/ccmain/tessedit.cpp
@@ -43,24 +43,25 @@ namespace tesseract {
 // Read a "config" file containing a set of variable, value pairs.
 // Searches the standard places: tessdata/configs, tessdata/tessconfigs
 // and also accepts a relative or absolute path name.
-void Tesseract::read_config_file(const char *filename, SetParamConstraint constraint) {
-  std::string path = datadir;
-  path += "configs/";
-  path += filename;
-  FILE *fp;
-  if ((fp = fopen(path.c_str(), "rb")) != nullptr) {
-    fclose(fp);
-  } else {
-    path = datadir;
-    path += "tessconfigs/";
-    path += filename;
-    if ((fp = fopen(path.c_str(), "rb")) != nullptr) {
-      fclose(fp);
-    } else {
-      path = filename;
-    }
-  }
-  ParamUtils::ReadParamsFile(path.c_str(), constraint, this->params());
+void Tesseract::read_config_file(const char *filename,
+                                 SetParamConstraint constraint) {
+  // Construct potential config file paths
+  std::vector<std::filesystem::path> config_paths = {
+      datadir / "configs" / filename,
+      datadir / "tessconfigs" / filename,
+      std::filesystem::path(filename)};
+
+  // Use the first existing file or fallback to the last (filename)
+  auto config_file = std::find_if(config_paths.begin(), config_paths.end(),
+                                  [](const std::filesystem::path &path) {
+                                    std::error_code ec;
+                                    return std::filesystem::exists(path, ec);
+                                  });
+  const std::filesystem::path &selected_path =
+      (config_file != config_paths.end()) ? *config_file : config_paths.back();
+
+  ParamUtils::ReadParamsFile(selected_path.string().c_str(), constraint,
+                             this->params());
 }
 
 // Returns false if a unicharset file for the specified language was not found
@@ -81,14 +82,11 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
                                          bool set_only_non_debug_params, TessdataManager *mgr) {
   // Set the language data path prefix
   lang = !language.empty() ? language : "eng";
-  language_data_path_prefix = datadir;
-  language_data_path_prefix += lang;
-  language_data_path_prefix += ".";
+  std::filesystem::path tessdata_path = datadir / (lang + "." + kTrainedDataSuffix);
 
   // Initialize TessdataManager.
-  std::string tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
-  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.c_str())) {
-    tprintf("Error opening data file %s\n", tessdata_path.c_str());
+  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string().c_str())) {
+    tprintf("Error opening data file %s\n", tessdata_path.string().c_str());
     tprintf(
         "Please make sure the TESSDATA_PREFIX environment variable is set"
         " to your \"tessdata\" directory.\n");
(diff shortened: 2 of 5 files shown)


egorpugin (approved):
Probably all is_directory() checks are unnecessary. User will just get any other random error if he points to a file instead of to a dir.

src/api/baseapi.cpp
```diff
@@ -146,6 +146,10 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
+  if (!std::filesystem::exists(datadir) ||
+      !std::filesystem::is_directory(datadir)) {
```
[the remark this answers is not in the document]
zdenop: So it turns out that when I create a  link correctly ( :-) ) it works on Windows (10 64bit, VS2019) too:

```sh
>mklink /D c:\temp\mydata_dir F:\Projects\Community\tessdata
symbolic link created for c:\temp\mydata_dir <<===>> F:\Projects\Community\tessdata

>ls -l c:\temp\ | grep mydata
lrwxrwxrwx 1 NBA+zdenop 197121      30 Dec 18 18:08 mydata -> /f/Projects/Community/tessdata
lrwxrwxrwx 1 NBA+zdenop 197121      30 Dec 18 18:08 mydata_dir -> /f/Projects/Community/tessdata

>tesseract --list-langs --tessdata-dir c:\temp\mydata_dir | head
List of available languages in "c:\temp\mydata_dir" (498):
3.04\eng
afr
amh
ara
asm
aze
aze_cyrl
bel
ben

>tesseract --list-langs --tessdata-dir c:\temp\mydata
exception: exists: Access is denied.: "c:\temp\mydata"
```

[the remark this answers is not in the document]
stweil: Is the test `!std::filesystem::exists(datadir)` needed, or can it be removed?

src/api/baseapi.cpp
```diff
@@ -396,7 +400,7 @@ void TessBaseAPI::GetLoadedLanguagesAsVector(std::vector<std::string> *langs) co
 void TessBaseAPI::GetAvailableLanguagesAsVector(std::vector<std::string> *langs) const {
   langs->clear();
   if (tesseract_ != nullptr) {
-    addAvailableLanguages(tesseract_->datadir, langs);
+    addAvailableLanguages(tesseract_->datadir.string().c_str(), langs);
```
stweil, line 403: addAvailableLanguages takes a std::string, so c_str() should be removed.

src/ccmain/tessedit.cpp
```diff
@@ -81,14 +82,11 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
...
-  language_data_path_prefix += lang;
-  language_data_path_prefix += ".";
+  std::filesystem::path tessdata_path = datadir / (lang + "." + kTrainedDataSuffix);
 
   // Initialize TessdataManager.
-  std::string tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
-  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.c_str())) {
-    tprintf("Error opening data file %s\n", tessdata_path.c_str());
+  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string().c_str())) {
+    tprintf("Error opening data file %s\n", tessdata_path.string().c_str());
```
stweil, line 89: Here the new stream output would not need c_str().

src/ccmain/tessedit.cpp
```diff
@@ -187,7 +185,7 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
     tprintf(
```
stweil, line 185: Use new stream output without need for c_str()?

src/ccutil/ccutil.cpp
```diff
@@ -33,68 +31,74 @@ CCUtil::CCUtil()
...
-      if (separator != nullptr) {
-        *separator = '\0';
-        std::string subdir = path;
-        subdir += "/tessdata";
-        if (std::filesystem::exists(subdir)) {
-          datadir = subdir;
-        }
-      }
+    std::filesystem::path path(argv0);
+    if (!std::filesystem::exists(path) ||
```
stweil, line 53: It should be sufficient to test the 2nd part of the condition (`!std::filesystem::is_directory(path)`).

zdenop:
@stweil : Thank you for review. Unittests needs be fixed yet.

src/ccmain/tessedit.cpp
zdenop: Fuzzer does not like it:
```src/ccmain/tessedit.cpp:89:45: error: cannot pass non-trivial object of type 'std::string' (aka 'basic_string<char>') to variadic function; expected type from format string was 'char *' [-Wnon-pod-varargs]
   89 |     tprintf("Error opening data file %s\n", tessdata_path.string());
      |                                      ~~     ^~~~~~~~~~~~~~~~~~~~~~
src/ccmain/tessedit.cpp:89:45: note: did you mean to call the c_str() method?
```

egorpugin:
Old tprintfs should be removed slowly in favor of std::format.

src/ccmain/tessedit.cpp
```diff
@@ -81,14 +82,11 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
...
-  language_data_path_prefix += lang;
-  language_data_path_prefix += ".";
+  std::filesystem::path tessdata_path = datadir / (lang + "." + kTrainedDataSuffix);
 
   // Initialize TessdataManager.
-  std::string tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
-  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.c_str())) {
-    tprintf("Error opening data file %s\n", tessdata_path.c_str());
+  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string().c_str())) {
+    tprintf("Error opening data file %s\n", tessdata_path.string());
```
stweil, line 89:
```suggestion
    tesserr << "Error opening data file tessdata_path.string() << '\n';
```

src/ccmain/tessedit.cpp
stweil: Yes, tprintf had to be replaced by tesserr. I fixed this now.

stweil:
> Old tprintfs should be removed slowly in favor of std::format.

Isn't the tesserr stream even better?

egorpugin:
> Isn't the tesserr stream even better?

I mean with `std::format()` the usage will be something like `tesserr << std::format("format string", args ... );`.

egorpugin:
I've slightly updated `tprintf()` to C++ style.

https://github.com/tesseract-ocr/tesseract/commit/8cb04183c1953f9889558986a786d46bcd832395

stweil:
> I've slightly updated `tprintf()` to C++ style.

Now I get numerous compiler warnings:
```
../../../src/ccutil/tprintf.h:37:33: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
```

egorpugin:
1. Warning should be silenced. Updated code is completely fine and valid. We should remove `-Wformat-security` flag if set or set `-Wno-format-security`.
2. What is code size change? In general this should be ignored. But I'm curious what the difference is. Probably 5 bytes for each `get_debugfp()` and some pre- and post call setups.

egorpugin:
Can you try changing it to
```
template <typename ... Types>
auto tprintf(const char *format, Types && ... args) {
  return fprintf(get_debugfp(), format, std::forward<Types>(args)...);
}
```
?

stweil:
> Can you try changing it to [...]

The suggested change still produces the compiler warnings, but I think it improves the code because like that calling `tprintf()` without any argument would be caught as an error.

egorpugin:
It will be an error too if we call `tprintf()` even without `const char *format`.

```
1>tesseract\src\ccutil\tprintf.h(37,10): error C2660: 'fprintf': function does not take 1 arguments
1>(compiling source file '../../../../../src/ccutil/tprintf.cpp')
1>    C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt\stdio.h(830,37):
1>    see declaration of 'fprintf'
1>    tesseract\src\ccutil\tprintf.h(37,10):
1>    while trying to match the argument list '(FILE *)'
1>    tesseract\src\ccutil\tprintf.h(37,10):
1>    the template instantiation context (the oldest one first) is
1>        tesseract\src\ccutil\tprintf.cpp(74,3):
1>        see reference to function template instantiation 'auto tesseract::tprintf<>(void)' being compiled
```

egorpugin:
Added format in https://github.com/tesseract-ocr/tesseract/commit/d95e9f7905cc9427d4d2d351e3ea696e98739004

I'll check ci for other warnings.

egorpugin:
I don't see warnings on gcc-14 (even without format).
Only with clang.

Possible solution is
```template <typename ... Types>
auto tprintf(const char *format, Types && ... args) {
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wformat-security"
  return fprintf(get_debugfp(), format, std::forward<Types>(args)...);
#pragma clang diagnostic pop
}```

**Comment by zdenop** [comment]
Seems like unittests GA are failing on the main (too), so it is not related to this PR:
image
Did I understand it correctly?

**Comment by zdenop** [comment]
@egorpugin : I cherry-picked your commits regarding tprintf to my local branch it does not work for 'tessdata_path.string()'.
E.g. it produces output like `Error opening data file ��]�` for `tesseract a b -l x` (Win11 VS2022 64bit).
When I use `tessdata_path.string().c_str()` output is correct
 `Error opening data file F:\Projects\Community\tessdata\x.traineddata`....

**Comment by egorpugin** [comment]
How do you use it?

**Comment by zdenop** [comment]
I am sorry, it looks like I messed something up... Today I started from clean setup and it worked for me:
```
git pull
git fetch origin pull/4372/head:pr4372
git switch pr4372
git cherry-pick 8cb04183c1953f988
git cherry-pick d95e9f7905cc9427d
git cherry-pick 2a944fbe98ed4408a

./autogen.sh && ./configure --prefix=/usr && make -j4 && make -j4 training
sudo make install && sudo make training-install
```

**Comment by zdenop** [comment]
I looks like unittest fails if `api_.Init` uses `tesseract::OEM_TESSERACT_ONLY` but I am not able to find reason why (well api_. returns empty string, but why?)

If I made simplified test based on `unittest/pagesegmode_test.cc` it work for me. So where is problem? (gtest is installed from unittest/third_party/googletest)

```
/* g++ ocr_gtest.cpp -o ocr_gtest -lleptonica -ltesseract -lgtest
*/
##include <gtest/gtest.h>

TEST(OCRTest, ReadImage) {
    tesseract::TessBaseAPI api_;
    std::string TessdataPath = "/opt/Projects/tessdata";
    std::string filename = "test/testing/segmodeimg.tif";

    Pix *image = pixRead(filename.c_str());
    ASSERT_NE(image, nullptr) << "Failed to read image";

    ASSERT_EQ(api_.Init(TessdataPath.c_str(), "eng"), 
              tesseract::OEM_TESSERACT_ONLY) << "Failed to initialize Tesseract";

    api_.SetPageSegMode(tesseract::PSM_SINGLE_WORD);
    api_.SetImage(image);
    api_.SetRectangle(237, 393, 256, 36);
    char *ocr_text = api_.GetUTF8Text();
    ASSERT_NE(ocr_text, nullptr) << "OCR returned null";
    
    std::string ocr_output(ocr_text);
    printf("OCR output:\n'%s'\n", ocr_output.c_str());

    EXPECT_EQ(ocr_output, "What should\n");

    delete[] ocr_text; // Free OCR text
    api_.End(); // Cleanup Tesseract API
    pixDestroy(&image); // Cleanup image
}

int main(int argc, char** argv) {
    ::testing::InitGoogleTest(&argc, argv);
    return RUN_ALL_TESTS();
}
```

File: src/api/baseapi.cpp [hunk 7]
```diff
@@ -146,6 +146,9 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
+  if (!std::filesystem::is_directory(datadir))
```
stweil, line 149: Do we still want this check? My impression is that we agreed to prefer reporting the exception which is thrown without this conditional code.

zdenop: If throwing an exception is the agreed-upon approach, then this check is no longer necessary.

stweil (changes requested):
We cannot merge this PR as long as several CI tests fail. Some of them seem to be related to layout recognition (for example layout_test) and could indicate a severe regression.

src/ccutil/ccutil.cpp
```diff
@@ -33,68 +32,72 @@ CCUtil::CCUtil()
...
-
-  const char *tessdata_prefix = getenv("TESSDATA_PREFIX");
-
-  // Ignore TESSDATA_PREFIX if there is no matching filesystem entry.
-  if (tessdata_prefix != nullptr && !std::filesystem::exists(tessdata_prefix)) {
-    tprintf("Warning: TESSDATA_PREFIX %s does not exist, ignore it\n", tessdata_prefix);
-    tessdata_prefix = nullptr;
-  }
-
+std::filesystem::path find_data_path(const std::string &argv0) {
```
stweil, line 50:
```suggestion
static std::filesystem::path find_data_path(const std::string &argv0) {
```

src/ccutil/ccutil.cpp
```diff
@@ -33,68 +32,72 @@ CCUtil::CCUtil()
...
 
+
+/**
+ * @brief CCUtil::main_setup - set location of tessdata and name of image
+ *
+ * @param argv0 - paths to the directory with language files and config files.
+ */
+void CCUtil::main_setup(const std::string &argv0, const std::string &basename) {
+  imagebasename = basename; /**< name of image */
+  datadir = find_data_path(argv0);
```
stweil, line 101: The old `datadir` was always terminated with a '/' which is missing for the new datadir`. This is an API change, and it might also be the reason for the failing CI tests.

src/ccmain/tessedit.cpp
```diff
@@ -81,17 +83,13 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
...
-  language_data_path_prefix += ".";
+  std::filesystem::path tessdata_path = datadir / (lang + "." + kTrainedDataSuffix);
 
   // Initialize TessdataManager.
-  std::string tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
-  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.c_str())) {
-    tprintf("Error opening data file %s\n", tessdata_path.c_str());
-    tprintf(
+  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string().c_str())) {
+    tesserr << "Error opening data file " << tessdata_path.string() << '\n' <<
```
stweil, line 90:
```suggestion
    tesserr << "Error opening data file " << tessdata_path << '\n' <<
```

src/ccutil/ccutil.cpp
stweil, line 101:
```suggestion
  datadir = find_data_path(argv0) / "";
```

src/ccmain/tessedit.cpp
```diff
@@ -81,17 +83,13 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
...
-  language_data_path_prefix += lang;
-  language_data_path_prefix += ".";
+  std::filesystem::path tessdata_path = datadir / (lang + "." + kTrainedDataSuffix);
 
   // Initialize TessdataManager.
-  std::string tessdata_path = language_data_path_prefix + kTrainedDataSuffix;
-  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.c_str())) {
-    tprintf("Error opening data file %s\n", tessdata_path.c_str());
-    tprintf(
+  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.string().c_str())) {
```
stweil, line 89:
```suggestion
  if (!mgr->is_loaded() && !mgr->Init(tessdata_path.c_str())) {
```

src/ccmain/tessedit.cpp
zdenop: Regarding the difference in path formatting between using `tesseract.exe a b -l c` with `tessdata_path.string()` versus `tessdata_path`:

- `tessdata_path.string()` produces a native OS path (e.g., on Windows):
```
Error opening data file X:\Projects\tesseract\c.traineddata
```
- `tessdata_path` produces an escaped path:
```
Error opening data file "X:\\Projects\\tesseract\\c.traineddata"
```
I believe it’s more user-friendly and intuitive to display the native OS path. Is there a specific reason we should use the escaped version instead? If not, I recommend sticking with the native path for error messages.

src/ccmain/tessedit.cpp
```diff
@@ -184,10 +182,8 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
   }
 #ifndef DISABLED_LEGACY_ENGINE
   else if (!mgr->GetComponent(TESSDATA_UNICHARSET, &fp) || !unicharset.load_from_file(&fp, false)) {
-    tprintf(
-        "Error: Tesseract (legacy) engine requested, but components are "
-        "not present in %s!!\n",
-        tessdata_path.c_str());
+    tesserr << "Error: Tesseract (legacy) engine requested, but components are "
+        "not present in " << tessdata_path.string() << "!!\n";
```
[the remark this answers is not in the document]
zdenop: The same as above:  I recommend sticking with the native path for error messages

src/ccmain/tessedit.cpp
zdenop: This change would broke msvc build:
`error C2664: 'bool tesseract::TessdataManager::Init(const char *)': cannot convert argument 1 from 'const std:
:filesystem::path::value_type *' to 'const char *'`

src/ccutil/ccutil.cpp
zdenop: Previously, `datadir` was a `string`, and appending a `/` was used as a workaround to handle cases where users omitted the directory separator.
However, with `datadir` now being a `std::filesystem::path`, this manual string manipulation is no longer necessary. The path class automatically normalizes path separators and handles path operations (like concatenation) correctly for the underlying OS. 
As a result, we can safely remove the workaround and rely on the standard library to manage path separators and concatenation.
Or do I miss something?

src/ccmain/tessedit.cpp
```diff
@@ -81,17 +83,13 @@ bool Tesseract::init_tesseract_lang_data(const std::string &arg0,
                                          bool set_only_non_debug_params, TessdataManager *mgr) {
   // Set the language data path prefix
   lang = !language.empty() ? language : "eng";
-  language_data_path_prefix = datadir;
```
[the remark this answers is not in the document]
zdenop: Thanks. Fixed [2a296fa](https://github.com/tesseract-ocr/tesseract/pull/4372/commits/2a296fabd8fd3d3d2257409e95373e2cfb6897a6)

src/ccutil/ccutil.cpp
```diff
@@ -33,68 +32,72 @@ CCUtil::CCUtil()
...
-    datadir = TESSDATA_PREFIX "/tessdata/";
-    // Note that some software (for example conda) patches TESSDATA_PREFIX
-    // in the binary, so it might be shorter. Recalculate its length.
-    datadir.resize(std::strlen(datadir.c_str()));
-#else
-    datadir = "./";
-#endif /* TESSDATA_PREFIX */
+  // Check environment variable if argv0 is not specified
+  if (const char *tessdata_prefix = std::getenv("TESSDATA_PREFIX")) {
+    std::filesystem::path path(tessdata_prefix);
```
danpla, line 62: getenv() returns null if the variable is not set.

PR #4386, opened 2025-01-25
Fix duplicated IDs in ALTO XML when multiple pages are present

jankal:
When running tesseract on a list of images and validating the alto output using the schema (https://www.loc.gov/alto/v3/alto-3-0.xsd), I had the following error:

```
 Element '{http://www.loc.gov/standards/alto/ns-v3#}ComposedBlock', attribute 'ID': 'cblock_0' is not a valid value of the atomic type '{http://www.loc.gov/standards/alto/ns-v3#}BlockTypeID'.
```

Digging into the type of `BlockTypeID`, I found a `xsd:ID` restriction - meaning the `BlockTypeID` can only exist once in the whole document / is unique per document.

I incorporated the current `page_number` into the IDs thus resolving the issue.

### DIFF
--- a/src/api/altorenderer.cpp
+++ b/src/api/altorenderer.cpp
@@ -51,6 +51,20 @@ static void AddBoxToAlto(const ResultIterator *it, PageIteratorLevel level,
   }
 }
 
+static std::string GetID(const char *prefix, int page_number, int counter) {
+  std::stringstream idstr;
+  // IDs will only have the counter for the first page to keep them consistent
+  // with the IDs assigned before this change was made.
+  // From the second page on, IDs will also contain the page number to make them unique.
+  if (page_number == 0) {
+    idstr << prefix << "_" << counter;
+  } else {
+    idstr << prefix << "_" << page_number << "_" << counter;
+  }
+
+  return idstr.str();
+}
+
 ///
 /// Append the ALTO XML for the beginning of the document
 ///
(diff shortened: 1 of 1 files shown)


stweil:
Thanks for the fix. I just wonder whether we should try to keep the current IDs for the most common use case where only a single page is processed. This could be achieved  for example by
- conditional code
- using counters (bcnt, ...) which are not reset for each additonal image
- using an index of `(page_number - 1) * 1000 + bcnt` ... in the ID strings

jankal:
Thanks for your suggestion @stweil!

I added a function `GetID` to generate the ID. The new `GetID` function keeps the current ID structure for the first page and starts to add the `pag_number` from the second page on.

Disclosure: I haven't written CPP before, so pls check for rookie mistakes :)

src/api/altorenderer.cpp
```diff
@@ -51,6 +51,20 @@ static void AddBoxToAlto(const ResultIterator *it, PageIteratorLevel level,
   }
 }
 
+static std::string GetID(char const * prefix, int page_number, int counter) {
```
stweil, line 54:
```suggestion
static std::string GetID(const char *prefix, int page_number, int counter) {
```

stweil: The original code is technically fine, but my suggested change uses the coding style which is typical for Tesseract code.

stweil:
Thank you again!

PR #4404, opened 2025-03-27
Fix invalid empty interval in punct_stripped() for all-punctuation words

EnodoGH:
punct_stripped() may currently produce invalid empty intervals (where end < start) for all-punctuation words. As a result, EqualIgnoringCaseAndTerminalPunct() crashes in this loop due to an unsigned integer underflow in w1end - w1start:

  for (unsigned i = 0; i < w1end - w1start; i++) {
    if (uchset->to_lower(word1.unichar_id(w1start + i)) !=
        uchset->to_lower(word2.unichar_id(w2start + i))) {
      return false;
    }
  }
  
Bug was introduced with the following commit:

-----------------------------------------------------------
  
Revision: 97048fe3e4813d6c2350a370b915563688123f2c
Author: Stefan Weil <sw@weilnetz.de>
Date: 09/10/2021 20:50:39
Message:
ccstruct: Fix some signed/unsigned compiler warnings

Remove also a local buffer in function REJMAP::print.

-----------------------------------------------------------

There was no unsigned integer underflow before, because the interval was signed.

### DIFF
--- a/src/ccstruct/ratngs.cpp
+++ b/src/ccstruct/ratngs.cpp
@@ -370,7 +370,7 @@ void WERD_CHOICE::punct_stripped(unsigned *start, unsigned *end) const {
   while (*start < length() && unicharset()->get_ispunctuation(unichar_id(*start))) {
     (*start)++;
   }
-  while (*end > 0 && unicharset()->get_ispunctuation(unichar_id(*end - 1))) {
+  while (*end > *start && unicharset()->get_ispunctuation(unichar_id(*end - 1))) {
     (*end)--;
   }
 }


stweil:
@egorpugin, please wait a little bit longer before merging non-trivial pull requests. Otherwise there is not enough time for others to review the changes.

I am not sure that this change is the right solution. `punct_stripped` always could produce intervals where end was smaller than start, and `Dict::valid_bigram` even expects this behaviour which now was avoided.

A safer fix would have been possible in `EqualIgnoringCaseAndTerminalPunct`.

stweil:
@EnodoGH, can you provide a test image which triggers the crash?

EnodoGH:
> @EnodoGH, can you provide a test image which triggers the crash?

I probably could, but it would be highly non-trivial. It's one of slightly fewer than one million images, which our software takes a few days to process. There's a good chance it's a customer image, for which I'll need to get permission before passing it on to you. In that image, there are multiple calls to the Tesseract library, and I'd have to isolate the exact call and the exact data that was passed to it.

I'll try to do all that if it's too hard for you to reproduce the crash from scratch. For now, all I can tell you is that we reproduced the crash in EqualIgnoringCaseAndTerminalPunct() on two different machines. On mine, I managed to identify punct_stripped() as the reason the crash happened, fixed it, and the crash was gone for good on both machines.

Issue #2990, opened 2020-05-22
Error handling: Don't call exit(n) inside libtesseract


amitdo:
https://github.com/tesseract-ocr/tesseract/blob/01535706ecd11a495f9f6e65ee0118b3537a115a/src/textord/strokewidth.cpp#L121-L128

https://github.com/tesseract-ocr/tesseract/blob/2a37f5dd6226156bf32e60d1ae0209a383af4843/src/classify/intproto.cpp#L231-L238

https://github.com/tesseract-ocr/tesseract/blob/912c9978dacf9f7a7cc97ba9a906b40e3df8fc09/src/dict/permdawg.cpp#L111-L121

stweil:
Are there examples which trigger those exits? That would help testing fixes this issue.

amitdo:
No, I don't have examples which trigger those exits.

jbarlow83:
@stweil Any call to `exit()` from code intended for use in a library is a bug - it doesn't really matter how to trigger them. For a GUI that binds libtesseract, the application will be nuked from orbit by `exit()` and probably won't be able to get an error message to user.

Ideally Tesseract would just throw an `std::exception` because those exits are really deep in the call stack. Are exceptions still forbidden? The standard library can and does throw exceptions, they're already de facto in use. 

If we can't use exceptions, the thing to do is define a function pointer that defaults to `exit()`...
```c++
void (*tesseract_exit)(int error_code) = exit;
```
...and add an API call that lets the application provide its own exit handler. That way the application can throw an exception or do whatever makes sense in context.

amitdo:
The ban on exception came from Google coding rules. Since Ray leaved this project  years ago, we don't have to be limited by their internal coding standards. 

@stweil, should we use exceptions instead of these exit() calls?

jbarlow83:
C++ exceptions are zero cost when the exception does not occur.

bertsky:
> Calling `exit` or maybe better `abort` in case of an internal software bug is fine for me.

Calling `abort()` **in a library** is itself as much a bug as the original `exit()` formulation. The application will be gone along with the library and can not even do anything about it.

This issue is also much bigger IMO: it's not just these few places, but every `ASSERT_HOST` (any there are plenty) delegates to `abort()` as well. (Or, if compiled with `-g`, even to a segfault event.): https://github.com/tesseract-ocr/tesseract/blob/main/src/ccutil/errcode.cpp#L72-L83

[...]

PR #4420, opened 2025-05-07
ERRCODE::error: throw std::runtime_error instead of abort()

bertsky:
Related to https://github.com/tesseract-ocr/tesseract/issues/2990 (although it does not replace the `exit()` calls – yet).

@kba will provide an easily reproducable example of such failing assertions. (My current case is more special / harder to reconstruct.)

But regardless of the concrete issues that _cause_ these failed assertions, which of course need to be addressed themselves, this makes Tesseract usable productively as a library. (You can catch these exceptions.)

(No idea what's the matter with CI here, btw.)

### DIFF
--- a/src/ccutil/errcode.cpp
+++ b/src/ccutil/errcode.cpp
@@ -70,17 +70,11 @@ void ERRCODE::error(         // handle error
       return; // report only
     case TESSEXIT:
     case ABORT:
-#if !defined(NDEBUG)
-      // Create a deliberate abnormal exit as the stack trace is more useful
-      // that way. This is done only in debug builds, because the
-      // error message "segmentation fault" confuses most normal users.
-#  if defined(__GNUC__)
-      __builtin_trap();
-#  else
-      *reinterpret_cast<int *>(0) = 0;
-#  endif
-#endif
-      abort();
+      // This used to trigger a segfault or abort();
+      // However, at least for library use, only exceptions should be acceptable.
+      // Even in the standalone application case, exceptions are better,
+      // because the default handler will print the message along with the stack trace.
+      throw std::runtime_error(msg.str());
     default:
       BADERRACTION.error("error", ABORT);
   }


egorpugin (approved):

stweil:
I have several comments:

1. What will be the effect of throwing exceptions for C applications and other applications which depend on the C API?
2. This PR changes the behaviour of the `tesseract` executable for fatal errors. All releases until now could trigger a core dump which could be analyzed post mortem. It also returns a specific exit code. The new code raises an exception which is caught in main(), prints a message to stderr and returns the exit code 1.

So these changes are API changes. Maybe we should postpone.them to a release 6.

stweil (changes requested):
I think it is very important that `tesseract` runtime errors (like for example those detected by assertions) abort the program and create a usable stack trace in core dumps or debug sessions.

My use case is a mass production of OCR with thousands of `tesseract` runs. Ideally all runs work without any problem. But if there are single runs with runtime issues, such runs should be able to create core dumps which can be examined later.

With this PR my use case no longer works.

stweil:
>     1. What will be the effect of throwing exceptions for C applications and other applications which depend on the C API?

Meanwhile I did some tests on macOS and Linux. By default, the std::runtime_error exception is called by the C++ runtime library which calls abort(). So applications which don't catch C++ exceptions abort like before which is desired.

src/ccutil/errcode.cpp
```diff
@@ -70,17 +70,11 @@ void ERRCODE::error(         // handle error
...
-      __builtin_trap();
-#  else
-      *reinterpret_cast<int *>(0) = 0;
-#  endif
-#endif
-      abort();
+      // This used to trigger a segfault or abort();
+      // However, at least for library use, only exceptions should be acceptable.
+      // Even in the standalone application case, exceptions are better,
+      // because the default handler will print the message along with the stack trace.
```
stweil, line 76: Will it? It does not print a stack trace in my test on macOS.

stweil, line 76:
```suggestion
      // Abort unless std::runtime_error is caught by the application.
```

bertsky:
> 1. What will be the effect of throwing exceptions for C applications and other applications which depend on the C API?

Frankly, I don't know. I suppose we could differentiate C and C++ behaviour via ifdefs, and revert to the old (`abort`) implementation in plain C.

> 2. This PR changes the behaviour of the `tesseract` executable for fatal errors. All releases until now could trigger a core dump which could be analyzed post mortem. It also returns a specific exit code (134 for abort, 139 for SIGSEGV). The new code raises an exception which is caught in main(), prints a message to stderr and returns the exit code 1.

If you want that for the standalone CLI, I guess that would simply be doable by installing the right kind of exception handler somewhere in tesseractmain.cpp, or even catch exceptions explicitly and call `abort()` that way.

But I'm surprised by your statements regarding the new behaviour. For me the default handler on the CLI still delegates to `abort()`, which still writes a core dump (as long as `ulimit -c` is set appropriately).

> So these changes are API changes. Maybe we should postpone.them to a release 6.

That is an awkward position IMO. The API itself does not even change here. The only change is in the case of errors – where prior to the change, nothing could be done anymore. The changes now make errors actionable (via std exceptions). 

So in a word, nothing that used to work breaks here, this only fixes what happens if things don't work.

src/ccutil/errcode.cpp
bertsky: On Linux, I get:
```
ELIST_ITERATOR::forward:Error:List would have returned a nullptr data pointer
terminate called after throwing an instance of 'std::runtime_error'
  what():  ELIST_ITERATOR::forward:Error:List would have returned a nullptr data pointer
Abgebrochen (Speicherabzug geschrieben)
```

stweil: There used to be comments about the history of some code parts in the past, but most of them are removed in the recent code. I don't think that new comments of this kind are helpful.

stweil: That's not a stack trace. And please rebase your code with all recent updates from the main branch. Then you will no longer get a core dump.

bertsky:
> That's the normal nature of fatal errors. As their name says, they are not vital. 

That statement does not make any sense to me.

> Even without the few abort() calls in the release code of Tesseract, there remain numerous ways to cause other fatal runtime errors (segmentation faults, floating point exceptions and more).

Sure, there will always remain other levels of failure, enforced by the OS. Again, this is besides the point. We are talking about preventable failures here – errors which the application could recover from (for example by skipping the task, perhaps re-initializing Tesseract). Which is precisely what my Python use-case (error handling of ocrd-tesserocr) showed.

egorpugin:
@stweil 

> My use case is a mass production of OCR with thousands of `tesseract` runs.

I added global `catch(std::exception)` and `catch (...)` into tesseract main() not so long ago.

If you need a binary without them to make exceptions uncaught, considering your huge volume, it is probably worth it for you to use tess as library (libtess) and make custom binary wrapper without catching exceptions.

stweil:
> I added global `catch(std::exception)` and `catch (...)` into tesseract main() not so long ago.

I know. That's the reason why @bertsky's code now only prints an error message with `tesseract` instead of the desired aborting. His PR is based on an older main branch without your try / catch block, so he could not notice this.

stweil:
> @kba will provide an easily reproducable example of such failing assertions. (My current case is more special / harder to reconstruct.)

> @kba, @bertsky, please open issues for reproducable examples of failing assertions.

Did I miss the examples? @bertsky, could you please send my a link?

PR #4423, opened 2025-05-25
Use links to the git history and online release notes instead of local ChangeLog

stweil:


stweil:
See the [related discussion](https://github.com/tesseract-ocr/tesseract/issues/3673#issuecomment-2903623895).

amitdo (approved):

PR #4451, opened 2025-08-10
Fixed the GetUTF8Text method to return nullptr instead of asserting

zdenop:
### Description

This PR fixes a crash in the `GetUTF8Text` method by returning `nullptr` instead of asserting when `best_choice` is `nullptr`.

### Background

The issue was originally reported in [sirfz/tesserocr#324](https://github.com/sirfz/tesserocr/issues/324), where a crash/assertion failure occurred when passing an empty image to tesserocr/Tesseract. The error message was:

```
best_choice != nullptr:Error:Assert failed: in file F:\Projects\Community\tesseract\src\ccmain\ltrresultiterator.cpp, line 52
```

While reproducing the issue with a test C++ program, I encountered a similar crash in another location:

```
it_->word()->best_choice != nullptr:Error:Assert failed: in file F:\Projects\Community\tesseract\src\ccmain\resultiterator.cpp, line 709
```

### Changes

* Modified `GetUTF8Text` to safely return `nullptr` if `best_choice` is `nullptr`, instead of triggering an assertion failure.

### Additional Notes

* I tested this with Warp2, which suggested updating multiple instances of `ASSERT_HOST(best_choice != nullptr);`. This PR addresses the crash in the context of `GetUTF8Text`.
* There are still 3 other occurrences of this assertion in the following files:

  * `src/ccmain/recogtraining.cpp`
  * `src/ccstruct/pageres.cpp`

  It's unclear if those should also be replaced, as they may be used in different contexts (e.g., training or layout analysis). Further review is recommended before changing them.

### Reproduction

Here is a minimal test program that reproduces the crash (when using an empty image):

```cpp
//  test_ResultIterator.cpp

#include <iostream>

int main(int argc, char *argv[]) {

  if (api->Init(NULL, "chi_sim")) {
    exit(1);
  }

  // Create a small empty/blank image (1x1 pixel)
  Pix* pix = pixCreate(100, 100, 8);
  if (pix == nullptr) {
    std::cerr << "Could not create empty image." << std::endl;
    return 1;
  }
  api->SetImage(pix);

  api->SetPageSegMode(tesseract::PSM_SINGLE_CHAR);
  api->Recognize(0);
  int out_offset;
  float out_slope;
  if (api->GetTextDirection(&out_offset, &out_slope)) {
    std::cout << "out_offset: " << out_offset << std::endl;
    std::cout << "out_slope: " << out_slope << std::endl;
  } else {
    std::cout << "GetTextDirection failed." << std::endl;
  }
  // Get the iterator and try to imitate relevant part of MapWordConfidences function
  tesseract::ResultIterator* ri = api->GetIterator();
  tesseract::PageIteratorLevel level = tesseract::RIL_WORD;
  if (ri != nullptr) {
    do {
      const char* word = ri->GetUTF8Text(level);
      auto word_confidence = ri->Confidence(level);
      std::cout << "Word confidence(" << word << "): " << word_confidence << std::endl;
    } while (ri->Next(level));
  }

  delete ri;
  api->End();
  delete api;
  pixDestroy(&pix);
  std::cout << "Test completed successfully - no crash!" << std::endl;
  return 0;
}

```

### DIFF
--- a/src/ccmain/ltrresultiterator.cpp
+++ b/src/ccmain/ltrresultiterator.cpp
@@ -49,7 +49,9 @@ char *LTRResultIterator::GetUTF8Text(PageIteratorLevel level) const {
   std::string text;
   PAGE_RES_IT res_it(*it_);
   WERD_CHOICE *best_choice = res_it.word()->best_choice;
-  ASSERT_HOST(best_choice != nullptr);
+  if (best_choice == nullptr) {
+    return nullptr; // No recognition results available
+  }
   if (level == RIL_SYMBOL) {
     text = res_it.word()->BestUTF8(blob_index_, false);
   } else if (level == RIL_WORD) {
@@ -61,7 +63,9 @@ char *LTRResultIterator::GetUTF8Text(PageIteratorLevel level) const {
       do {            // for each text line in a paragraph
         do {          // for each word in a text line
           best_choice = res_it.word()->best_choice;
-          ASSERT_HOST(best_choice != nullptr);
+          if (best_choice == nullptr) {
+            break; // Skip words without recognition results
+          }
           text += best_choice->unichar_string();
           text += " ";
           res_it.forward();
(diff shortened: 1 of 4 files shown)


egorpugin (approved):

src/ccmain/ltrresultiterator.cpp
```diff
@@ -61,7 +63,9 @@ char *LTRResultIterator::GetUTF8Text(PageIteratorLevel level) const {
       do {            // for each text line in a paragraph
         do {          // for each word in a text line
           best_choice = res_it.word()->best_choice;
-          ASSERT_HOST(best_choice != nullptr);
+          if (best_choice == nullptr) {
```
stweil, line 66: Is there an example which triggers the assertion here? I am not sure whether the `break` is the right solution here. Maybe the loop should continue at line 71.

src/ccmain/ltrresultiterator.cpp
```diff
@@ -104,37 +108,47 @@ float LTRResultIterator::Confidence(PageIteratorLevel level) const {
     case RIL_BLOCK:
       do {
         best_choice = res_it.word()->best_choice;
-        mean_certainty += best_choice->certainty();
-        ++certainty_count;
+        if (best_choice != nullptr) {
```
stweil, line 111: Are there situations where the condition is false?

amitdo:
In your test code you didn't test that the block is not empty.

Try this:

 ```
 tesseract::ResultIterator* ri = api->GetIterator();
if (ri->Empty(RIL_BLOCK) {
  // print error message and then exit()
  .... 
}
else (...)
```

amitdo:
Hi,

Currently, I don't have time to review PRs.

PR #4452, opened 2025-08-10
Cmake optimization with warp2

### LINKED ISSUE #4442: RFC: Don't use GLOB in CMake files
https://cmake.org/cmake/help/latest/command/file.html#glob

>**Note** We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

https://github.com/search?q=repo%3Atesseract-ocr%2Ftesseract+glob+language%3Acmake&type=code

zdenop:
* Don't use GLOB in CMake files - fixes #4442
* Optimization to improve build speed:
  * Precompiled Headers (PCH) - Enabled by default via ENABLE_PRECOMPILED_HEADERS=ON 
  * Unity Builds (Jumbo Builds) - Disabled by default via ENABLE_UNITY_BUILD=OFF. Can reduce compilation time by 30-60% but may increase memory usage 
  * ccache Integration - Enabled by default via ENABLE_CCACHE=ON 
  * Ninja Job Pools - Enabled by default via ENABLE_NINJA_POOL=ON
* fix several CMake macos issues

### Usage Examples

```
cmake --build build  -G Ninja -DCMAKE_BUILD_TYPE=Release -DENABLE_UNITY_BUILD=ON
cmake --build build --config Release --target install
```

### DIFF
--- a/.github/workflows/cmake-win64.yml
+++ b/.github/workflows/cmake-win64.yml
@@ -112,7 +112,7 @@ jobs:
       - name: Build and Install tesseract
         shell: cmd
         run: |
-             cmake -Bbuild -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_PREFIX_PATH=${{env.ILOC}} -DCMAKE_INSTALL_PREFIX=${{env.ILOC}} -DSW_BUILD=OFF -DBUILD_SHARED_LIBS=ON -DENABLE_LTO=ON -DBUILD_TRAINING_TOOLS=OFF -DFAST_FLOAT=ON -DGRAPHICS_DISABLED=ON -DOPENMP_BUILD=OFF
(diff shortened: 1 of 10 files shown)


egorpugin:
What is warp2?

egorpugin (commented):
It is better to put any code changes outside of this CMake PR.

egorpugin:
What is the purpose of all this?

Someone says GLOBs = bad?
I say GLOBs = good.

What about the rest?
Need of unity/pch/ccache/job pools?

amitdo:
@zdenop,

Since this PR was merged, the macos+GCC CI job fails.

Could you fix the issue?

PR #4465, opened 2025-10-10
Fix crash when tessdata directory doesn't exist

kdt523:
Solution
Added two-layer protection in `addAvailableLanguages()`:

1. Preventive check: Verify directory exists before attempting iteration
2. Exception handling: Catch filesystem errors during traversal

Changes
- Add `std::filesystem::exists()` and `std::filesystem::is_directory()` checks
- Wrap `recursive_directory_iterator` in try-catch block
- Function now returns gracefully with empty language list instead of crashing

 Testing
- Verified fix handles missing directories without crashing
- Tested edge cases (empty paths, permission denied scenarios)

Impact
- After: Graceful degradation - returns empty language list and continues running

This fix improves robustness for users with incomplete Tesseract installations.

Contributing to Hacktoberfest 2025

### DIFF
--- a/src/api/baseapi.cpp
+++ b/src/api/baseapi.cpp
@@ -146,11 +146,6 @@ static void ExtractFontName(const char* filename, std::string* fontname) {
  */
 static void addAvailableLanguages(const std::string &datadir,
                                   std::vector<std::string> *langs) {
-  // Check if directory exists before attempting to iterate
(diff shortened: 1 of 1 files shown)


stweil (changes requested):

stweil:
I think this PR can be closed, see previous comment from @egorpugin.

PR #4482, opened 2025-12-02
Remove unused std::thread member from DocumentData class

ihhub:
The presence of std::thread::join() causes extra dependencies on some exotic platforms like AIX. Moreover, having a member class that does nothing is waste of resources.

The corresponding method was renamed to reflect its nature as well as the comment related to the method.

### DIFF
--- a/src/ccstruct/imagedata.cpp
+++ b/src/ccstruct/imagedata.cpp
@@ -36,6 +36,7 @@
 
 #include <cinttypes>    // for PRId64
 #include <fstream>      // for std::ifstream
+#include <thread>       // for std::this_thread
 
 namespace tesseract {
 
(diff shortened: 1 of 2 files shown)


ihhub:
Closing this pull request as it has been enough time to review it over half a year span.

stweil:
As far as I see this PR changes an asynchronous background task (loading of documents) into a synchronous action. I am afraid that this would slow down the training process. This can be rather bad (depending on the additional time required for a training).

stweil:
> Moreover, having a member class that does nothing is waste of resources.

I don't see a member class which does nothing. Which one was meant?

ihhub:
> As far as I see this PR changes an asynchronous background task (loading of documents) into a synchronous action. I am afraid that this would slow down the training process. This can be rather bad (depending on the additional time required for a training).

Can you please point out in the code where do you spawn a new thread? `std::thread thread;` has a default constructor and has no associated function with it.

stweil:
@ihhub, you are right, the current code no longer spawns a new thread. I removed that myself in commit 73a1bfc4e8 some years ago (bug fix), but did not remove the thread code which was no longer used. So your changes are basically correct. Could you please address the feedback from Copilot? And I think that this_thread is also a relict which might be removed.

PR #4514, opened 2026-02-19
Add environment path management for installation

eduardomozart:
## Description
This PR enhances the Windows installer by automatically adding the Tesseract installation directory to the system or user `PATH` environment variable. This allows users and automated scripts to run `tesseract.exe` from any command prompt immediately after installation without manual configuration.

### Motivation
Standardizing the `PATH` configuration during installation improves the out-of-the-box experience and simplifies automated, unattended infrastructure deployments. Previously, users had to manually edit their environment variables after installing.

### Key Changes
- **Dynamic Privilege Handling:** The script now checks both `$MultiUser.InstallMode` and the actual account privileges (`UserInfo::GetAccountType`). It safely writes to `HKLM` if running as an Administrator (All Users) or falls back to `HKCU` for standard users (Current User).
- **Safe Environment Variable Manipulation:** Integrated the `EnVar` plugin to modify the `PATH`. This is highly recommended over native NSIS registry commands to avoid truncating existing `PATH` strings that exceed 1024 characters.
- **Clean Uninstallation:** Added logic to the uninstaller to dynamically check the installation mode and safely remove the Tesseract directory from the `PATH`.

## Notes for Reviewers
⚠️ **Build Dependency Update:** Compiling `tesseract.nsi` now requires the [EnVar plugin](https://nsis.sourceforge.io/EnVar_plug-in) to be present in the NSIS plugins directory.

### DIFF
--- a/nsis/Makefile.am
+++ b/nsis/Makefile.am
@@ -12,11 +12,15 @@ Plugins/x86-unicode/INetC.dll:
 	curl -OsS https://nsis.sourceforge.io/mediawiki/images/c/c9/Inetc.zip
 	unzip Inetc.zip $@
 
+Plugins/x86-unicode/EnVar.dll:
+	curl -OsS https://nsis.sourceforge.io/mediawiki/images/7/7f/EnVar_plugin.zip
+	unzip EnVar_plugin.zip $@
+
 winpath.exe: winpath.cpp
 	x86_64-w64-mingw32-g++ -Os -o $@ $<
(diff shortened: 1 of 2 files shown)


stweil:
It looks like the new UI texts with translations are still missing.

eduardomozart:
Hello @stweil,
All Copilot suggestions have been addressed. Regarding the missing translation strings, the English string has been committed to #4516 (my bad), but the Portuguese string has been added.

PR #4517, opened 2026-02-19
Correct mutex call to prevent multiple instances

eduardomozart:
By dropping the "A" from ``CreateMutex``, the ``System`` plugin natively maps the call to ``CreateMutexW`` because the installer is running in Unicode mode. It will properly register the string "Tesseract-OCR" in Windows. When the second instance launches, it will hit that exact same string, receive error ``183``, and gracefully abort before corrupting the installation.

image


eduardomozart:
Added the suggested fix from Copilot to the code. Still working as the screenshot.

Issue #4046, opened 2023-04-05
ALTO output - add support for LANG attribute in TextBlock/TextLine elements


filak:
### Your Feature Request

It might be relatively simple to do this by looking at the hocrrenderrer https://github.com/tesseract-ocr/tesseract/blob/5f297dc0b8b500d57b7c073f4457e74ee537819f/src/api/hocrrenderer.cpp#L243

```
paragraph_lang = res_it->WordRecognitionLanguage();
if (paragraph_lang) {
  hocr_str << " lang='" << paragraph_lang << "'";
}
```

https://github.com/tesseract-ocr/tesseract/blob/5f297dc0b8b500d57b7c073f4457e74ee537819f/src/api/hocrrenderer.cpp#L302

```
const char *lang = res_it->WordRecognitionLanguage();
if (lang && (!paragraph_lang || strcmp(lang, paragraph_lang))) {
  hocr_str << " lang='" << lang << "'";
}
```

It could be adapted in altorenderer

https://github.com/tesseract-ocr/tesseract/blob/424b17f997363670d187f42c43408c472fe55053/src/api/altorenderer.cpp#L215

ie.
```
    if (res_it->IsAtBeginningOf(RIL_PARA)) {
      alto_str << "\t\t\t\t\t<TextBlock ID=\"block_" << tcnt << "\"";
      AddBoxToAlto(res_it, RIL_PARA, alto_str);
      paragraph_lang = res_it->WordRecognitionLanguage();
      if (paragraph_lang) {
        alto_str << " LANG='" << paragraph_lang << "'";
      }
      alto_str << "\n";
    }

    if (res_it->IsAtBeginningOf(RIL_TEXTLINE)) {
      alto_str << "\t\t\t\t\t\t<TextLine ID=\"line_" << lcnt << "\"";
      AddBoxToAlto(res_it, RIL_TEXTLINE, alto_str);
      const char *lang = res_it->WordRecognitionLanguage();
      if (lang && (!paragraph_lang || strcmp(lang, paragraph_lang))) {
        alto_str << " LANG='" << lang << "'";
      }
      alto_str << "\n";
    }
```

The lang codes shall be converted from Tesseract codes to standard 2-letter codes.

A mapping structure needs to be created (I have done the mapping before [codes_lookup.xml](https://github.com/filak/hOCR-to-ALTO/blob/master/codes_lookup.xml) but it definitely must be updated) which can be used in a function ie. 

[...]

PR #4531, opened 2026-03-20
Add languages parameter to hOCR output metadata

### LINKED ISSUE #4455: write languages parameter value to hocr output file
### expected

write languages parameter value to hocr output file

```html
<html ...>
 <head>
  <!-- ... -->
  <meta name='ocr-languages' content='deu+eng+rus'/>
 </head>
```

### why

this is useful for OCR proofreading with [hocr-editors](https://github.com/zacharywhitley/awesome-ocr/pull/7)
to run tesseract again on selected regions with the original tesseract arguments

### alternative

write all tesseract arguments to hocr output file
but this is harder to parse

```html
<html ...>
 <head>
  <!-- ... -->
  <meta name='ocr-arguments' content='tesseract src.jpg - -l deu+eng+rus hocr'/>
 </head>
```

### workaround

guess the languages parameter value
from `lang='...'` attributes in the hocr output file

parse the main language from
`<p class='ocr_par' id='[^']+' lang='([^']+)'`
and parse extra languages from
`<span class='ocrx_word' id='[^']+' title='[^']+' lang='([^']+)'`

but this workaround can fail and return a wrong order of languages

> parse

yeah i know, "parsing" xml with regex is bad
this is just an example, in a real app i would use a proper xml parser

### example

tesseract was called like
`tesseract src.jpg - -l deu+eng+rus hocr >dst.hocr`
then the main language is `deu`
and the extra languages are `eng` and `rus`

mvanhorn:
## Summary

Adds an `ocr-langs` meta tag to the hOCR HTML header containing the `-l` languages parameter (e.g. `eng+fra`).

## Problem

When processing images with `tesseract -l eng+fra image.tif output hocr`, the hOCR output includes `ocr-system` and `ocr-capabilities` meta tags but not the languages used for recognition. Downstream tools processing hOCR files have no way to determine which language models produced the results. Requested in #4455, related to #4046.

## Changes

- `include/tesseract/renderer.h`: Added `SetInputLanguages()` method and `input_languages_` member to `TessHOcrRenderer`
- `src/api/hocrrenderer.cpp`: Implemented `SetInputLanguages()` and emit `<meta name='ocr-langs' content='...' />` in `BeginDocumentHandler()` when languages are set
- `src/tesseract.cpp`: Call `renderer->SetInputLanguages(api.GetInitLanguagesAsString())` when creating the hOCR renderer (line 519)

The existing `GetInitLanguagesAsString()` API (`baseapi.cpp:372`) already returns the `-l` value - it just wasn't being passed to the hOCR renderer.

Example output with `-l eng+fra`:
```html
<meta name='ocr-system' content='tesseract 5.5.1' />
<meta name='ocr-capabilities' content='ocr_page ocr_carea ...'/>
<meta name='ocr-langs' content='eng+fra' />
```

## Testing

The meta tag is only emitted when `SetInputLanguages()` is called with a non-null, non-empty string. The C API (`TessHOcrRendererCreate`/`TessHOcrRendererCreate2`) is unchanged - the meta tag won't appear unless `SetInputLanguages` is explicitly called.

Fixes #4455

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/include/tesseract/renderer.h
+++ b/include/tesseract/renderer.h
@@ -173,6 +173,8 @@ public:
   explicit TessHOcrRenderer(const char *outputbase, bool font_info);
   explicit TessHOcrRenderer(const char *outputbase);
 
+  void SetInputLanguages(const char *languages);
+
 protected:
   bool BeginDocumentHandler() override;
   bool AddImageHandler(TessBaseAPI *api) override;
@@ -180,6 +182,7 @@ protected:
 
 private:
   bool font_info_; // whether to print font information
+  std::string input_languages_;
 };
 
 /**
--- a/src/api/hocrrenderer.cpp
+++ b/src/api/hocrrenderer.cpp
@@ -477,6 +477,12 @@ TessHOcrRenderer::TessHOcrRenderer(const char *outputbase, bool font_info)
   font_info_ = font_info;
 }
 
+void TessHOcrRenderer::SetInputLanguages(const char *languages) {
+  if (languages) {
+    input_languages_ = languages;
+  }
+}
+
 bool TessHOcrRenderer::BeginDocumentHandler() {
   AppendString(
       "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
@@ -496,8 +502,13 @@ bool TessHOcrRenderer::BeginDocumentHandler() {
   if (font_info_) {
     AppendString(" ocrp_font ocrp_fsize");
   }
+  AppendString("'/>\n");
+  if (!input_languages_.empty()) {
+    AppendString("  <meta name='ocr-langs' content='");
+    AppendString(input_languages_.c_str());
+    AppendString("' />\n");
+  }
   AppendString(
-      "'/>\n"
       " </head>\n"
       " <body>\n");
 
--- a/src/api/hocrrenderer.cpp
+++ b/src/api/hocrrenderer.cpp
@@ -481,13 +481,15 @@ TessHOcrRenderer::TessHOcrRenderer(const char *outputbase, bool font_info)
 }
 
 void TessHOcrRenderer::SetInputLanguages(const char *languages) {
-  if (languages) {
+  if (languages && languages[0] != '\0') {
     input_languages_ = languages;
+  } else {
+    input_languages_.clear();
   }
 }
 
 static const std::unordered_map<std::string, std::string> &Iso639Map() {
-  static const auto *map = new std::unordered_map<std::string, std::string>{
+  static const std::unordered_map<std::string, std::string> map{
       {"afr", "af"}, {"amh", "am"}, {"ara", "ar"}, {"asm", "as"},
       {"aze", "az"}, {"bel", "be"}, {"ben", "bn"}, {"bod", "bo"},
       {"bos", "bs"}, {"bre", "br"}, {"bul", "bg"}, {"cat", "ca"},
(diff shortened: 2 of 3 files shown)


egorpugin (approved):

amitdo:
https://github.com/kba/hocr-spec

https://kba.github.io/hocr-spec/1.2/


[ocr-langs](https://kba.github.io/hocr-spec/1.2/#propdef-ocr-langs)

[ocr-scripts](https://kba.github.io/hocr-spec/1.2/#propdef-ocr-scripts)

[metadata example](https://kba.github.io/hocr-spec/1.2/#metadata-example)

stweil (changes requested):
hOCR expects [ISO 639-1](https://www.loc.gov/standards/iso639-2/php/code_list.php) codes, so this pull request requires more work for languages.

In addition, Tesseract's `-l` parameter can also get script models which are not language specific and which require a different handling.

amitdo:
Your implementation is currently not compatible with the spec.

* "eng" -> "en"
* "fra -> "fr"
* "eng+fra" -> "eng fr"

You also need to take into account the script [models](https://github.com/tesseract-ocr/tessdata_best/tree/main/script), e.g. Latin, Greek, Hebrew

amitdo:
@stweil, I didn't see your comment. I basically wrote the same thing roughly in parallel time.... :-)

mvanhorn:
Pushed a fix. Languages are now converted to ISO 639-1 codes (space-separated), and script models emit a separate `ocr-scripts` meta tag. Covers 90+ language mappings and all 37 script models from tessdata_best. Codes without an ISO 639-1 equivalent pass through as-is.

Example with `-l eng+fra+Latin`:
```html
<meta name='ocr-langs' content='en fr' />
<meta name='ocr-scripts' content='Latin' />
```

stweil:
@copilot, please review this pull request. It currently does not handle more complex model names like for example `deu_latf` or `chi_tra_vert`. Suggest also a better solution for arbitrary model names like `german_print`.

mvanhorn:
Good point about compound model names. The current implementation does exact lookup, so `deu_latf` passes through unmapped instead of resolving to `de`.

I'll add prefix-based fallback: strip trailing `_suffix` parts progressively and retry shorter keys. `deu_latf` -> try `deu` -> "de". `chi_tra_vert` -> try `chi_tra` -> "zh".

For truly arbitrary names like `german_print` where no prefix matches, they pass through as-is. The spec doesn't define a mapping for custom model names, and guessing would be worse than emitting the raw identifier.

Also addressing Copilot's suggestions in the same push:
- Switch from `new` to function-local `static const` for Iso639Map/ScriptNames
- Clear `input_languages_` on nullptr
- Add `HOcrEscape()` for meta tag content
- Add unit tests

Re ABI: the new `std::string` member does change class layout. Happy to move it behind an opaque pointer if that's a concern - let me know.

mvanhorn:
Addressed in 43706b8e:

- Compound model names now resolve via prefix fallback: `deu_latf` -> try `deu` -> "de", `chi_tra_vert` -> try `chi_tra` -> "zh". Arbitrary names like `german_print` pass through as-is since no ISO 639-1 mapping exists for custom models.
- Switched `Iso639Map`/`ScriptNames` from heap-allocated `new` to function-local `static const`
- `SetInputLanguages(nullptr)` now clears stale metadata
- Added `HOcrEscape()` for meta tag content

Still need to add unit tests - will push those next.

mvanhorn:
Addressed the Copilot findings. Items 1-3 and 6 were already fixed in 43706b8e (Copilot ran against the prior revision):

- Static function-local maps (not `new`/leak) - already done
- `SetInputLanguages(nullptr)` clears the string - already done
- `HOcrEscape()` applied to meta tag content - already done

On the remaining items:

**ABI (item 4):** Fair point. `input_languages_` adds a `std::string` member to an exported class. If tesseract guarantees ABI stability across patch releases, this should go behind a pimpl. Happy to refactor if maintainers prefer that approach.

**Tests (item 5):** Will add a test exercising `BeginDocumentHandler()` with and without languages set.

**Raw value vs ISO 639-1 (item 7):** The hOCR spec requires ISO 639-1 codes in `ocr-langs` (per the spec @amitdo linked). @stweil and @amitdo both confirmed this direction. The conversion is intentional and spec-compliant.

PR #4551, opened 2026-05-08
Add sketch-sheet prototype generator, ingest pipeline, and launcher.

cfviz01:
Create a standalone Python workflow to generate printable sketch sheets, ingest photographed pages into segmented cells, and provide a clickable Finder launcher for quick local use.


amitdo:
You can host this code in your own repo.

PR #4559, opened 2026-06-03
autotools: Fix linker warning on macOS

stweil:


stweil:
See [this](https://github.com/tesseract-ocr/tesseract/actions/runs/26847606628/job/79171194037#step:7:225) and more of the linker warnings in the CI for macOS.

PR #4561, opened 2026-06-04
Modernize code

stweil:


egorpugin (approved):

egorpugin:
Further notes:

- `DISTRIBUTION::normal` uppercase enums can be set to just title case `Distribution::normal`
- I consider `static_cast<>` as very low prio. C style code is ok too `(int)var`. (Too boring to change them all.)

@stweil

amitdo:
https://github.com/tesseract-ocr/tesseract/blob/56e873189891ba28d9b009445c2f65e0ae5c0496/unittest/util/utf8/unicodetext.cc#L23-L25

Can we remove these commented includes?

stweil:
> Can we remove these commented includes?

This is now done in PR #4568.

PR #4576, opened 2026-07-02
Handle send() failures in SVNetwork::Flush

Ramya-9353:
Handle `send()` failures in `SVNetwork::Flush()`.

This change avoids removing data from the outgoing buffer when `send()` fails and retries interrupted `send()` calls (`EINTR`) on POSIX systems.

### DIFF
--- a/src/viewer/svutil.cpp
+++ b/src/viewer/svutil.cpp
@@ -168,8 +168,23 @@ void SVNetwork::Send(const char *msg) {
 void SVNetwork::Flush() {
   std::lock_guard<std::mutex> guard(mutex_send_);
   while (!msg_buffer_out_.empty()) {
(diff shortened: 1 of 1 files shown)


stweil (changes requested):
@Ramya-9353, thank you for your contribution. Please have a look on the review comments from Copilot.

stweil (commented):
I noticed that your author name is "Ramya-9353". If possible, you should prefer your real name to get the correct attribution for your commits.

src/viewer/svutil.cpp
stweil, line 29:
```suggestion
#include <cerrno>
#include <cstdio>
```

stweil (approved):
The fix is good, thank you! Some minor improvements can be applied later.

PR #4585, opened 2026-07-18
Small code improvements

stweil:


src/wordrec/plotedges.cpp
```diff
@@ -82,8 +82,8 @@ void draw_blob_edges(TBLOB *blob) {
  **********************************************************************/
 void mark_outline(EDGEPT *edgept) { /* Start of point list */
   auto window = edge_window;
-  float x = edgept->pos.x;
-  float y = edgept->pos.y;
+  int x = edgept->pos.x;
+  int y = edgept->pos.y;
```
egorpugin, line 86: auto? What is type of `edgept->pos.{x,y};`?

bchelimo706-cloud (approved):

src/wordrec/plotedges.cpp
stweil: It is TDimension which can either be int16_t (default) or int32_t. The function which get x, y as arguments expect int, so I think this is better than auto here.

PR #4600, opened 2026-08-11
Add OCR text aggregation for multiple boxes

### LINKED ISSUE #4366: Recognize multiple rectangles on a page without clearing results inbetween
### Your Feature Request

Using SetRectangle and Recognize in a loop will clear what has been found between each run resulting in only the last result being returned. It would be nice if it would nice it was possible to keep all results by telling it not to clear. Alternatively give it a range of rectangles and not just one. The latter is probably faster, since overhead can be reduced.

In turn when inserting result via AddImage in a PDFRenderer, the page is automatically advanced to the next, so only one result can be inserted per page. It would also be nice if at least it was possible to tell it not/or when to advance, or which page you are adding to, thereby build the pages from OCR results.

These features would help me achieve what I estimate is 80% of the time spent, or put it in another way 5x speed. Unsure about memory saved, but might help here also.

tianrking:
Fixes #4366

### Summary

Add `TessBaseAPI::GetUTF8TextForBoxes` and the matching C API entry point. The method accepts a Leptonica `Boxa`, reuses the configured image and OCR engine, and returns the recognized UTF-8 text in rectangle order. This gives callers a single production API for multi-region OCR instead of duplicating the SetRectangle/GetUTF8Text loop.

### Validation

- Built `libtesseract` with CMake (C++20, legacy engine disabled).
- Ran a smoke program against `HelloGoogle.tif` and `eng.traineddata`; two identical boxes produced exactly two concatenated results.
- Added `TesseractTest.UTF8TextForBoxes` covering the public C++ API.
- `git diff --check` passes.

### DIFF
--- a/include/tesseract/baseapi.h
+++ b/include/tesseract/baseapi.h
@@ -284,6 +284,17 @@ public:
                       int height);
 
   /**
+   * Recognize each rectangle in a Boxa and concatenate the UTF-8 results in
+   * box order. SetImage must be called before this method. The caller owns
+   * the returned string and must free it with delete[].
+   *
+   * This keeps the image and initialized engine shared while processing a
+   * collection of regions, which avoids duplicating the rectangle loop in
+   * applications that need OCR for multiple regions on one image.
+   */
+  char *GetUTF8TextForBoxes(const Boxa *boxes);
+
+  /**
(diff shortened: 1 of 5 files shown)


stweil:
@tianrking, do you want to address the comments from Copilot?

Issue #4430, opened 2025-06-20
Tesseract stucks when processing image.


paulius-petkus:
### Current Behavior

Tesseract stucks (hangs) when processing specific file with with page segmentation mode: "**SparseText**".

**Command:**
```
C:\Program Files\Tesseract-OCR\tesseract.exe" page.png page --psm 11 -l eng
```
**Test File:**
https://drive.google.com/file/d/1zg0nHfiudF89PbER12U1Gi3lQzBcyoiw/view?usp=sharing

For reference reasons I am attaching original PDF (cause my goal is to do PDF OCR):
[53402_01040415_doc002.pdf](https://github.com/user-attachments/files/20836181/53402_01040415_doc002.pdf)

### Expected Behavior

Processing should perform fast like with segmentation mode: "Auto".

### tesseract -v

5.5.0.20241111

### Operating System

Windows 11

### Compiler

tesseract --version:
--
tesseract v5.5.0.20241111
 leptonica-1.85.0
  libgif 5.2.2 : libjpeg 8d (libjpeg-turbo 3.0.4) : libpng 1.6.44 : libtiff 4.7.0 : zlib 1.3.1 : libwebp 1.4.0 : libopenjp2 2.5.2
 Found AVX2
 Found AVX
 Found FMA
 Found SSE4.1
 Found libarchive 3.7.7 zlib/1.3.1 liblzma/5.6.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.6
 Found libcurl/8.11.0 Schannel zlib/1.3.1 brotli/1.1.0 zstd/1.5.6 libidn2/2.3.7 libpsl/0.21.5 libssh2/1.11.0

### CPU

intel i7-13700H

stweil:
`tesseract page.png page` does not find any text, but `tesseract page.png page -c thresholding_method=1` does after about 12 seconds.

I assume that page segmentation mode 11 simply takes a lot more time. With thresholding method 1 it terminates after 3:33 minutes in my test.

amitdo:
https://tesseract-ocr.github.io/tessdoc/ImproveQuality.html#binarisation

>Tesseract 5.0.0 added two new Leptonica based binarization methods: Adaptive Otsu and Sauvola. Use tesseract --print-parameters | grep thresholding_ to see the relevant configurable parameters.

PR #4603, opened 2026-08-11
Reject non-text sparse partitions

ilkecan:
## Summary

Refine sparse partition extraction so that a multi-blob `BRT_UNKNOWN` partition is accepted only when its text flow is not `BTFT_NONTEXT`.

Sparse segmentation can classify background texture as unknown partitions with an explicit non-text flow. The extractor previously accepted every multi-blob unknown partition, turning those regions into separate text blocks and making pathological inputs extremely expensive to process.

All text region types remain accepted regardless of flow. Multi-blob unknown partitions with other flow values also remain accepted.

Related to #4430.

## Performance

GCC 15.2 RelWithDebInfo build, with OEM 3, PSM 11 and `OMP_THREAD_LIMIT=1`:

| Input | Before | After |
| --- | ---: | ---: |
| `input-4267x3200.png` | 52.90 s | 0.85 s |
| #4430 `page.png` | >300 s | 5.02 s |

The minimal repro changes from 1,681 extracted blocks and words to 48, and from 6,723 bytes of OCR text to 191 bytes. The removed output came from unknown partitions whose flow was explicitly classified as non-text.
The `main` run ("before") for the #4430 image did not complete within 300 seconds, so it has no completed OCR output for comparison.

<details>

<summary>input-4267x3200.png</summary>

input-4267x3200

</details>

<details>

<summary>background-tile-16.png</summary>

background-tile-16

</details>

## Validation

- Four focused extraction-policy cases pass, covering rejected and retained unknown partitions plus preservation of text region types.
- Full CTest suite: 62/62 passed.
- PSM 11 and PSM 12 output remained byte-identical to `main` across 14 runs covering seven English, Hebrew and Arabic test images.

`BTFT_NONTEXT` is a segmentation heuristic, so weak real text over photos or textures could be excluded if it is classified as both `BRT_UNKNOWN` and `BTFT_NONTEXT`. The focused tests and available smoke set showed no regression but broader scene-text corpus evaluation would provide additional confidence.

### DIFF
--- a/unittest/colpartition_test.cc
+++ b/unittest/colpartition_test.cc
@@ -9,11 +9,42 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "blobbox.h"
 #include "colpartition.h"
+#include "colpartitiongrid.h"
+#include "ocrblock.h"
 
 #include "include_gunit.h"
 
 namespace tesseract {
+
+namespace {
+
+int ExtractedBlockCount(BlobRegionType type, BlobTextFlowType flow, int box_count) {
+  ColPartitionGrid grid(10, ICOORD(0, 0), ICOORD(100, 100));
+  ColPartition *part =
+      ColPartition::FakePartition(TBOX(10, 10, 20, 20), PT_FLOWING_TEXT, type, flow);
+  for (int i = 1; i < box_count; ++i) {
+    const TBOX box(10 + i * 12, 10, 20 + i * 12, 20);
+    part->AddBox(new BLOBNBOX(C_BLOB::FakeBlob(box)));
+  }
+  // Make each BLOBNBOX own its C_BLOB so the fake blobs aren't leaked.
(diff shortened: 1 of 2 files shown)


unittest/colpartition_test.cc
```diff
@@ -9,12 +9,43 @@
...
+#include "blobbox.h"
 #include "colpartition.h"
+#include "colpartitiongrid.h"
+#include "ocrblock.h"
 
 #include "include_gunit.h"
 
 namespace tesseract {
 
+namespace {
```
stweil, line 21: Why was this namespace added?

ilkecan: It’s an anonymous namespace, added to give `ExtractedBlockCount` internal linkage. The function is only used by the tests in this file, so it shouldn’t be exposed as an external symbol in the `tesseract` namespace.

PR #4604, opened 2026-08-13
Add new GitHub workflow which checks commits in a pull request

stweil:
Currently it checks whether the commits have a well formatted author name and a real email address.

Assisted-by: Claude Sonnet 4.6 (Anthropic)

### DIFF
--- a/.github/workflows/check-commit.yml
+++ b/.github/workflows/check-commit.yml
@@ -1,0 +1,157 @@
+name: Check Commit Authors
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


stweil:
We get pull requests from many different authors. I think it is important for legal reasons that these authors don't use some pseudonym or alias name in the commits.

stweil:
In the future we could add more tests to this workflow, for example expect a Signed-off-by if this is desired.

zdenop (approved):

zdenop:
Thank you! Great idea.

PR #4616, opened 2026-08-25
Fix malformed ALTO WC for single-digit and 100 confidences

### LINKED ISSUE #4615: Incorrect ALTO word confidence conversion
### Current Behavior

In the ALTO outputs, the conversion of Tesseract word confidences to ALTO confidence is incorrect. Tesseract confidence is an integer and ALTO confidence is a float from 0.0 to 1.0. The current conversion applies concatenation to "0.", but this is incorrect for 100 percent confidence (yielding "0.100" instead of "1.0") and for confidences under 10 percent (e.g. 9 percent confident leading to "0.9", or 90 percent confident).

### Expected Behavior

The same confidence should be reported through the ALTO XML api as from Tesseract.

### Suggested Fix

This has a small, clear fix and I'll submit a PR.

### tesseract -v

tesseract 5.5.2
 leptonica-1.87.0
  libgif 5.2.2 : libjpeg 8d (libjpeg-turbo 3.1.3) : libpng 1.6.58 : libtiff 4.7.1 : zlib 1.2.12 : libwebp 1.6.0 : libopenjp2 2.5.4
 Found NEON
 Found libarchive 3.8.5 zlib/1.2.12 liblzma/5.8.3 bz2lib/1.0.8 liblz4/1.10.0 libzstd/1.5.7 expat/expat_2.7.3 CommonCrypto/system libb2/system
 Found libcurl/8.7.1 SecureTransport (LibreSSL/3.3.6) zlib/1.2.12 nghttp2/1.64.0

### Operating System

macOS 26 Tahoe

### Other Operating System

_No response_

### uname -a

Darwin HLSGXL6PX0R74 24.6.0 Darwin Kernel Version 24.6.0: Tue Apr 21 20:19:12 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T6041 arm64

### Compiler

Apple clang version 17.0.0 (clang-1700.4.4.1)

### CPU

Apple Silicon M4

_No response_

### Other Information

_No response_

davidlowryduda:
This commit corrects word confidence translation from Tesseract to ALTO. Tesseract uses integer confidence, ALTO uses floats from 0.0 to 1.0. The current code converts Tesseract to ALTO by concatenating onto the string "0.", but this is incorrect for confidences under 10 percent or exactly 100 percent.

For example, if Tesseract reports 9 percent confident, the current implementation converts this to 0.9, or 90 percent confident in ALTO.

We update the conversion.

This closes #4615

### DIFF
--- a/src/api/altorenderer.cpp
+++ b/src/api/altorenderer.cpp
@@ -45,7 +45,8 @@ static void AddBoxToAlto(const ResultIterator *it, PageIteratorLevel level,
 
   if (level == RIL_WORD) {
     int wc = it->Confidence(RIL_WORD);
-    alto_str << " WC=\"0." << wc << "\"";
+    alto_str << " WC=\"" << wc / 100 << "." << (wc % 100 < 10 ? "0" : "")
+             << wc % 100 << "\"";
   } else {
     alto_str << ">";
   }
--- a/src/api/altorenderer.cpp
+++ b/src/api/altorenderer.cpp
@@ -45,8 +45,11 @@ static void AddBoxToAlto(const ResultIterator *it, PageIteratorLevel level,
 
   if (level == RIL_WORD) {
     int wc = it->Confidence(RIL_WORD);
-    alto_str << " WC=\"" << wc / 100 << "." << (wc % 100 < 10 ? "0" : "")
-             << wc % 100 << "\"";
+    // Note: std::fixed and std::setprecision persist on alto_str, but this
+    // is the only floating-point value ever written to that stream, so the
+    // formatting state cannot affect any other output.
+    alto_str << " WC=\"" << std::fixed << std::setprecision(2)
+             << wc / 100.0f << "\"";
   } else {
     alto_str << ">";
   }


davidlowryduda:
I'll also note that I considered adding a test for this behavior, something like

```cpp
// ALTO's WC attribute is a float in 0..1, but Confidence(RIL_WORD) is an
// integer percentage.
TEST_F(TesseractTest, AltoWordConfidenceIsWellFormed) {
  tesseract::TessBaseAPI api;
  if (api.Init(TessdataPath().c_str(), "eng", tesseract::OEM_LSTM_ONLY) == -1) {
    GTEST_SKIP();
  }
  // A deliberately degraded page: upside-down text recognises badly enough to
  // produce single-digit word confidences, which is the case that regressed.
  Image src_pix = pixRead(TestDataNameToPath("phototest-rotated-180.png").c_str());
  CHECK(src_pix);
  api.SetImage(src_pix);
  std::unique_ptr<char[]> alto(api.GetAltoText(0));
  ASSERT_TRUE(alto != nullptr);

  const std::regex wc_attr(R"re(WC="([^"]*)")re");
  const std::regex well_formed(R"re([01]\.[0-9]{2})re");
  const std::string text(alto.get());
  int words = 0;
  int single_digit_confidences = 0;
  for (std::sregex_iterator it(text.begin(), text.end(), wc_attr), end; it != end; ++it) {
    const std::string value = (*it)[1].str();
    ++words;
    EXPECT_TRUE(std::regex_match(value, well_formed))
        << "malformed ALTO WC=\"" << value << "\"; expected a two-decimal float in 0..1";
    const double parsed = std::stod(value);
    EXPECT_GE(parsed, 0.0);
    EXPECT_LE(parsed, 1.0);
    if (parsed > 0.0 && parsed < 0.10) {
      ++single_digit_confidences;
    }
  }
  EXPECT_GT(words, 0) << "no WC attributes in ALTO output";
  // Guards against the fixture silently ceasing to exercise the bug: without a
  // confidence in 1..9 this test would pass against the old encoding too.
  EXPECT_GT(single_digit_confidences, 0)
      << "fixture no longer produces single-digit confidences; pick a more "
         "degraded image or this test no longer covers the regression";
  src_pix.destroy();
}
```

But I decided against it. A successful test requires manufacturing an example where Tesseract confidence is either very small or perfect, and both feel fragile to Tesseract improvements in the future.

src/api/altorenderer.cpp
```diff
@@ -45,7 +45,8 @@ static void AddBoxToAlto(const ResultIterator *it, PageIteratorLevel level,
 
   if (level == RIL_WORD) {
     int wc = it->Confidence(RIL_WORD);
-    alto_str << " WC=\"0." << wc << "\"";
+    alto_str << " WC=\"" << wc / 100 << "." << (wc % 100 < 10 ? "0" : "")
+             << wc % 100 << "\"";
```
stweil:
```suggestion
    // Note: std::fixed and std::setprecision persist on alto_str, but this
    // is the only floating-point value ever written to that stream, so the
    // formatting state cannot affect any other output.
    alto_str << " WC=\"" << std::fixed << std::setprecision(2)
             << wc / 100.0f << "\"";
```

stweil: I'd prefer this less tricky code.

stweil (changes requested):
Thank you for reporting this very old bug.

src/api/altorenderer.cpp
davidlowryduda: Thanks, will do.

stweil:
Thanks again for your contribution!

PR #4620, opened 2026-08-29
fix: require TIFF version bytes before treating stdin as TIFF

### LINKED ISSUE #4619: inputs from imagelist or stdin with filenames starting with 'MM' misdetected as tiff
### Current Behavior

When supplying a 'filelist' containing filenames that start with MM *as stdin*, tesseract is falsely detecting the filelist itself as a tiff file, then failing to open it.  If the filelist is supplied as a file, tesseract initially detects it as a tiff, fails, but then falls back to correctly reading it as a list of filenames.

```
> ls MM11_YP_FMPS_MP00317_part-6_wire_list/*final.pnm | tesseract - OUT txt
Error in pixReadMemFromMultipageTiff: tiff stream not opened
```
where prefixing the path with `./` gets around the problem:
```
> ls ./MM11_YP_FMPS_MP00317_part-6_wire_list/*final.pnm | tesseract - OUT txt
Page 0 : ./MM11_YP_FMPS_MP00317_part-6_wire_list/page-0000.final.pnm
Estimating resolution as 339
Detected 61 diacritics
Page 1 : ./MM11_YP_FMPS_MP00317_part-6_wire_list/page-0001.final.pnm
...
```

When supplied as a file directly:
```
> ls MM11_YP_FMPS_MP00317_part-6_wire_list/*final.pnm > filelist
> tesseract filelist OUT txt
Error in findTiffCompression: tif not opened
Page 0 : MM11_YP_FMPS_MP00317_part-6_wire_list/page-0000.final.pnm
Estimating resolution as 339
Detected 61 diacritics
Page 1 : MM11_YP_FMPS_MP00317_part-6_wire_list/page-0001.final.pnm
...
```

The magic number for a big-endian tiff is `MM##` (where ## is the version number; defacto always 42), so the behavior is annoying but correct.  For little-endian files, the magic number is `II##`.  A potential fix would be load the file first, then detect format.  If that is easily implementable, I defer to the experts.

### Expected Behavior

_No response_

### Suggested Fix

_No response_

### tesseract -v

_No response_

### Operating System

_No response_

### Other Operating System

_No response_

### uname -a

_No response_

### Compiler

_No response_

### CPU

_No response_

_No response_

### Other Information

_No response_

BetterAndBetterII:
## Summary

Leptonica's `findFileFormatBuffer` only checks the TIFF byte-order marker (`II`/`MM`). A stdin filelist whose first filename starts with those letters was misdetected as TIFF, so OCR never walked the list (issue #4619).

After the Leptonica probe reports TIFF, require the classic TIFF version field (`42`) before taking the multipage TIFF path. Otherwise fall through to the existing filelist handler.

## Test plan

- [x] `./build/bin/filelist_test` (3 cases: MM missing, II missing, MM listed image sets `GetInputName`)

Fixes #4619

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1158,6 +1158,7 @@ if !DISABLED_LEGACY_ENGINE
 check_PROGRAMS += equationdetect_test
 endif # !DISABLED_LEGACY_ENGINE
 check_PROGRAMS += fileio_test
+check_PROGRAMS += filelist_test
 check_PROGRAMS += fullyconnected_test
 check_PROGRAMS += genericvector_test
 check_PROGRAMS += heap_test
@@ -1296,6 +1297,10 @@ endif # !DISABLED_LEGACY_ENGINE
 fileio_test_SOURCES = unittest/fileio_test.cc
 fileio_test_CPPFLAGS = $(unittest_CPPFLAGS)
 fileio_test_LDADD = $(TRAINING_LIBS)
+
+filelist_test_SOURCES = unittest/filelist_test.cc
+filelist_test_CPPFLAGS = $(unittest_CPPFLAGS)
+filelist_test_LDADD = $(TRAINING_LIBS) $(LEPTONICA_LIBS)
 
 fullyconnected_test_SOURCES = unittest/fullyconnected_test.cc
 fullyconnected_test_CPPFLAGS = $(unittest_CPPFLAGS)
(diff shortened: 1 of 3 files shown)


stweil:
Wouldn't it be better to fix this in Leptonica?

stweil:
Thanks for trying this fix, but I decided to fix it in Leptonica as it is a bug there.

